PR #9, opened 2026-01-31
fixed tree-sitter dependency versions to enable installation

jasques:
Hello Colby!

Thanks for the codegraph, it looks super interesting. The `packages.json` in the main is broken, and it fails to install the `codegraph` script. This PR fixes it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
   "scripts": {
     "build": "tsc && npm run copy-assets",
     "postinstall": "node scripts/postinstall.js",
-    "copy-assets": "cp -r src/extraction/queries dist/extraction/ && cp src/db/schema.sql dist/db/",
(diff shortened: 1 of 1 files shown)


colbymchenry:
Thanks for flagging the install issues @jasques! The codebase has since gone through a major overhaul — the `package.json` has been completely restructured with cross-platform scripts, and the problematic native deps (`tree-sitter-kotlin`, `tree-sitter-liquid`, `sqlite-vss`, `tree-sitter-dart`) have been moved to `optionalDependencies` so installs degrade gracefully on platforms that can't build them.

This PR would conflict on essentially every line now, so closing it. Appreciate the contribution though — the `overrides` trick for pinning `tree-sitter` is a good one to keep in mind.

PR #80, opened 2026-04-06
feat: optionally use pgsql for faster similarity search

bhushan:


colbymchenry:
Hey Bhushan — thanks for the work on this, but closing.

This PR adds PostgreSQL as an alternative backend for the vector embedding / semantic search feature, but that feature was removed from codegraph entirely in commit 453c39d (see `git log --all -- 'src/vectors/*'`). The `src/vectors/` directory no longer exists on main, so there's no longer a vector store to make pluggable.

The decision to remove vectors predates your PR — codegraph deliberately moved toward deterministic AST-only intelligence (no AI-generated summaries, no embedding-based similarity). If we ever bring semantic search back, the design would happen on the current codebase rather than re-applying this diff.

Appreciate the time you put in, sorry it landed on a now-removed feature.

PR #98, opened 2026-04-26
fix: correctness bugs (Svelte off-by-one, watcher dropped changes, traversal defaults, context clamping)

andreinknv:
## Summary

Four user-visible correctness bugs caught by an independent codebase audit. Each fix has a regression test that fails before the change and passes after.

### 1. Svelte symbols were reported on the wrong line — every time

`src/extraction/svelte-extractor.ts:144`

The `<script>` block regex captures content *starting with* the newline that follows `>`. The inner extractor sees that newline as line 1 of its (1-indexed) input and the first real code on line 2. The old code did `contentStartLine = scriptTagLine + openingTagLines + 1` and added that to the inner extractor's already-1-indexed line numbers — shifting every Svelte symbol's `startLine` / `endLine` off by 1. Dropping the `+1` produces correct positions. Five new tests cover single-line, multi-line opening tag, template-offset, single-line no-newline, and dual module/instance script blocks.

### 2. Watcher silently dropped pending changes on sync failure

`src/sync/watcher.ts:177`

```ts
this.hasChanges = false;     // cleared BEFORE the attempt
this.syncing = true;
try {
  await this.syncFn();
} catch (err) {
  this.onSyncError?.(error); // no retry — batch is lost
}
```

If the sync threw (DB locked, transient FS error) the pending batch was forgotten until a *new* file event arrived. Now `hasChanges = true` is re-set in the catch path so a transient failure schedules a retry on its own. Regression test mocks a fail-then-succeed `syncFn` and asserts the second call happens automatically without a new file event.

### 3. Graph traversal default `maxDepth` was `Infinity`

`src/graph/traversal.ts:14`, `src/types.ts:301`

`limit: 1000` capped returned nodes, but during traversal the `visited` set and BFS/DFS frontier can grow far beyond `limit` on highly connected graphs before the cap kicks in. Default is now `10`. The `TraversalOptions.maxDepth` JSDoc documents that callers who really need exhaustive traversal can still pass `Infinity` explicitly. **This is a public-API behavior change** — existing tests pass with the new default.

Also caps `findPath`'s BFS queue at `FIND_PATH_MAX_QUEUE = 100_000` entries (each holds a cloned path array, so on dense graphs the queue could otherwise consume gigabytes before either finding a path or exhausting the search). Cap is checked at the top of the BFS loop so it's a hard ceiling, not a near-miss.

### 4. `findRelevantContext` did not bound caller-supplied limits

`src/context/index.ts:284`

`searchLimit` is multiplied by 5 inside `findNodesByExactName` and feeds several other unbounded operations; a caller passing `searchLimit: 1_000_000` would pull millions of rows. Now clamped:
- `searchLimit ∈ [1, 100]`
- `maxNodes ∈ [1, 1000]`
- `traversalDepth ∈ [0, 10]`

Regression test asserts a 1e9 input is bounded.

## Files changed

| File | Change |
|---|---|
| `src/extraction/svelte-extractor.ts` | Drop the `+1` in `contentStartLine` |
| `src/sync/watcher.ts` | Re-set `hasChanges` on sync failure |
| `src/graph/traversal.ts` | Default `maxDepth: 10`, add `findPath` queue cap |
| `src/types.ts` | Update `TraversalOptions.maxDepth` doc |
| `src/context/index.ts` | Clamp `searchLimit`/`maxNodes`/`traversalDepth` |
| `__tests__/extraction.test.ts` | 5 Svelte regression tests |
| `__tests__/watcher.test.ts` | 1 retry-after-failure test |
| `__tests__/context.test.ts` | 1 clamp test |

## Test plan

- [x] `npm test` (serialized): 387/387 pass (was 380, added 7)
- [x] Independent reviewer pass before pushing — picked up missing dual-script and single-line test cases (now added) and tightened the `findPath` queue cap to be a hard ceiling

### DIFF
--- a/__tests__/context.test.ts
+++ b/__tests__/context.test.ts
@@ -210,6 +210,19 @@ export function validateEmail(email: string): boolean {
 
       expect(result.nodes.size).toBeLessThanOrEqual(5);
     });
+
+    it('should clamp absurd searchLimit/maxNodes values to safe upper bounds', async () => {
+      // Without clamping, the internal `findNodesByExactName` query would
+      // request `searchLimit * 5` rows — passing 1e9 here would blow out
+      // memory. The call should complete in normal time and not return more
+      // than the hard cap on maxNodes (1000).
+      const result = await cg.findRelevantContext('function', {
+        searchLimit: 1_000_000_000,
+        maxNodes: 1_000_000_000,
+        traversalDepth: 1_000,
+      });
+      expect(result.nodes.size).toBeLessThanOrEqual(1000);
+    });
   });
 
   describe('buildContext()', () => {
--- a/__tests__/extraction.test.ts
+++ b/__tests__/extraction.test.ts
@@ -3079,3 +3079,72 @@ describe('Directory Exclusion', () => {
     expect(files.every((f) => !f.includes('vendor'))).toBe(true);
   });
 });
+
+// =============================================================================
+// Svelte line-number regressions (audit fix)
+// =============================================================================
+
+describe('Svelte line numbering', () => {
+  it('reports symbol line numbers relative to the .svelte file, not the script content', () => {
+    // Line 1: <script>
+    // Line 2: function add(a, b) { return a + b; }
+    // Line 3: </script>
+    const code = `<script>\nfunction add(a, b) { return a + b; }\n</script>\n`;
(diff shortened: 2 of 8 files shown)


colbymchenry:

PR #99, opened 2026-04-26
fix: defense-in-depth hardening (path validation, atomicity, ReDoS, worker lifecycle)

andreinknv:
## Summary

Seven defense-in-depth findings from the audit pass, bundled into one PR for review tractability. Companion to #98 (which covers the user-visible correctness bugs).

### 1. Symlink-resistant path validation in sync paths

`src/extraction/index.ts` — sync() and getChangedFiles() previously did `path.join(rootDir, filePath)` and read the result without checking that the file actually stays within the project (the lexical `validatePathWithinRoot` is used for the indexAll path, but the sync paths skipped it). Both paths now use a new `validatePathWithinRootReal` (in `src/utils.ts`) that calls `realpathSync` so a regular file swapped for a symlink between scan and read is rejected rather than followed.

### 2. `storeExtractionResult` is now atomic

`src/extraction/index.ts:1018-1098`, `src/db/queries.ts` (new `QueryBuilder.transaction`)

The delete + insert + upsert sequence now runs inside a single SQLite transaction. A process kill mid-write can no longer leave the file's old data wiped while the new data is missing. Nested transactions inside `insertNodes` / `insertEdges` / `insertUnresolvedRefsBatch` / `deleteFile` collapse correctly via better-sqlite3's SAVEPOINT support (verified with a quick standalone repro before committing).

### 3. `recycleWorker` is now actually async

`src/extraction/index.ts:573`

Previously declared `void` but called via `await` — the `await` was on `undefined`, so the new worker spawned before the old one finished terminating. Now properly awaits `worker.terminate()` with a 1s timeout, and fires a fire-and-forget terminate after the timeout to avoid zombie threads if WASM is wedged.

### 4. Parse-worker handles unknown messages

`src/extraction/parse-worker.ts:60-78`

If a message arrives with an unknown `type` and an `id`, the worker now sends back a structured error instead of silently dropping it (which previously caused the main thread to block until the per-file timeout). Messages without an `id` are still ignored — no pending promise to unblock.

### 5. ReDoS-safe glob matching

`src/utils.ts` (new `globToSafeRegex`), `src/bin/codegraph.ts:1163`, `src/graph/queries.ts:196`

The `--filter` flag and `findByQualifiedName` previously built regexes by simple substitution (`*` → `[^/]*`, `**` → `.+`). A hostile input like `*****` produced `[^/]*[^/]*[^/]*[^/]*[^/]*` — five chained quantifiers that catastrophically backtrack on long inputs.

The new helper walks character-by-character and coalesces runs of `*`: any run of 2+ stars maps to `.*`, a single `*` to `[^/]*`. Caps input at 1024 chars. No sentinel strings (an earlier draft with a sentinel was rewritten after a reviewer flagged the collision risk if user input happened to contain the sentinel).

### 6. `fileUriToPath` fallback hardening

`src/mcp/index.ts:36-40`

The catch-block fallback for malformed file:// URIs now runs through `path.resolve()` so a `file:///../etc/passwd`-style input is normalized rather than handed raw to downstream filesystem code.

### 7. Reference dedup at extraction + dead-code removal

`src/extraction/tree-sitter.ts` (new `dedupeReferences`), `src/db/queries.ts` (`deleteUnresolvedByNode` removed)

A function that calls `foo()` 100 times now produces 1 unresolved reference instead of 100. The resolver collapsed duplicates eventually, but extraction was wasteful — this is mostly a perf win on large files. Also removed the never-called `deleteUnresolvedByNode` helper (FK cascade handles cleanup automatically).

## Files changed

| File | Change |
|---|---|
| `src/utils.ts` | Add `validatePathWithinRootReal`, replace `globToSafeRegex` with sentinel-free implementation |
| `src/extraction/index.ts` | Use `validatePathWithinRootReal` in 4 sync sites, wrap `storeExtractionResult` in a transaction, make `recycleWorker` async |
| `src/extraction/tree-sitter.ts` | Add `dedupeReferences` applied at end of `extract()` |
| `src/extraction/parse-worker.ts` | Handle unknown message types |
| `src/db/queries.ts` | Add `QueryBuilder.transaction()`, remove dead `deleteUnresolvedByNode` |
| `src/bin/codegraph.ts` | Use `globToSafeRegex` for `--filter` |
| `src/graph/queries.ts` | Use `globToSafeRegex` for `findByQualifiedName` |
| `src/mcp/index.ts` | Run fallback `fileUriToPath` through `path.resolve()` |
| `__tests__/security.test.ts` | 3 new tests for `globToSafeRegex` |

## Test plan

- [x] `npm test` (serialized): 383/383 pass (was 380, added 3 ReDoS tests)
- [x] Standalone repro confirmed nested-transaction safety in better-sqlite3
- [x] Independent reviewer pass — caught a sentinel-collision risk in an earlier draft of `globToSafeRegex` (rewritten without sentinels), a missed force-kill after the recycleWorker timeout (added), and the unused `validatePathWithinRootReal` (now wired into the sync sites)

### DIFF
--- a/__tests__/security.test.ts
+++ b/__tests__/security.test.ts
@@ -533,3 +533,36 @@ describe('Symlink Cycle Detection', () => {
     expect(files).toContain('src/valid.ts');
   });
 });
+
+describe('ReDoS-safe glob matching', () => {
+  it('coalesces runs of `*` so hostile inputs do not produce nested quantifiers', async () => {
+    const { globToSafeRegex } = await import('../src/utils');
+    // Two or more stars collapse to a single recursive wildcard. This is the
+    // ReDoS protection: `*****` doesn't expand to `[^/]*[^/]*[^/]*[^/]*[^/]*`,
+    // which on a long input could catastrophically backtrack.
+    expect(globToSafeRegex('*****')).toBe('.*');
+    expect(globToSafeRegex('**')).toBe('.*');
+
+    // Even a constructed-from-hostile-input regex matches in linear time.
+    const regex = new RegExp(`^${globToSafeRegex('*****')}foo$`);
+    const start = Date.now();
+    // 100k 'a's followed by something that doesn't end in 'foo'.
+    expect(regex.test('a'.repeat(100000) + 'bar')).toBe(false);
+    expect(Date.now() - start).toBeLessThan(500);
+  });
+
+  it('rejects pathologically long glob inputs', async () => {
+    const { globToSafeRegex } = await import('../src/utils');
+    expect(globToSafeRegex('*'.repeat(2000))).toBeNull();
+  });
+
+  it('preserves the standard glob semantics for common patterns', async () => {
+    const { globToSafeRegex } = await import('../src/utils');
+    const body = globToSafeRegex('src/**/*.test.ts');
+    expect(body).toBeDefined();
+    const regex = new RegExp(`^${body}$`);
+    expect(regex.test('src/lib/foo.test.ts')).toBe(true);
+    expect(regex.test('src/lib/foo.ts')).toBe(false);
+    expect(regex.test('other/src/foo.test.ts')).toBe(false);
+  });
+});
--- a/src/bin/codegraph.ts
+++ b/src/bin/codegraph.ts
@@ -23,6 +23,7 @@ import * as path from 'path';
 import * as fs from 'fs';
 import { getCodeGraphDir, isInitialized } from '../directory';
 import { createShimmerProgress } from '../ui/shimmer-progress';
+import { globToSafeRegex } from '../utils';
 
 // Lazy-load heavy modules (CodeGraph, runInstaller) to keep CLI startup fast.
 async function loadCodeGraph(): Promise<typeof import('../index')> {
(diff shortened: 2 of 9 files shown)


colbymchenry:
Thanks for the deep hardening pass here, @andreinknv — there's genuine signal in this PR. After review we're going to close it rather than merge, for two blocking reasons:

1. **The atomic `storeExtractionResult` wrapper breaks indexing on the shipping backend.** It wraps calls like `insertNodes` / `insertEdges` / `insertUnresolvedRefsBatch` in an outer `db.transaction()`, but each of those already opens its own `db.transaction()`. The backend is now `node:sqlite` (`NodeSqliteAdapter`), whose `transaction()` is a bare `BEGIN`/`COMMIT` with no savepoint nesting — so the nested `BEGIN` throws *"cannot start a transaction within a transaction"*, which breaks all indexing. (This PR was validated against better-sqlite3, which we no longer ship.)

2. **`dedupeReferences` regresses caller counts.** Its dedup key is `(fromNodeId, kind, name)` and drops the line number, so multiple call sites of the same callee collapse into a single reference → one edge instead of N. That changes `callCount` and breaks multi-call-site navigation.

The async `recycleWorker`, the parse-worker unknown-message handling, and the ReDoS-safe glob are all good. If you'd like to open a focused PR with just those three (no transaction wrap, no `dedupeReferences`), we'd be glad to review and merge it. Appreciate the effort.

PR #100, opened 2026-04-26
fix(sync): detect HEAD-moving git operations to prevent stale index

andreinknv:
## Summary

Sync silently leaves the index stale after any HEAD-moving git operation (merge, pull, checkout, rebase, reset, even post-commit) because `git status --porcelain` only reports working-tree dirtiness vs HEAD. When the working tree is clean, sync used to short-circuit and report nothing changed, even though the DB still held pre-operation content hashes. MCP queries then returned stale call graphs and missing or deleted symbols to AI assistants with no warning.

## What changed

- **`getGitChangedFiles`** now also unions `git diff --name-status <lastSyncedHead>..HEAD` into the change set, catching commits that arrived without dirtying the working tree.
- **`last_synced_head`** is recorded in the existing `project_metadata` table after every successful `indexAll` and `sync`. No schema change.
- If the recorded HEAD is unreachable (force-push, gc), the function returns `needsFullReindex: true` and the caller drops to the filesystem-scan fallback, which is correct regardless of git history state.
- Same wiring applied to `getChangedFiles()` so MCP staleness signals stay accurate.
- 5 regression tests covering: `git merge`, `git checkout` to a branch with diverged content, committed deletion, unreachable last-synced HEAD (amended commit), and the no-op clean-tree sanity case.

## Affected scenarios (before fix)

Anyone working on a team or across branches:

- `git pull` from a remote — every pull leaves the index stale
- `git checkout <branch>` — every branch switch
- `git rebase` / `git reset --hard`
- `git commit` followed by a post-commit-style auto-sync — also affected, since by post-commit time the working tree is clean

Single-developer single-branch workflows that only edit-then-sync without git operations were unaffected.

## Test plan

- [x] `npm test` — 385/385 pass on macOS
- [x] New regression tests in `__tests__/sync.test.ts` cover the four failure modes plus a clean-tree no-op sanity check
- [x] Pre-existing fs.watch flake (`should trigger sync after file change`) is unrelated and passes when the file is run in isolation

### DIFF
--- a/__tests__/sync.test.ts
+++ b/__tests__/sync.test.ts
@@ -259,4 +259,140 @@ describe('Sync Module', () => {
       expect(result.changedFilePaths).toBeUndefined();
     });
   });
+
+  // Regression tests for the "stale index after HEAD-moving git operation"
+  // bug. `git status` only reports working-tree dirtiness vs HEAD, so a
+  // merge / pull / checkout / rebase / reset (and even post-commit) leaves
+  // a clean tree and used to trick sync into reporting "up to date" while
+  // the DB still held pre-operation content hashes. The fix detects HEAD
+  // movement by comparing current HEAD against a stored last-synced HEAD
+  // and unioning `git diff` output into the changed-file set.
+  describe('HEAD-moving git operations', () => {
+    let testDir: string;
+    let cg: CodeGraph;
+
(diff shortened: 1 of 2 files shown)


colbymchenry:

PR #101, opened 2026-04-26
fix: extraction/resolution accuracy (BOM, retry strip, framework regex)

andreinknv:
## Summary

Three extraction/resolution accuracy bugs caught by an audit pass, bundled in one PR for review tractability.

### 1. UTF-8 BOM caused spurious "modified" hash mismatches

`src/extraction/index.ts` (`hashContent`)

A file written with a BOM by one editor and re-saved without a BOM by another (VSCode strips by default; some Windows editors preserve it) hashed to two different values. Sync then reported the file as modified on every run despite no real change. `hashContent` now strips a leading `U+FEFF` before hashing.

### 2. Parse-retry comment strip was a no-op for Python, Ruby, etc.

`src/extraction/index.ts` (last-resort retry path), `src/utils.ts`

The "shrink the file by removing comment-only lines" fallback (which fires when a file repeatedly crashes the WASM parser) used `/^\s*\/\//.test(line)` for every language. Python (`#`), Ruby (`#`), shell, etc. had nothing stripped — the retry parsed the same content that had already crashed and the file silently stayed unindexed. Added a per-language `LINE_COMMENT_MARKER` table and a `stripCommentLinesForRetry` helper at the retry call site.

### 3. Framework route extractors matched docstrings and comments

`src/resolution/frameworks/{python,express,laravel,rust,csharp}.ts`

`pattern.exec(content)` ran the route regex over raw file content, so a route example in a Python docstring or a commented-out route in JS/Rust/C# XML-doc comment was extracted as a real `route` node. AI assistants then saw phantom routes that don't exist in the running app.

Added `stripCommentsForRegex(content, language)` in `utils.ts` which neutralizes:
- Block comments (`/* ... */`) for C-family languages
- Whole-line single-line comments per language (`//`, `#`)
- Triple-quoted strings (`"""…"""`, `'''…'''`) for Python (the docstring carrier)
- `=begin…=end` blocks for Ruby

Newlines are preserved so `match.index` derived from the stripped content maps back to the same line number in the original. Deliberately does **not** strip arbitrary string literals — those carry the real route paths the regex needs to see.

Applied at the top of every affected framework extractor (django, flask, fastapi, express, laravel, actix/rocket/axum, aspnet).

## Files changed

| File | Change |
|---|---|
| `src/utils.ts` | Add `stripBom`, `stripCommentsForRegex`, `stripCommentLinesForRetry`, `LINE_COMMENT_MARKER` |
| `src/extraction/index.ts` | Strip BOM in `hashContent`; use language-aware retry strip |
| `src/resolution/frameworks/python.ts` | Strip comments before django/flask/fastapi route regex |
| `src/resolution/frameworks/express.ts` | Strip comments before express route regex |
| `src/resolution/frameworks/laravel.ts` | Strip comments before laravel route regex |
| `src/resolution/frameworks/rust.ts` | Strip comments before actix/rocket/axum route regex |
| `src/resolution/frameworks/csharp.ts` | Strip comments before aspnet route regex |
| `__tests__/extraction-resolution-accuracy.test.ts` | 21 regression tests |

## Test plan

- [x] `npm test`: 400/400 pass on macOS (one pre-existing fs.watch flake under parallel load, passes in isolation — unrelated to this change)
- [x] 21 new tests cover: BOM normalization, per-language line stripping (including unknown-language no-op), and false-positive prevention for every affected framework

### DIFF
--- a/__tests__/extraction-resolution-accuracy.test.ts
+++ b/__tests__/extraction-resolution-accuracy.test.ts
@@ -231,6 +231,24 @@ describe('Framework regex no longer matches docstrings/comments (bug #4)', () =>
       expect(names.some((n) => n.includes('/real'))).toBe(true);
       expect(names.some((n) => n.includes('/docfake'))).toBe(false);
     });
+
+    it('skips minimal-API MapGet/MapPost calls inside comments', () => {
+      // Regression: the minimalApiPattern loop below the routePatterns
+      // loop was initially missed when applying the strip helper, leaving
+      // commented-out `app.MapGet("/x")` calls extracted as real routes.
+      const content = [
+        '// app.MapGet("/linefake", h);',
+        '/*',
+        ' * app.MapPost("/blockfake", h);',
+        ' */',
+        'app.MapGet("/real", h);',
+      ].join('\n');
+      const nodes = aspnetResolver.extractNodes!('Program.cs', content);
+      const names = nodes.map((n) => n.name);
+      expect(names.some((n) => n.includes('/real'))).toBe(true);
+      expect(names.some((n) => n.includes('/linefake'))).toBe(false);
+      expect(names.some((n) => n.includes('/blockfake'))).toBe(false);
+    });
   });
 });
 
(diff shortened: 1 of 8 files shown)


colbymchenry:

PR #102, opened 2026-04-26
fix(db): enforce UNIQUE on edges so INSERT OR IGNORE actually dedupes

andreinknv:
## Summary

The `edges` table has `id INTEGER PRIMARY KEY AUTOINCREMENT` and **no other UNIQUE constraint**. The codebase uses `INSERT OR IGNORE INTO edges ...` clearly intending dedup, but autoincrement IDs never conflict, so the `OR IGNORE` was silently a no-op. Any path that re-emits the same edge (resolver retries, partial-failure re-runs, framework extractors that double-emit) accumulated duplicates → inflated call graphs in `codegraph_callers` / `codegraph_callees` results.

## What changed

- **Schema**: `src/db/schema.sql` adds `UNIQUE INDEX idx_edges_unique ON edges(source, target, kind, COALESCE(line, -1), COALESCE(col, -1))` for fresh installs.
- **Migration v4** (`src/db/migrations.ts`): existing databases first dedupe pre-existing duplicates (`DELETE ... WHERE id NOT IN (SELECT MIN(id) ... GROUP BY ...)`) then create the index. Both run inside the migration transaction wrapper, so a crash leaves the DB consistent.
- **`COALESCE(line, -1), COALESCE(col, -1)`** in the unique key: SQLite treats raw NULLs in a UNIQUE index as distinct, which would defeat dedup for edges that don't carry a line/column. The codebase's line numbers are 1-indexed and column numbers are non-negative everywhere — `-1` is a safe sentinel (verified via grep across `src/`).
- `CURRENT_SCHEMA_VERSION` bumped 3 → 4. Two existing version-pinned tests updated.

## Files changed

| File | Change |
|---|---|
| `src/db/schema.sql` | Add `UNIQUE INDEX idx_edges_unique` for fresh installs |
| `src/db/migrations.ts` | Bump version to 4; add migration v4 (dedup + index) |
| `__tests__/edges-unique.test.ts` | 7 regression tests |
| `__tests__/foundation.test.ts` | Update expected schema version |
| `__tests__/pr19-improvements.test.ts` | Update expected schema version |

## Test coverage

- Exact duplicates (same source/target/kind/line/col) collapse to one row
- NULL line edges with same source/target/kind collapse via `COALESCE`
- Different lines / different kinds remain distinct
- Batch `insertEdges` dedupes within one call
- Re-emission across 100 cycles leaves a single row
- Migration test: simulate a v3-shaped DB, insert duplicates via raw SQL, run `runMigrations(db, 3)`, assert dedup occurred and the constraint is now enforced

## Test plan

- [x] `npm test`: 387/387 pass on macOS (one pre-existing fs.watch flake under parallel load, passes in isolation — unrelated to this change)
- [x] Independent reviewer pass before pushing — verdict APPROVE (informational nits only: schema.sql header comment still says "Version 1", and the migration test simulates the v3 state by surgically rolling back rather than reproducing the historical schema.sql; both deemed acceptable)

### DIFF
--- a/__tests__/edges-unique.test.ts
+++ b/__tests__/edges-unique.test.ts
@@ -1,0 +1,166 @@
+/**
+ * Edge Uniqueness Tests
+ *
+ * Regression tests for the bug where `INSERT OR IGNORE INTO edges` was
+ * silently a no-op: the only candidate key was the AUTOINCREMENT id (which
+ * never conflicts), so duplicate edges accumulated on every re-emission /
+ * re-resolution.
+ *
+ * Fix: a UNIQUE index on (source, target, kind, COALESCE(line, -1),
+ * COALESCE(col, -1)) backs a fresh-install schema and is also applied via
+ * migration v4 (with a dedup pass over existing rows).
+ */
+
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as os from 'os';
+import { DatabaseConnection } from '../src/db';
+import { QueryBuilder } from '../src/db/queries';
+import { Edge, Node } from '../src/types';
+import { runMigrations, getCurrentVersion, CURRENT_SCHEMA_VERSION } from '../src/db/migrations';
+
+function tempDb(): { dir: string; db: DatabaseConnection; q: QueryBuilder } {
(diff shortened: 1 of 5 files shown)


colbymchenry:

PR #103, opened 2026-04-26
fix(scan): honor .codegraphignore on the git fast path

andreinknv:
## Summary

The `.codegraphignore` marker (per-directory opt-out from indexing) was respected by `scanDirectoryWalk` (the filesystem-walk fallback) but silently ignored by `getGitVisibleFiles` (the git fast path) and `getGitChangedFiles` (sync's git path). Same project gave different file sets depending on whether `.git` existed — typically the marker "worked" only on non-git scratch projects and was a no-op everywhere else, the opposite of how most users encounter it.

## What changed

Two helpers in `src/extraction/index.ts`:

- `findCodegraphIgnoredDirs(rootDir, files)` — walks parent directories of the given file list, returns the set of dirs that contain a `.codegraphignore` marker. Walks once per unique parent directory with an early-out on shared ancestors.
- `isUnderCodegraphIgnoredDir(filePath, ignoredDirs)` — true if `filePath` lives under any of those dirs.

Applied at three sites:
- `scanDirectory` and `scanDirectoryAsync` — between the git file list and the include-pattern filter
- `getGitChangedFiles` — refactored to a two-pass collect-then-bucketize so the ignored-dir set is built once from the candidate paths

The marker file itself does not need to be tracked by git — `fs.existsSync` catches it whether it was committed or added as a local override.

## Files changed

| File | Change |
|---|---|
| `src/extraction/index.ts` | Add `findCodegraphIgnoredDirs` + `isUnderCodegraphIgnoredDir`; apply in `scanDirectory`, `scanDirectoryAsync`, `getGitChangedFiles` |
| `__tests__/codegraphignore.test.ts` | 6 regression tests |

## Test coverage

- `scanDirectory` honors marker in a subdir on the git fast path
- Marker at project root excludes everything
- Marker in one subdir does not affect siblings
- Marker added as local override (untracked) is still respected
- Non-git fallback parity (sanity, pre-existing behavior unchanged)
- Sync (`getGitChangedFiles`) ignores changes inside marker dirs

## Known limitation (pre-existing, out of scope)

If a `.codegraphignore` marker is added **after** files in that directory have already been indexed, the next sync via the git fast path won't proactively delete those stale rows — `git status` doesn't report unchanged files. The next full `indexAll` (or a sync that falls into the filesystem-walk path) will clean them up. This is an existing characteristic of the git fast path; documenting here for transparency rather than fixing in this PR.

## Test plan

- [x] `npm test`: 386/386 pass on macOS (one pre-existing fs.watch flake under parallel load, passes in isolation)
- [x] Independent reviewer pass before pushing — verdict APPROVE; addressed two info-level cleanups (JSDoc accuracy, removed dead try/catch around `fs.existsSync` which never throws)

### DIFF
--- a/__tests__/codegraphignore.test.ts
+++ b/__tests__/codegraphignore.test.ts
@@ -1,0 +1,168 @@
+/**
+ * .codegraphignore Tests
+ *
+ * Regression test for the bug where the .codegraphignore marker file was
+ * honored by the filesystem-walk fallback (`scanDirectoryWalk`) but
+ * silently ignored by the git fast path (`getGitVisibleFiles` and
+ * `getGitChangedFiles`). Same project gave different file sets depending
+ * on whether `.git` existed.
+ */
+
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as os from 'os';
+import { execFileSync } from 'child_process';
+import { scanDirectory } from '../src/extraction';
+import { DEFAULT_CONFIG, CodeGraphConfig } from '../src/types';
+import CodeGraph from '../src/index';
+
+function tempDir(prefix: string): string {
+  return fs.mkdtempSync(path.join(os.tmpdir(), prefix));
+}
+
+function git(cwd: string, ...args: string[]) {
+  execFileSync('git', args, { cwd, stdio: 'pipe' });
+}
+
+const config: CodeGraphConfig = {
+  ...DEFAULT_CONFIG,
+  include: ['**/*.ts'],
+  exclude: [],
+};
+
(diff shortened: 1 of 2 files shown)


colbymchenry:
Thanks @andreinknv. Looking closely, `.codegraphignore` doesn't currently exist on `main`, so this is effectively a new opt-out *feature* rather than the git-fast-path parity *fix* it's framed as — nothing regresses without it. We've decided to stick with `.gitignore` (which the scanner already honors) as the single ignore mechanism, rather than introduce a second one with its own semantics to document and maintain. Closing for that reason.

The implementation itself was clean and the tests were thorough — if `.gitignore` ever proves insufficient for a real workflow, we're open to revisiting.

Issue #87, opened 2026-04-17
vector search


zzqq131:
Hi, I saw that the  commit wiped out the whole vector search and embedding module. I’m kinda curious—what’s the reasoning behind that? I’d love to hear your thoughts.

andreinknv:
For anyone landing here looking for context — sharing what I dug up plus a complementary path.

## Why it was removed (timeline)

- **Mar 18 (#61)** — Quantized embeddings shipped to fight V8 WASM Zone OOM crashes from #54.
- **Apr 3 (#74)** — Embeddings *promoted* to always-on, with concrete evidence that semantic search outperformed FTS for AI-assistant queries.
- **Apr 7** — Issue #81 reports the same WASM Zone OOM persists on Node 25.x. Same day, the maintainer:
  - posted a diagnosis on #81 blaming V8 turboshaft + Node 25
  - committed `7507605` (Node 25+ runtime warning)
  - committed `453c39d` removing all embeddings code

Stated reason: *"Reduces package size and complexity while maintaining core static analysis capabilities."* Implied (from the timing): `@xenova/transformers` ships ONNX as a *second* WASM toolchain on top of tree-sitter grammars, and V8's turboshaft Zone allocator can't handle JIT-compiling both. Cutting one of the two WASMs is the most reliable mitigation. (Apr 24 — `cadam11` reports the same OOM on Node 24, confirming the underlying V8 issue isn't fully fixable from userspace alone.)

## A complementary path: improve FTS without bringing back model deps

Most of what semantic search bought wasn't conceptual reasoning — it was robustness to:
- **Partial-identifier queries**: SQLite's default tokenizer treats `getParser` as one indivisible token, so a query for `parser` never matches it.
- **Morphological variants**: `parsing` vs `parser` vs `parses`.
- **Natural-language noise**: `"how does X work"` floods the BM25 ranking with stopword hits against any prose docstring.

Each of those is fixable in pure SQLite FTS5 without any model dependency. I prototyped + measured against codegraph's own `src/`, using the exact queries from #74:

| Query | Baseline (current main) | After (subwords + Porter + stopwords) |
|---|---:|---:|
| `ExtractionOrchestrator` | #1 | **#1** |
| `how does file parsing work` (target: `getParser`) | NOT FOUND in 20 | **#2** |
| `database connection management` (target: `DatabaseConnection`) | #18 | **#1** |
| `resolves references between modules` (target: `resolveOne`) | #19 | **#2** |

[...]

Issue #11, opened 2026-01-31
UNIQUE constraint failed: nodes.id during indexing on React/JSX codebase


ronniehyslop:
## Description

When running `codegraph index` on a medium-to-large React/JavaScript codebase, the indexing process fails with a SQLite unique constraint violation. The error occurs consistently at around 10-34% progress through the indexing phase.

## Error Message

```
✗ Failed to index: UNIQUE constraint failed: nodes.id
```

## Environment

- **CodeGraph Version:** 0.3.1
- **Node.js Version:** v20.19.0
- **Platform:** Linux (WSL2) 6.6.87.2-microsoft-standard-WSL2
- **Installation Method:** `npm install -g @colbymchenry/codegraph --ignore-scripts` followed by `npm rebuild`

## Codebase Characteristics

- ~1,000+ source files (excluding e2e tests)
- Mixed JavaScript (.js, .jsx) and TypeScript (.ts, .tsx)
- React 19 application with Vite
- Standard project structure: `src/components/`, `src/features/`, `src/hooks/`, etc.

## Steps to Reproduce

1. Initialize codegraph in a React project:
   ```bash
   codegraph init .
   ```

2. Run the indexer:
   ```bash
   codegraph index .
   ```

3. Observe the failure at approximately 10-34% progress during the "Parsing code" phase.

## Partial Index State

Before the failure, the indexer successfully processes some files:

```
Index Statistics:
  Files:     106
  Nodes:     412
  Edges:     21

Nodes by Kind:
  function        385
  method          21
  class           3
  interface       3

Files by Language:
  javascript      57
  jsx             46
  typescript      3
```

## Attempted Workarounds

1. **Excluding e2e directory:** Added `"**/e2e/**"` to the exclude list in `config.json` - issue persists
2. **Fresh initialization:** Removed `.codegraph/` directory and reinitialized - same error
3. **Multiple attempts:** The failure point varies slightly but always occurs in the same general range

## Analysis

The error suggests that the node ID generation algorithm is producing duplicate IDs for different code symbols. This could be caused by:

- Hash collisions in the ID generation logic
- Similar function/variable names across different files producing identical IDs
- Re-exported symbols being processed multiple times

## Expected Behavior

The indexer should complete successfully, handling any potential ID conflicts gracefully (e.g., by incorporating file path or additional context into the ID hash).

## Additional Context

[...]

PR #104, opened 2026-04-26
feat(search): subword tokens + Porter stemmer + stopword filter for FTS

andreinknv:
## Context

Embeddings were removed in 453c39d (issue #87). Since then, all search quality has to come from FTS. The original PR #74 documented several queries where FTS-only badly trailed semantic search — the dominant cause is that SQLite's default tokenizer treats `getParser` as one indivisible token, so a query for `parser` never matches it at the FTS layer no matter how clever the post-hoc rescorer is.

This PR closes most of that gap with three compounding, model-free changes.

## What changed

1. **Subword tokens.** New `name_subwords` column on `nodes` populated with the camel/snake split of the identifier (kept alongside the original) and indexed by FTS5 at weight 10×. A query for `parser` now finds `getParser` *at the FTS layer*, not just via post-hoc rescoring on the limited candidate set BM25 surfaces.

2. **Porter stemmer.** `tokenize="porter unicode61"` on the FTS table collapses morphological variants — `parser` / `parsing` / `parses` all stem to `pars` so a natural-language query matches identifier subwords and docstring prose alike.

3. **Stopword stripping.** `searchNodesFTS` filters stopwords from the query before the OR-join. Without this, words like `how` / `does` / `the` become OR'd FTS hits against any prose-bearing docstring and crowd out the actually-relevant identifier tokens. Reuses the existing `STOP_WORDS` set in `src/search/query-utils.ts` via a new shared `filterStopwords` helper (no second source of truth).

## Empirical results

Bench: codegraph's own `src/`, 1242 nodes, 71 files. Same query set as PR #74's evidence. Lower rank = better.

| Query | baseline (current main) | this PR |
|---|---:|---:|
| `ExtractionOrchestrator` | **1** | **1** |
| `how does file parsing work` (target: `getParser`) | **NOT FOUND in 20** | **#2** |
| `database connection management` (target: `DatabaseConnection`) | **#18** | **#1** |
| `resolves references between modules` (target: `resolveOne`) | **#19** | **#2** |

Mean rank: **~14 → 1.5**. Same workload, no model dependencies, no install-time changes, no Node version constraints.

## What I tested and rejected

**Concept-mode docstring re-weighting** (swap BM25 weights to favor docstring 20× when query contains stopwords): regressed `how does file parsing work` from #4 to #11 because amplifying docstring weight floods results with prose-keyword spam more than it lifts truly relevant prose. Not included.

## Migration v4

Existing v3 databases get migrated by:
- Adding the `name_subwords` column to `nodes` (idempotent `PRAGMA table_info` guard so a re-run after partial DDL failure doesn't fail with "duplicate column")
- Dropping the old FTS table + triggers (FTS5 tokenizer cannot be `ALTER`ed)
- Recreating FTS *without triggers*
- Backfilling `name_subwords` for every existing node
- Rebuilding the FTS index in one shot via `INSERT INTO nodes_fts(nodes_fts) VALUES('rebuild')`
- Recreating the triggers afterward (so they don't fire mid-backfill — earlier prototype runs corrupted FTS5 when triggers fired against half-populated shadow tables)

`CURRENT_SCHEMA_VERSION` 3 → 4. Two existing version-pinned tests updated.

## Files changed

| File | Change |
|---|---|
| `src/utils.ts` | Add `splitIdentifierTokens`, `buildNameSubwords` (with dedup) |
| `src/search/query-utils.ts` | Add shared `filterStopwords` helper using existing `STOP_WORDS` |
| `src/db/schema.sql` | Add `name_subwords` column, add it to `nodes_fts`, add `tokenize="porter unicode61"`, update three triggers |
| `src/db/migrations.ts` | Bump version to 4; add migration v4 with idempotent ALTER guard |
| `src/db/queries.ts` | Populate `name_subwords` on insert/update; new BM25 weights `(0,20,5,1,2,10)`; stopword filter in `searchNodesFTS` |
| `__tests__/foundation.test.ts`, `__tests__/pr19-improvements.test.ts` | Update expected schema version |
| `__tests__/search-quality.test.ts` (NEW) | 21 regression tests: helpers, end-to-end search, full v3-to-v4 migration with a real v3-shape DB, and migration idempotency on partial-DDL re-run |

## Test plan

- [x] `npm test`: 404/404 pass on macOS (one pre-existing fs.watch flake under parallel load, passes in isolation — unrelated)
- [x] Bench script confirms the rank lifts in the table above
- [x] Independent reviewer pass before pushing — addressed three findings:
  - merged duplicate stopword sets (now uses `STOP_WORDS` from `query-utils.ts`)
  - dedup tokens in `buildNameSubwords` (`parse` no longer stores `"parse parse"`)
  - made migration idempotent on partial-DDL re-run (re-running after a crash that landed only `ALTER TABLE` no longer fails)

## Notes for review

- This PR is **independent of #102** (which also bumps schema version to 4). If #102 lands first, I'll rebase to v5.
- No new dependencies. No install-time changes. No Node version constraints.
- The change is gradual — fresh installs hit the new schema directly via `schema.sql`, existing installs run migration v4 on first open.

### DIFF
--- a/__tests__/foundation.test.ts
+++ b/__tests__/foundation.test.ts
@@ -305,7 +305,7 @@ describe('Database Connection', () => {
 
     const version = db.getSchemaVersion();
     expect(version).not.toBeNull();
-    expect(version?.version).toBe(3);
+    expect(version?.version).toBe(4);
 
     db.close();
   });
--- a/__tests__/pr19-improvements.test.ts
+++ b/__tests__/pr19-improvements.test.ts
@@ -299,7 +299,7 @@ describe('Best-Candidate Resolution', () => {
 describe('Schema v2 Migration', () => {
   it.skipIf(!HAS_SQLITE)('should have correct current schema version', async () => {
     const { CURRENT_SCHEMA_VERSION } = await import('../src/db/migrations');
-    expect(CURRENT_SCHEMA_VERSION).toBe(3);
+    expect(CURRENT_SCHEMA_VERSION).toBe(4);
   });
 
   it.skipIf(!HAS_SQLITE)('should have migration for version 2', async () => {
--- a/__tests__/search-quality.test.ts
+++ b/__tests__/search-quality.test.ts
@@ -1,0 +1,302 @@
+/**
+ * Search Quality Tests
+ *
+ * Regression tests for the FTS improvements that bring natural-language
+ * and partial-identifier queries into the top of the result set:
+ *   - Subword tokens (camel/snake split) so `parser` finds `getParser`.
+ *   - Porter stemmer so `parsing` matches `parser`/`parses`.
+ *   - Stopword stripping so `"how"` / `"the"` don't crowd out the
+ *     real terms via docstring matches.
+ *
+ * All measurements were captured against codegraph's own src/ during
+ * development. Targets that previously ranked #18, #19, or weren't in
+ * the top 20 jump to the top 5.
+ */
+
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as os from 'os';
+import { DatabaseConnection } from '../src/db';
+import { QueryBuilder } from '../src/db/queries';
+import { Node } from '../src/types';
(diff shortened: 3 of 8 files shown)


colbymchenry:

PR #114, opened 2026-04-27
feat(graph): config_refs — env var read sites as queryable graph data

andreinknv:
> **🔄 Rebased onto post-refactor patterns (config-refs (env var read sites))**
> This PR has been rebased to use the per-thing file layout established by refactor PRs #116/#117/#118/#119. Behavior shipped is identical to the original; only the internal code shape changed.
>
> **What this PR now ships:**
>
> - **`src/index-hooks/config-refs.ts`** — registered `IndexHook` (incremental on sync, full on indexAll)
> - **`src/db/migrations/006-config-refs.ts`** — `config_refs` table
> - **`src/config-refs/index.ts`** — pure regex extractor
> - **`src/mcp/tools/config.ts`** — `CONFIG_TOOL: ToolModule` for `codegraph_config`
> - `src/index-hooks/registry.ts`, `src/db/migrations/index.ts`, `src/mcp/tools/registry.ts`, `src/mcp/tools/types.ts` — registry entries
> - `src/db/queries.ts`, `src/db/schema.sql`, `src/types.ts`, `src/default-config.ts`, `src/config.ts`, `src/index.ts`, `src/mcp/tools.ts` — methods + handler
> - `__tests__/config-refs.test.ts`
>
> **What you'll see in the diff:** until #116-#119 land, the diff against `main` includes the refactor commits as part of the branch history. After they merge, GitHub auto-shrinks this PR's diff to just the files above. Full context: #120.
>
> ---

## Summary

Extracts environment-variable read sites (`process.env.X`, `os.getenv("X")`, `os.environ`, `os.Getenv`, `System.getenv`, `ENV[...]`, `env!`, `std::env::var`) into a new `config_refs` table. Surfaces via a new `codegraph_config` MCP tool to answer "what reads `OBSIDIAN_PORT`?" and "what config does this codebase read?" without grep noise.

**Why a separate table, not graph nodes/edges** — env vars are a global namespace with no single source-of-truth file. Giving them a synthetic `file_path` would pollute the main graph. Treating them like `symbol_issues` (a side-table queried via a dedicated tool) keeps the core graph clean.

## Empirical signal on `mcp-obsidian-extended` (the spike target)

Indexing that repo with this PR's build reproduces the spike numbers exactly:

| Key | Reads |
|---|---:|
| `TOOL_PRESET` | 8 |
| `OBSIDIAN_PORT` | 8 |
| `OBSIDIAN_SCHEME` | 7 |
| `INCLUDE_TOOLS` | 7 |
[...]

PR #113, opened 2026-04-27
feat(graph): symbol-level issue history mined from `Fixes #N` commits

andreinknv:
> **🔄 Rebased onto post-refactor patterns (Issue-history attribution)**
> This PR has been rebased to use the per-thing file layout established by refactor PRs #116/#117/#118/#119. Behavior shipped is identical to the original; only the internal code shape changed.
>
> **What this PR now ships:**
>
> - **`src/index-hooks/issue-history.ts`** — registered `IndexHook` (full re-mine + force-push recovery)
> - **`src/db/migrations/005-symbol-issues.ts`** — `symbol_issues` table
> - **`src/issue-history/index.ts` + `src/issue-history/parse-diff.ts`** — pure miner
> - `src/index-hooks/registry.ts`, `src/db/migrations/index.ts` — 2 lines each
> - `src/db/queries.ts`, `src/db/schema.sql`, `src/types.ts`, `src/default-config.ts`, `src/config.ts`, `src/index.ts` — methods + config
> - `src/mcp/tools.ts` — `codegraph_node` annotation surfaces issue history line
> - `__tests__/issue-history.test.ts`
>
> **What you'll see in the diff:** until #116-#119 land, the diff against `main` includes the refactor commits as part of the branch history. After they merge, GitHub auto-shrinks this PR's diff to just the files above. Full context: #120.
>
> ---

## Summary

Mines git history for `Fixes #N` / `Closes #N` / `Resolves #N` commits and attributes their diff hunks to the symbols they touched. Stored in a new `symbol_issues` table; surfaced inline in `codegraph_node` MCP responses as `**Issues:** #37, #68, #69 (modified)`.

**Why hunk-level, not file-level** — spike data showed file-level issue→symbol gives ~40 symbols/issue (mostly noise) because every issue touches files with many irrelevant symbols. Hunk-level cuts noise 78% AND uniquely enables the agent-useful query "which symbols appear in multiple bug fixes?" which file-level cannot answer cleanly.

## Empirical signal on this repo's history

Running on this branch's main (7 distinct issues mined):

| Symbol | # Issues | File |
|---|---:|---|
| `runInstaller` | 3 | `src/installer/index.ts` |
| `writeMcpConfig` | 2 | `src/installer/config-writer.ts` |
| `writeJsonFile` | 2 | `src/installer/config-writer.ts` |
| `getMcpServerConfig` | 2 | `src/installer/config-writer.ts` |
[...]

PR #112, opened 2026-04-27
feat(graph): PageRank centrality + per-file churn metrics

andreinknv:
> **🔄 Rebased onto post-refactor patterns (Centrality + churn + hotspots)**
> This PR has been rebased to use the per-thing file layout established by refactor PRs #116/#117/#118/#119. Behavior shipped is identical to the original; only the internal code shape changed.
>
> **What this PR now ships:**
>
> - **`src/index-hooks/centrality.ts`** + **`src/index-hooks/churn.ts`** — two registered hooks
> - **`src/db/migrations/004-centrality-churn.ts`** — adds `nodes.centrality` + `files.{commit_count,loc,first_seen_ts,last_touched_ts}`
> - **`src/mcp/tools/hotspots.ts`** — `HOTSPOTS_TOOL: ToolModule` for `codegraph_hotspots`
> - **`src/centrality/index.ts`** + **`src/churn/index.ts`** — pure compute modules
> - `src/index-hooks/registry.ts`, `src/db/migrations/index.ts`, `src/mcp/tools/registry.ts`, `src/mcp/tools/types.ts` — registry entries (~2 lines each)
> - `src/db/queries.ts`, `src/db/schema.sql`, `src/types.ts`, `src/default-config.ts`, `src/config.ts`, `src/index.ts`, `src/mcp/tools.ts` — query methods, config flags, public API, handler
> - `__tests__/centrality.test.ts`, `__tests__/churn.test.ts`
>
> **What you'll see in the diff:** until #116-#119 land, the diff against `main` includes the refactor commits as part of the branch history. After they merge, GitHub auto-shrinks this PR's diff to just the files above. Full context: #120.
>
> ---

## Summary

Adds two graph-derived signals — **PageRank centrality** on nodes and **per-file churn** mined from git — and combines them into a `codegraph_hotspots` MCP tool that ranks files by `risk = centrality × churn`. Empirically validated before implementation: top hotspot on this repo's own history is `src/extraction/tree-sitter.ts` (47 commits × 2347 LOC), exactly the file PR #77 split.

Why both signals together: centrality alone surfaces architecturally-important code; churn alone surfaces volatile code. The intersection — central code that changes often — is where bugs accumulate, and that's the agent-facing query this PR ships.

## What's added

[...]

PR #95, opened 2026-04-26
feat: add SQL language support (Postgres / MySQL / SQLite / T-SQL)

andreinknv:
> **🔄 Rebased onto post-refactor patterns (SQL language)**
> This PR has been rebased to use the per-thing file layout established by refactor PRs #116/#117/#118/#119. Behavior shipped is identical to the original; only the internal code shape changed.
>
> **What this PR now ships:**
>
> - **`src/extraction/languages/sql.ts`** — `SQL_DEF: LanguageDef` (custom + grammar)
> - **`src/extraction/sql-extractor.ts`** — `SqlExtractor` class
> - **`src/extraction/wasm/tree-sitter-sql.wasm`** — vendored grammar
> - `src/extraction/languages/registry.ts` — 2 lines
> - `src/types.ts` — 1 line (`'sql'` in `Language` union)
> - `__tests__/extraction.test.ts` — SQL tests
>
> **What you'll see in the diff:** until #116-#119 land, the diff against `main` includes the refactor commits as part of the branch history. After they merge, GitHub auto-shrinks this PR's diff to just the files above. Full context: #120.
>
> ---

## Summary

Adds extraction for SQL DDL files (`.sql` / `.ddl` / `.dml`) using the [DerekStride/tree-sitter-sql v0.3.11](https://github.com/DerekStride/tree-sitter-sql) grammar (MIT). Covers ANSI SQL plus the major dialects' shared shape: **PostgreSQL, MySQL / MariaDB, SQLite, and T-SQL** — the most-used and most-popular ones. Closes #85.

The wasm is built from the released source via `tree-sitter-cli build --wasm` (which downloads its own wasi-sdk; no Docker or local emcc needed) and vendored under `src/extraction/wasm/tree-sitter-sql.wasm`, alongside the existing `tree-sitter-pascal.wasm`. The Pascal-only special case in `grammars.ts` is generalized into a `VENDORED_WASM_LANGUAGES` set.

## How SQL maps to the graph

SQL is declarative — no functions/classes/methods in the OO sense — so this uses a self-contained `SqlExtractor` (same pattern as `LiquidExtractor` / `HclExtractor`) dispatched from `extractFromSource()`.

| SQL statement | NodeKind | qualified name |
|---|---|---|
| `CREATE TABLE [schema.]name` | `class` | `[schema.]name` |
| `CREATE VIEW [schema.]name` | `class` | `[schema.]name` |
| `CREATE FUNCTION [schema.]name(...)` | `function` | `[schema.]name` |
[...]

PR #115, opened 2026-04-27
feat(graph): sql_refs — SQL string-literal call sites (pairs with #95)

andreinknv:
> **🔄 Rebased onto post-refactor patterns (sql-refs (SQL string-literal call sites))**
> This PR has been rebased to use the per-thing file layout established by refactor PRs #116/#117/#118/#119. Behavior shipped is identical to the original; only the internal code shape changed.
>
> **What this PR now ships:**
>
> - **`src/index-hooks/sql-refs.ts`** — registered `IndexHook` (atomic replace on indexAll, incremental on sync)
> - **`src/db/migrations/007-sql-refs.ts`** — `sql_refs` table
> - **`src/sql-refs/index.ts`** — pure regex extractor with comment-strip + SQL-keyword pre-filter
> - **`src/mcp/tools/sql.ts`** — `SQL_TOOL: ToolModule` for `codegraph_sql`
> - `src/index-hooks/registry.ts`, `src/db/migrations/index.ts`, `src/mcp/tools/registry.ts`, `src/mcp/tools/types.ts` — registry entries
> - `src/db/queries.ts`, `src/db/schema.sql`, `src/types.ts`, `src/default-config.ts`, `src/config.ts`, `src/index.ts`, `src/mcp/tools.ts` — methods + handler
> - `__tests__/sql-refs.test.ts`
>
> **What you'll see in the diff:** until #116-#119 land, the diff against `main` includes the refactor commits as part of the branch history. After they merge, GitHub auto-shrinks this PR's diff to just the files above. Full context: #120.
>
> ---

## Summary

Extracts SQL string-literal references to tables (`SELECT FROM users`, `INSERT INTO orders`, `UPDATE x SET`, `DELETE FROM y`, `CREATE TABLE z`) from app code into a new `sql_refs` table. Surfaces via a new `codegraph_sql` MCP tool to answer "what code touches the `users` table?" and "what tables does this codebase access?".

## Pairs with #95 — standalone v1 here

This is the call-site half of the strongest item from a recent feature spike. Once #95 (SQL language extractor) lands upstream and SQL DDL nodes (tables, columns) are indexed, `table_name` strings can be joined against those nodes for full cross-language navigation: "callers of the `users` table" returns both the DDL definition AND every app-code call site. Until #95 merges, `sql_refs` is queryable on its own as a "what code touches table X" index — already useful, just not yet linked to the schema graph.

## Why a separate table, not graph nodes/edges

Tables aren't declared in code that the existing extractors parse — they live in `.sql` migration files that #95 will own. Treating SQL refs as a side-table (mirror of `config_refs` from #114) keeps the core graph clean while still enabling the agent-useful queries.

## Empirical signal on this repo's own history

After the false-positive filter is tightened: **117 refs, 31 distinct tables**. Top results match codegraph's actual SQLite schema:

| Table | Reads | Writes | DDL |
|---|---:|---:|---:|
| `nodes` | 16 | 3 | 0 |
| `unresolved_refs` | 10 | 5 | 0 |
| `files` | 10 | 2 | 0 |
| `edges` | 9 | 2 | 0 |
| `sql_refs` | 4 | 6 | 1 |
| `schema_versions` | 3 | 1 | 0 |
| `project_metadata` | 2 | 1 | 0 |

Plus `users` from test fixtures. Beats grep because the regex requires a SQL keyword prefix (FROM/JOIN/INTO/UPDATE/DELETE), filtering out `const users = ...` style false matches.

## What's added

- **`src/sql-refs/index.ts`** — per-language regex catalogue covering TS/JS, Python, Go, Java/Kotlin, Ruby, Rust, PHP, C#. Captures backtick (MySQL), double-quoted (Postgres), and unquoted identifiers. Drops schema prefixes (`public.users` → `users`).
- **Three-layer false-positive defense:** (a) comment-stripping per language (`//` + `/* */` for C-family, `#` for Python/Ruby) BEFORE the regex sees the line; (b) pre-filter requires both a quote AND a SQL verb; (c) per-match check requires a quote before the matched position.
- **Migration v4** — `CREATE TABLE IF NOT EXISTS sql_refs` with case-insensitive index on `table_name`, FK cascade-deletes on node removal. Defensive `ensureSqlRefsTable()` at every persistence path covers the migration-collision case with #112/#113/#114.
- **`runSqlRefsPass`** — atomic `replaceAllSqlRefs` for `indexAll` (clear + reinsert in one transaction). Incremental on `sync` with delete-then-reinsert per file. Per-pass enclosing-function resolver cache.
- **`codegraph_sql` MCP tool** — no-arg form lists top-N tables with read/write/ddl counts; `table` arg lists call sites with enclosing function; optional `op` filter.
- **Config flag**: `enableSqlRefs` (default true).

## Tests

407 pass total, **27 new**:

- Parser core (12): FROM/JOIN/INSERT/UPDATE/DELETE/CREATE/ALTER/DROP, backtick + double-quote idents, schema-qualified, per-line multi-op, no-keyword-prefix rejection, line-number 1-indexed, resolver threading, missing file
- **Comment-stripping (4)**: rejects line-comment-with-quoted-SQL (`// example: SELECT FROM the docs`), rejects same-line block-comment-with-quoted-SQL, keeps real SQL with trailing comment, Python `#` comments
- End-to-end (8): basic + enclosing function, reverse view, case-insensitive lookup, `enableSqlRefs:false`, incremental sync replace, **regression for empty-rows data-corruption** (file edited to remove last SQL ref), deletion sweep, **v4-collision defense**

## Reviewer pass

Independent reviewer ran once. Three **REQUEST_CHANGES** + two **INFO** addressed:

- **False-positive class** (`// "SELECT FROM the docs"` style): added per-language comment-stripping BEFORE the pre-filter. Tests assert the rejection.
- **Atomic clear+reinsert** for `indexAll` path: new `replaceAllSqlRefs` wraps both ops in one transaction.
- Test gap for the false-positive class: explicit test cases for line-comment, block-comment, and Python `#` shapes.
- Removed unused `MERGE`/`REPLACE` from the pre-filter verb list.
- Documented the DELETE-produces-read+write behavior on `getSqlTables`.

## Coexistence with #112 / #113 / #114

All four PRs target migration v4. Whichever lands second/third/fourth renumbers their migration in the array — standard rebase. `ensureSqlRefsTable()` at every persistence path creates the table on demand as insurance.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `npm test` — 407/407 passing
- [x] End-to-end smoke on this repo: 117 refs, 31 distinct tables, top results match codegraph's actual schema
- [x] False-positive smoke: confirmed `// "from the database"` style comments produce zero refs (vs. 33+ before the comment-stripper)
- [x] Migration v3 → v4 round-trip verified
- [x] Reviewer agent: REQUEST_CHANGES → APPROVE after fixes (with regression tests)


andreinknv:
**Heads-up: this branch has been force-pushed.** This PR's contents have been rebased to use the per-thing file layout established by refactor PRs #116-#119. Functionality is identical to the original; the internal file shape now matches the post-refactor architecture so this PR can be merged after the refactors with zero conflicts.

The current diff against main looks larger than the feature alone because it transitively includes the 4 refactor commits + any prerequisite PRs. That diff auto-shrinks as the prerequisites merge in the order documented in #120. See https://github.com/colbymchenry/codegraph/issues/120#issuecomment-4331117018 for the full rebase summary.

colbymchenry:

PR #111, opened 2026-04-26
feat: LLM symbol summaries, semantic search, RAG Q&A, dir/role/dead-code/naming + agent-as-LLM bridge

andreinknv:
> **🔄 Rebased onto post-refactor patterns + empirically validated against a real local LLM (qwen2.5-coder:32b via Ollama)**
>
> See [Validation results](#validation) below. tl;dr: 787/787 tests pass, agent-bridge tier works without any LLM endpoint, summarizer generates real high-quality summaries (sample verified), dead-code judge runs but is bottlenecked by the underlying graph-signal noise (LLM correctly returns "uncertain" rather than risking false-positive deletes).
>
> ---

## Summary

Adds LLM-mediated tier-3 enrichment on top of the deterministic graph: symbol summaries, embeddings + semantic search, directory summaries, role classification, dead-code judging, change-intent, naming check, and an **agent-bridge** so projects without a local LLM can still benefit by having the calling agent do the LLM work.

### Migrations (file-based pattern, post-refactor #118)

| File | Adds |
|---|---|
| `011-symbol-summaries.ts` | `symbol_summaries` table |
| `012-summary-embeddings.ts` | `embedding BLOB` + `embedding_model` columns |
| `013-directory-summaries.ts` | `directory_summaries` table |
| `014-summary-roles.ts` | `role` + `role_model` columns |

### MCP tools (per-tool ToolModule pattern, post-refactor #117)

| Tool | Purpose |
|---|---|
| `codegraph_pending_summaries` | Agent-bridge: pull batch of symbols needing summaries |
| `codegraph_save_summaries` | Agent-bridge: persist with content_hash re-validation |
| `codegraph_dead_code` | Graph signal + LLM judge for dead-code candidates |
| `codegraph_role` | Filter symbols by classified role (api_endpoint, business_logic, …) |
| `codegraph_module` | Directory-level synthesis paragraph |
| `codegraph_ask` | Hybrid-retrieve + chat-model RAG over the codebase |
| `codegraph_similar` | Semantic-similarity search via embeddings |

### Infrastructure (`src/llm/`)

11 new modules: `client.ts`, `summarizer.ts`, `embeddings.ts`, `ask.ts`, `dir-summarizer.ts`, `classifier.ts`, `change-intent.ts`, `dead-code.ts`, `naming.ts`, `agent-bridge.ts`, `detect.ts` (auto-discovers Ollama / LM Studio / vLLM / llama.cpp on common ports).

## Validation

[...]

PR #110, opened 2026-04-26
feat(mcp): codegraph_review_context — structured PR-review context from a diff

andreinknv:
> **🔄 Rebased onto post-refactor patterns (review-context MCP tool)**
> This PR has been rebased to use the per-thing file layout established by refactor PRs #116/#117/#118/#119. Behavior shipped is identical to the original; only the internal code shape changed.
>
> **What this PR now ships:**
>
> - **`src/mcp/tools/review-context.ts`** — `REVIEW_CONTEXT_TOOL: ToolModule`
> - **`src/review/index.ts` + `src/review/diff-parser.ts`** — pure diff→context builder
> - `src/mcp/tools/registry.ts` — 2 lines
> - `src/mcp/tools/types.ts` — 1 `HandlerKey` entry
> - `src/mcp/tools.ts` — `handleReviewContext` method on `ToolHandler` + JSON-cap serializer
> - `src/index.ts` — `buildReviewContext` public method
> - `__tests__/review-context.test.ts`
>
> **What you'll see in the diff:** until #116-#119 land, the diff against `main` includes the refactor commits as part of the branch history. After they merge, GitHub auto-shrinks this PR's diff to just the files above. Full context: #120.
>
> ---

## Summary

Adds a new MCP tool **`codegraph_review_context`** that takes a unified diff and returns structured review context for an LLM-driven PR reviewer. Codegraph becomes a substrate for Greptile/CodeRabbit-style products **without itself doing the synthesis** — the LLM consumer (Claude Code, ultrareview, etc.) writes the review comments, decides severity, and posts to GitHub.

## What it returns (JSON)

**Per changed file:**
- Status (added / modified / deleted / renamed)
- Hunks (line ranges)
- Affected symbols (line-range overlap with hunks)
- Tests covering the file (via PR #106's `tests` edges; graceful no-op if those edges aren't present)

**Per affected symbol:**
- Signature, docstring
- Top-N callers (PR #108's batch lookups make this fast)
- Top-N callees
- Impact-radius node count

**For deleted files:**
- `brokenIncomingRefs` — every distinct caller of a vanishing symbol (deduped by source name + file + line)

**Co-change warnings (the genuinely novel signal):**
- For each changed file, surface up to N historical co-changers that were **NOT** touched in this PR
- Catches *"you changed `schema.sql` but didn't update `migrations.ts`"* coupling violations that no static analyzer can see
- Graceful degrade if PR #105's `co_changes` table isn't present

## Components

[...]

PR #117, opened 2026-04-27
refactor: per-tool MCP registry — eliminate tools[] + case-switch conflicts

andreinknv:
> **Reviewers:** for a 5-minute review path that surfaces only what to verify, see the **[review checklist comment](https://github.com/colbymchenry/codegraph/pull/117#issuecomment-4336252236)** below. The original description is preserved here for full technical detail.

---

## Summary

**Today every PR adding an MCP tool conflicts on the same two shared lists in `src/mcp/tools.ts`:** the `tools[]` array (the `list_tools` surface) and the `case` switch in `execute()`. After this refactor:

**Adding a new MCP tool:**

1. Drop a file at `src/mcp/tools/<name>.ts` exporting a `<NAME>_TOOL: ToolModule` (definition + handlerKey).
2. Add **one** import line and **one** array entry to `src/mcp/tools/registry.ts`.
3. Implement `handle<Name>(args)` on `ToolHandler` in `tools.ts` and add the new key to `HandlerKey` in `tools/types.ts`.

Step 3 is the only remaining "shared method on a single class" conflict surface. Extracting handler bodies into per-tool files (making step 3 also a single-file addition) is left as a follow-up — the cost/benefit favors landing this incremental win now.

## What's new

| File | Purpose |
|---|---|
| `src/mcp/tool-types.ts` (NEW) | Extracted `ToolDefinition`, `ToolResult`, `PropertySchema`, `projectPathProperty` so per-tool files import without circular deps. |
| `src/mcp/tools/types.ts` (NEW) | `ToolModule`, `HandlerKey` string union, `ToolHandlerLike` interface — `ToolHandler` now `implements ToolHandlerLike`, making dispatch fully compile-time-checked. |
| `src/mcp/tools/<name>.ts` × 9 (NEW) | One file per existing tool: callees, callers, context, explore, files, impact, node, search, status. ~25-30 lines each. |
| `src/mcp/tools/registry.ts` (NEW) | Static-import barrel + lookup helpers + derived `tools[]` array. |
| `src/mcp/tools.ts` | ~200 lines deleted (inline types + tools[] array). `execute()` rewrites the case-switch to a registry lookup with type-safe `this[mod.handlerKey](args)` dispatch. `private async handle*` → public `async handle*` to match the interface. |
| `src/mcp/index.ts` | Tool-existence check switched from linear `tools.find()` scan to O(1) `getToolModule()` Map lookup. |

## Tests

[...]

PR #121, opened 2026-04-28
feat(mcp): emit server-level instructions in initialize response

andreinknv:
## Summary

The MCP `initialize` response can include an `instructions` field that clients (Claude Code, Cursor, opencode, LangChain, OpenAI Agent SDK, …) surface in the agent's system prompt automatically. Today codegraph emits an empty initialize response — agents only see individual tool descriptions, with no overall guidance on how to compose them.

This adds the missing playbook in a new `src/mcp/server-instructions.ts` module, wired into the `initialize` handler.

## Empirical validation (A/B test)

Tested in-session by running the same task two ways and counting tool calls:

**Task:** "Predict the blast radius of changing `extractFromSource` in the codegraph codebase."

| Approach | Calls | Output | Completeness |
|---|---|---|---|
| Path A — naive (no playbook) | `codegraph_search` → `codegraph_callers` → ~5 more recursive walks | ≈7 calls, fragmented output | Partial |
| Path B — playbook-guided | `codegraph_impact("extractFromSource")` | **1 call**, 152 transitive symbols across 14 files | **Complete at depth 2** |

The playbook's mapping `"What would changing this break?" → codegraph_impact` saved ~6 redundant tool calls and produced a more complete answer. The benefit isn't theoretical — without the meta-guidance, the natural agent instinct is to start with `codegraph_search` (the most general-sounding tool) and walk the call graph manually. Tool descriptions alone don't redirect that instinct.

## What the instructions teach the agent

- **Tool selection by intent** — quick map from "what is X" / "how does X work" / "what would changing X break" to the right tool.
- **Common chains** — onboarding (context first), PR review (review_context), refactor planning (search → callers → impact), debugging a regression.
- **Tier discipline** — start at the cheap deterministic tier (search, context, callers, callees, impact, node, explore, files, status), escalate to conditional tools only when their data exists, reach for LLM-mediated tools only when the cheap path doesn't suffice.
- **Agent-bridge tier** — explicit recipe for projects without a local LLM where the agent itself summarizes via `codegraph_pending_summaries` + `codegraph_save_summaries`.
- **Anti-patterns** — don't grep when search exists, don't chain search+node when context covers it, don't query the index immediately after a write.

## Why MCP-level vs CLAUDE.md

CLAUDE.md is a Claude-Code-only convention. The MCP `instructions` protocol field reaches every client. Both can coexist — the existing CLAUDE.md template still covers the Claude-Code-specific Explore-agent pattern. This PR adds the universal playbook on top.

## Why a separate PR

Originally considered as part of #111 (LLM tools), but pulled out because:
- Most users won't run a local LLM, but everyone benefits from tool-selection guidance for the deterministic tools.
- The instructions reference tools that exist on `main` today; they don't presume #110/#112-#115/#111 have landed. After those merge, the relevant sections of the guidance simply start applying.

## Per-language guidance — intentionally not included

Considered (and explicitly rejected): per-language sections like "in Python, callers includes decorators." Reasons:
1. Token cost on every session for content irrelevant ~80% of the time (codegraph supports 19+ languages but typical sessions touch 1-3).
2. The tools themselves are language-agnostic; result *shape* differs per-language but tool *usage* doesn't.
3. Where language matters (`codegraph_sql`, `codegraph_config`), the tool description already self-documents.
4. Project-specific patterns belong in project-local CLAUDE.md, not the universal MCP instructions.

If we ever want this, the principled implementation is dynamic per-project tailoring at initialize time (only emit the SQL section if the project has SQL nodes, etc.). Out of scope here.

## Test plan

- [x] `npx vitest run` clean (no test changes — the JSON-RPC initialize response is structurally compatible)
- [x] A/B test (above) — validated the playbook reduces tool calls on a representative task

## Files changed

| File | Change |
|---|---|
| `src/mcp/server-instructions.ts` | New module (~75 lines, mostly the instructions string) |
| `src/mcp/index.ts` | 1 import + 1 line in the initialize result |

### DIFF
--- a/__tests__/mcp-tool-registry.test.ts
+++ b/__tests__/mcp-tool-registry.test.ts
@@ -1,0 +1,79 @@
+/**
+ * MCP tool registry: structural invariants.
+ *
+ * Guards against the failure mode where a future PR adds a
+ * ToolModule but forgets to implement the matching `handle<Name>`
+ * method on ToolHandler (or vice versa).
+ */
+import { describe, it, expect } from 'vitest';
+import { getToolModules, tools as registryTools } from '../src/mcp/tools/registry';
+import { ToolHandler, tools } from '../src/mcp/tools';
+
+describe('MCP tool registry — single source of truth', () => {
+  it('every tool module has a non-empty name and description', () => {
+    for (const m of getToolModules()) {
+      expect(m.definition.name).toMatch(/^codegraph_[a-z_]+$/);
+      expect(m.definition.description.length).toBeGreaterThan(20);
+    }
+  });
+
+  it('handlerKey is a string starting with "handle"', () => {
+    for (const m of getToolModules()) {
+      expect(m.handlerKey).toMatch(/^handle[A-Z][A-Za-z]+$/);
+    }
+  });
+
+  it('every registered tool has a corresponding ToolHandler method', () => {
+    const handler = new ToolHandler(null);
+    for (const m of getToolModules()) {
+      const fn = (handler as unknown as Record<string, unknown>)[m.handlerKey];
+      expect(typeof fn).toBe('function');
+    }
+  });
+
+  it('exported `tools` array exactly mirrors the registry', () => {
+    const fromRegistry = registryTools.map((t) => t.name).sort();
+    const fromExport = tools.map((t) => t.name).sort();
+    expect(fromExport).toEqual(fromRegistry);
+  });
+
+  it('all 9 main-line tools are registered (regression guard)', () => {
+    const expected = [
+      'codegraph_callees',
+      'codegraph_callers',
+      'codegraph_context',
+      'codegraph_explore',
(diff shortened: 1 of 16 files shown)


colbymchenry:
Closing — same situation as #122. The PR is also stacked on #117's tool-registry refactor (the per-tool `tools/<name>.ts` files in the diff are #117's), AND the instructions string itself references a lot of tools that aren't on main (`codegraph_biomarkers`, `codegraph_coverage`, `codegraph_hotspots`, `codegraph_config`, `codegraph_sql`, the LLM stack). Roughly half the playbook is for features not in the codebase.

The core idea is good — telling agents to reach for `codegraph_impact` instead of walking callers manually is a real win. Going to land a scoped version directly: only the tools that exist on main, no `(PR #X, when present)` references, ~40 lines instead of 75.

PR #122, opened 2026-04-28
perf(db): drop redundant idx_edges_source and idx_edges_target

andreinknv:
## Summary

Drops two redundant secondary indexes from the `edges` table:

- `idx_edges_source` — fully covered by `idx_edges_source_kind` via SQLite's left-prefix scan.
- `idx_edges_target` — fully covered by `idx_edges_target_kind`.

These indexes pre-date the `(source, kind)` / `(target, kind)` composites added later. Once the wider indexes were introduced, the narrow ones became dead weight on every write and every disk page — but they were never removed.

## Empirical validation

Measured on a synthesized 50K-node / 250K-edge SQLite database that mirrors a realistic mid-size codebase. Reproducer: [`scripts/spikes/spike-edge-indexes.mjs`](scripts/spikes/spike-edge-indexes.mjs).

| Metric | Baseline (with redundant) | Stripped | Δ |
|---|---|---|---|
| DB size | 34.7 MB | 27.0 MB | **-22.2%** |
| Bulk insert (250K edges) | 590 ms | 431 ms | **1.37× faster** |
| `WHERE source = ?` latency | reference | 0.88× | no regression |
| `WHERE target = ?` latency | reference | 0.95× | no regression |

EXPLAIN output on the stripped DB confirms the kind-prefixed indexes still cover source-only / target-only queries:

```
SEARCH edges USING COVERING INDEX idx_edges_source_kind (source=?)
```

Run yourself with `node scripts/spikes/spike-edge-indexes.mjs`.

## Why this is safe

| Query shape | Old plan | New plan |
|---|---|---|
| `WHERE source = ?` | `idx_edges_source` | `idx_edges_source_kind` (left-prefix, covering) |
| `WHERE target = ?` | `idx_edges_target` | `idx_edges_target_kind` (left-prefix, covering) |
| `WHERE source = ? AND kind = ?` | `idx_edges_source_kind` | `idx_edges_source_kind` (unchanged) |
| `WHERE target = ? AND kind = ?` | `idx_edges_target_kind` | `idx_edges_target_kind` (unchanged) |

SQLite's query planner uses a B-tree's leading column(s) for prefix matches, so the composite indexes are a strict superset of the dropped ones for these queries.

## Migration

Adds schema migration **v4** (`CURRENT_SCHEMA_VERSION` 3 → 4):

```sql
DROP INDEX IF EXISTS idx_edges_source;
DROP INDEX IF EXISTS idx_edges_target;
```

[...]

PR #118, opened 2026-04-27
refactor: file-based migrations — eliminate silent version-collision bugs

andreinknv:
> **Reviewers:** for a 5-minute review path that surfaces only what to verify, see the **[review checklist comment](https://github.com/colbymchenry/codegraph/pull/118#issuecomment-4336283128)** below. The original description is preserved here for full technical detail.

---

refactor: file-based migrations — eliminate version-collision bug class

Today every PR adding a schema migration claims
`CURRENT_SCHEMA_VERSION = next` AND adds an array entry to
`migrations: Migration[]` in src/db/migrations.ts. Two PRs both
claiming the same version resolve as: "second PR's v4 silently
no-ops on existing DBs" — a real silent-data-loss bug class
(PR #113's reviewer caught one).

After this refactor:

  Adding a new schema migration:
  1. Pick the next free 3-digit prefix (`git ls-files
     'src/db/migrations/[0-9]*.ts'` shows what's taken).
  2. Create `src/db/migrations/<NNN>-<short-name>.ts` exporting a
     `MIGRATION: MigrationModule` (description + up).
  3. Add one import line and one entry to
     `src/db/migrations/index.ts`'s REGISTERED_MODULES array.

Two PRs both creating `004-foo.ts` collide on the FILESYSTEM —
the maintainer sees it instantly. No more silent skipped
migrations.

## What's new

- `src/db/migrations/types.ts` — `MigrationModule { description,
  up }` and `Migration extends MigrationModule { version }`.
- `src/db/migrations/002-project-metadata.ts` — extracted v2
  body verbatim.
- `src/db/migrations/003-lower-name-index.ts` — extracted v3
  body verbatim.
- `src/db/migrations/index.ts` — central registry. Static-imports
  each migration, parses the version FROM THE FILENAME (no
  hand-typed version field that can drift), enforces strict
  `NNN-kebab-name.ts` shape, validates uniqueness/sort at module
  load (throws loudly on collision), exposes ALL_MIGRATIONS and
  CURRENT_SCHEMA_VERSION.
- `src/db/migrations.ts` — refactored to a thin runner. Same
  exported surface (CURRENT_SCHEMA_VERSION, getCurrentVersion,
  runMigrations, needsMigration, getPendingMigrations,
  getMigrationHistory, Migration type) — every existing import
  keeps working unchanged.
- `__tests__/migrations-registry.test.ts` — 8 invariant tests:
  registry non-empty, versions unique + strictly ascending,
[...]

Issue #120, opened 2026-04-27
Merge-order guide for the open PR backlog (4 refactor PRs unblock the rest)


andreinknv:
# Maintainer guide: how to merge the open PR backlog with minimal friction

> **Update — April 28, 2026:** 14 of my PRs have been pre-rebased onto the post-refactor patterns. The agent-callable code-intelligence layer is fleshing out: **#124 (per-symbol coverage from lcov reports)** and **#125 (static-analysis biomarkers + Code Health score)** add two new MCP tools so an AI agent can ask "is this function tested?" and "is this function risky to change?" with a single structured query, instead of reading the source. Both stack on **#118** (file-based migrations); #125 also runs as an `IndexHook`. **#122** remains a tiny self-contained perf win in monolithic shape that targets `main` directly. See the [rebase summary comment](https://github.com/colbymchenry/codegraph/issues/120#issuecomment-4331117018) for the full list. The rebase templates below remain authoritative documentation for external PRs (#58, #91, #89) and for any future PR that needs to land on top of the refactors.

---

There are ~24 open PRs touching `colbymchenry/codegraph`. Many of them collide on the same handful of monolithic data structures (lists, switches, version numbers, hardcoded method additions). I've opened **4 refactor PRs** that eliminate those collision surfaces. This document tells you the order to merge things in to minimize manual conflict-resolution work, and shows you exactly what the rebase of every existing open PR looks like.

---

## Easiest path for the maintainer: 12 clean merges in 4 rounds

I walked this path end-to-end against a fresh `colbymchenry:main` checkout three times to confirm — twice with intermediate rebases that broke things in subtle ways, then once cleanly. **12 explicit merges, zero hand-resolution. 829/829 tests pass at the tip.** Final state matches [`andreinknv:integration/all-prs`](https://github.com/andreinknv/codegraph/tree/integration/all-prs) (22 MCP tools, 8 hooks, 18 migrations).

### Round 1 — five PRs clean against current `main`, any order

| PR | What |
|----|------|
| #116 | refactor: per-language registry |
| #117 | refactor: per-tool MCP registry |
| #118 | refactor: file-based migrations |
| #119 | refactor: index-hook framework |
| #127 | fix(gitignore): anchor `coverage/` to repo root |

### Round 2 — three PRs clean once Round 1 is in

| PR | Why it's now clean |
|----|--------------------|
| #92  | HCL/Terraform language — needs #116 |
| #94  | R language — needs #116 |
[...]

PR #105, opened 2026-04-26
feat(cochange): file-level co-change graph mined from git history

andreinknv:
> **🔄 Rebased onto post-refactor patterns (Co-change graph)**
> This PR has been rebased to use the per-thing file layout established by refactor PRs #116/#117/#118/#119. Behavior shipped is identical to the original; only the internal code shape changed.
>
> **What this PR now ships:**
>
> - **`src/index-hooks/cochange.ts`** — registered `IndexHook` (afterIndexAll + afterSync)
> - **`src/db/migrations/010-co-changes.ts`** — `co_changes` table
> - **`src/cochange/index.ts`** — pure miner module
> - `src/index-hooks/registry.ts` — 2 lines
> - `src/db/migrations/index.ts` — 2 lines
> - `src/db/queries.ts`, `src/db/schema.sql`, `src/types.ts`, `src/default-config.ts`, `src/config.ts`, `src/index.ts` — query methods + config flag
> - `__tests__/cochange.test.ts` — full miner + integration suite
>
> **What you'll see in the diff:** until #116-#119 land, the diff against `main` includes the refactor commits as part of the branch history. After they merge, GitHub auto-shrinks this PR's diff to just the files above. Full context: #120.
>
> ---

## Summary

Adds a file-level **co-change graph** mined from `git log`. Surfaces coupling between files that historically change together — sibling extractors that share patterns, tests that assert schema state, config files coupled to the code that reads them. This information is **invisible to static analysis** but recoverable from git in seconds with **zero new dependencies**.

## Empirical results (codegraph's own history)

| File pair | Jaccard | Why coupled |
|---|---:|---|
| `src/extraction/languages/csharp.ts` ↔ `java.ts` | 0.75 | Sibling language extractors share patterns; no import relation |
| `__tests__/foundation.test.ts` ↔ `src/db/schema.sql` | 0.57 | Test asserts schema version; coupled by intent, not import |
| `src/db/migrations.ts` ↔ `src/db/schema.sql` | 0.57 | Schema changes always need a matching migration |
| `src/installer/config-writer.ts` ↔ `src/installer/index.ts` | 0.50 | Two files in same module, always change together |

For `src/extraction/index.ts`, the top co-changer is `src/index.ts` (J=0.41, the wrapper class). For `src/db/schema.sql`, the top co-changers are the test that asserts schema version, the migrations file, and the db index. The signal is meaningful and matches what a maintainer would intuit.

## What's added

[...]

PR #123, opened 2026-04-28
perf(db): split embeddings + in-memory similarity cache + drop redundant co_changes index

andreinknv:
> **🔗 Stacked PR — depends on #111 (LLM symbol summaries) and #118 (file-based migrations)**
>
> Until those merge, this PR's diff transitively includes their commits. The intrinsic change for *this* PR is just the new files in `src/db/migrations/015-*.ts`, `src/db/migrations/016-*.ts`, the `EmbeddingCache` additions in `src/llm/embeddings.ts`, the integration in `src/index.ts`, the queries.ts query rewrites, and the spike script + tests — see "Files changed" below. After #111 + #118 land, the diff auto-cleans to just those.
>
> Companion to #122 (`perf/drop-redundant-edge-indexes`) which lands the analogous `edges`-side optimization on `main` directly with no dependencies.

## Summary

Three measurable wins on the LLM-tier data layer, validated by spike *before* implementing each:

| | Change | Win | Spike file |
|---|---|---|---|
| **F2** | Drop `idx_co_changes_a` | Covered by `(file_a, file_b)` PK | (analogous to #122) |
| **G** | Split embeddings into `symbol_embeddings` table | **3.22× faster** summary-only scans | `scripts/spikes/spike-embedding-split.mjs` |
| **H** | In-memory `EmbeddingCache` for similarity search | **4.4× faster** top-K cosine | `scripts/spikes/spike-embedding-split.mjs` |

## Empirical validation

Run yourself: `node scripts/spikes/spike-embedding-split.mjs`. Output on a 50K-summary / 768d-embedding synthetic DB:

```
--- Spike G: storage layout (inline vs split) ---

  inline DB: 196.5 MB
  split  DB: 204.0 MB

  Test: scan summaries by role (common path)
  inline: 46ms avg over 50 queries
  split : 14ms avg over 50 queries
  Δ summary-only: split is 3.22× faster

  Test: scan summaries WITH embedding (rare path)
  inline (single table)   : 71ms avg over 50 queries
  split  (join required)  : 80ms avg over 50 queries
  Δ summary+embedding: 1.12× cost penalty for split

--- Spike H: in-memory embedding cache ---

  cold (per-query SQLite fetch + decode): 104ms avg over 20 queries
  warm (in-memory Float32Array matrix)  : 24ms avg over 20 queries
  Δ similarity search: 4.4× speedup with in-memory cache
```

The 1.12× cost on summary+embedding scans is dwarfed by the 3.22× win on summary-only scans, which dominate by ~50× in real usage (FTS-anchor lookups, role filters, freshness checks all read summaries-without-embeddings).

### F2 — Migration 015: drop `idx_co_changes_a`
- `co_changes` has `PRIMARY KEY (file_a, file_b)`, which automatically creates a B-tree leading on `file_a`. SQLite covers `WHERE file_a = ?` lookups via that PK index — the standalone `idx_co_changes_a` was redundant.
- `idx_co_changes_b` (on `file_b` alone) is kept because the PK leads with `file_a`, so it cannot serve `WHERE file_b = ?` lookups.
- Fresh-DB schema (`src/db/schema.sql`) updated to skip `idx_co_changes_a` and dedupe a pre-existing duplicate of `idx_co_changes_b`.

### G — Migration 016: split embeddings into `symbol_embeddings`

```sql
CREATE TABLE symbol_embeddings (
    node_id TEXT PRIMARY KEY,
    embedding BLOB NOT NULL,
    embedding_model TEXT NOT NULL,
    FOREIGN KEY (node_id) REFERENCES symbol_summaries(node_id) ON DELETE CASCADE
);
INSERT OR IGNORE INTO symbol_embeddings (node_id, embedding, embedding_model)
  SELECT node_id, embedding, embedding_model
  FROM symbol_summaries
  WHERE embedding IS NOT NULL AND embedding_model IS NOT NULL;
DROP INDEX IF EXISTS idx_summaries_embedding_model;
ALTER TABLE symbol_summaries DROP COLUMN embedding;
ALTER TABLE symbol_summaries DROP COLUMN embedding_model;
```

- Requires SQLite 3.35+ for `ALTER TABLE DROP COLUMN`. Both `better-sqlite3` and `node-sqlite3-wasm` ship with newer versions, so this is safe.
- `queries.ts` methods (`getEmbeddableSummaries`, `getAllEmbeddings`, `upsertSymbolEmbedding`) updated to use the new table.
- `clear()` and `clearCoChanges()` extended to wipe both tables (the FK cascade would handle it, but explicit is safer if foreign-key enforcement gets disabled).

### H — `EmbeddingCache` in `src/llm/embeddings.ts`

```ts
export class EmbeddingCache {
  get(fetcher: EmbeddingFetcher, model: string): CachedEmbeddings;
  invalidate(): void;
}
```

- Decodes every embedding into a flat `Float32Array` matrix once per (model, generation).
- New `topKByCosineMatrix(query, matrix, ids, dim, k)` operates directly on the flat layout.
- Owned by the `CodeGraph` instance.
- Invalidated on:
  - `indexAll` end (when `filesIndexed > 0`)
  - `sync` end (when files added/modified/removed)
  - After `embedAllSummaries` (when `generated > 0`)
  - `clear()` (table was emptied)
- Mismatched-dim rows are skipped on rebuild — produces a packed matrix with no sparse holes.

## Independent review

Reviewed by an independent reviewer agent (read-only, fresh context). Surfaced 3 issues, all addressed in the same diff:

1. **`CodeGraph.clear()` was missing a cache invalidation** — fixed: added `this.embeddingCache.invalidate()`.
2. **Duplicate `idx_co_changes_b` in `schema.sql`** (pre-existing, harmless due to `IF NOT EXISTS`) — fixed: deduped.
3. **`EmbeddingCache` left sparse `undefined` holes in `ids` on dim-mismatch** — fixed: rewrote to push aligned entries instead of pre-allocating, with a regression test.

## Test plan

- [x] `npx vitest run` — **794/794 tests pass**
- [x] New tests: `__tests__/migrations-015-016.test.ts` (upgrade + fresh-DB paths for both migrations)
- [x] New tests in `__tests__/embeddings.test.ts`:
  - `topKByCosineMatrix` matches `topKByCosine` on the same data
  - `EmbeddingCache` hit/miss/invalidate
  - Cache returns empty result without re-fetching
  - Cache skips dim-mismatched rows
- [x] Spike script (`scripts/spikes/spike-embedding-split.mjs`) reproduces the headline numbers

## Files changed

| File | Change |
|---|---|
| `src/db/migrations/015-prune-co-changes-index.ts` | New: drop `idx_co_changes_a` |
| `src/db/migrations/016-split-symbol-embeddings.ts` | New: split embeddings into dedicated table |
| `src/db/migrations/index.ts` | Register both new migrations |
| `src/db/schema.sql` | Fresh-DB schema reflects new layout; dedupe `idx_co_changes_b` |
| `src/db/queries.ts` | `getAllEmbeddings`, `upsertSymbolEmbedding`, `getEmbeddableSummaries` use `symbol_embeddings` |
| `src/llm/embeddings.ts` | `EmbeddingCache` + `topKByCosineMatrix` |
| `src/index.ts` | Cache integrated into `searchHybrid` and `findSimilar`; invalidations wired |
| `__tests__/migrations-015-016.test.ts` | New: focused migration tests |
| `__tests__/embeddings.test.ts` | New: cache + matrix tests |
| `__tests__/foundation.test.ts` | Updated schema-version expectation |
| `__tests__/pr19-improvements.test.ts` | Updated schema-version expectation |
| `scripts/spikes/spike-embedding-split.mjs` | New: G + H reproducer |


andreinknv:
**Merge guide:** see #120 for the full backlog merge order. Short version for this PR: ✅ already in post-refactor file-based shape (migrations `015-prune-co-changes-index.ts` + `016-split-symbol-embeddings.ts`). Stacked on **#118** (file-based migrations) + **#111** (LLM features — provides the `symbol_summaries.embedding` column the split migration moves) + **#105** (cochange — provides `co_changes` table the index drop targets). Until those three merge, this PR's diff transitively includes their commits — that's expected and auto-cleans on merge. Spike-validated (`scripts/spikes/spike-embedding-split.mjs`): 3.22× faster summary-only scans, 4.4× faster top-K cosine search. Independent reviewer pass caught and fixed 3 issues before opening (missing cache invalidation, schema dupe, sparse-array hole).

andreinknv:
## Scaling check — do these wins hold past 50K nodes?

Reasonable concern given the original spike was at 50K only. Re-ran the same measurements at **100K** and **250K** to characterise the trend, and the speedup ratios stay flat (the optimizations are structural, not threshold-sensitive):

| | 50K nodes | 100K nodes | 250K nodes | **1M (linear projection)** |
|---|---|---|---|---|
| **G: summary-only scan** | 48 → 14 ms (**3.47×**) | 93 → 38 ms (**2.43×**) | 256 → 106 ms (**2.42×**) | ~1024 → 424 ms (~2.4×) |
| **H: similarity cold** | 107 ms | 226 ms | 609 ms | **~2.4 s/query** |
| **H: similarity warm** | 25 ms | 49 ms | 136 ms | **~540 ms/query** |
| **H: cache speedup** | 4.35× | 4.65× | 4.49× | **~4.5×** |

> The 3.47× → 2.4× drop on G between 50K and 100K is the OS page cache effect — at 50K the inline DB still fits in cache, so both inline and split run roughly memory-bound; past that the structural advantage (no BLOB on the summary page chain) settles at a steady ~2.4×.

**The interesting threshold for H is absolute latency, not ratio.** Without the cache, similarity search crosses the user-perceptible 1-second mark somewhere around **250K-500K nodes** and grows linearly past that. With the cache it stays comfortably interactive even at 1M (~540 ms = roughly the same as a cold search at 250K). For codegraph users on million-node monorepos, this is the difference between "snappy" and "noticeable wait on every query."

**Reproducer:** the original `scripts/spikes/spike-embedding-split.mjs` plus a parameterised version that takes a NODES list. Happy to commit the parameterised one if useful.

**What this doesn't change:** the F2 win (`idx_co_changes_a` drop) is unrelated to scale — it's just dead weight removed.

**Beyond ~500K nodes** the next-step optimisations (`sqlite-vec` once HNSW ships, per-language partitioning) start to matter more than these three. But these three are the right thing to land first because they're zero-dependency and pay off at every scale.

colbymchenry:

PR #124, opened 2026-04-28
feat(coverage): per-symbol code coverage from lcov reports + codegraph_coverage MCP tool

andreinknv:
> **🔗 Stacked on #118 (file-based migrations).** Until that lands, this PR's diff transitively includes the refactor commit. The intrinsic addition for *this* PR is the new files in `src/coverage/`, `src/db/migrations/018-*.ts`, `src/mcp/tools/coverage.ts`, the QueryBuilder methods, the CLI command, and the tests — see "Files changed" below. After #118 lands, the diff auto-cleans.

## Summary

Per-symbol code coverage from external CI artifacts (lcov today; cobertura/jacoco are deliberate scope cuts for v1). Joining coverage to the graph's existing signals unlocks the killer query an AI agent has been missing:

```
mode=ranked, minCentrality=0.001, maxPct=0.5
```

— "high-impact code that's barely tested" in one MCP call. Currently the agent has to ask for centrality, then ask for callers, then guess about test coverage by reading test file names. Now it's one row.

## What ships

| Piece | Purpose |
|---|---|
| `src/coverage/lcov.ts` | Streaming lcov.info parser (handles DA/BRDA, CRLF, missing trailing `end_of_record`, Istanbul's `-1` excluded-line sentinel) |
| `src/coverage/index.ts` | Ingestion orchestrator: maps report paths to indexed nodes (with monorepo suffix-matching), rolls hits up to symbol spans |
| Migration 018 | `node_coverage` table keyed on `(node_id, source)` so multiple test suites coexist |
| `getNodeCoverage`, `getCoverageRanked`, `getCoverageStats`, `clearCoverageSource` | QueryBuilder methods |
| `codegraph_coverage` MCP tool | Three modes: `symbol` (single lookup), `ranked` (worst-coverage-first), `stats` (project rollup) |
| `codegraph coverage <report>` CLI | Ingest with `--source <name>` and `--clear` flags |

## The agent-first angle

Most coverage tooling is dashboards-for-humans. Codegraph's positioning is *code intelligence as MCP tools agents call.* That makes the join with existing signals the differentiator:

```sql
-- "What untested function does the most callers depend on?"
SELECT n.name, n.centrality, c.covered_lines, c.total_lines
FROM nodes n
JOIN node_coverage c ON c.node_id = n.id
WHERE n.centrality > 0.001
  AND (CAST(c.covered_lines AS REAL) / c.total_lines) <= 0.5
ORDER BY n.centrality DESC LIMIT 10;
```

[...]

PR #127, opened 2026-04-28
fix(gitignore): anchor "coverage/" to repo root (unblocks src/coverage/, packages/*/coverage/)

andreinknv:
## Summary

The `coverage/` rule on line 21 of `.gitignore` is unanchored — it matches any `coverage/` directory in the tree, not just the test-output one at the repo root. This is a real footgun: when adding a `src/coverage/` module (or `packages/*/coverage/` in a monorepo), `git add` silently drops the files with no warning.

This bit me on PR #124 — the `src/coverage/` module was written and tested locally, but never made it into the commit because `git add src/coverage/` was silently swallowed. The PR shipped with `__tests__/coverage.test.ts` importing a module that didn't exist; whoever rebases or merges it hits a build failure that's confusing to debug.

The fix anchors the rule to the repo root: `/coverage/`. This still ignores the test-output `coverage/` at root (the original intent) while allowing source paths like `src/coverage/` to be committed normally.

## Test plan

- [x] `git check-ignore -v src/coverage/foo.ts` → no longer matches
- [x] `git check-ignore -v coverage/lcov.info` → still matches the root-level rule
- [x] No code changes; behavior is purely the gitignore semantics

Independent of the issue #120 merge guide — this can land any time and helps every future PR that touches anything named `coverage/` outside the repo root.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,7 @@ dist/
 Thumbs.db
 
 # Test coverage
-coverage/
+/coverage/
 .nyc_output/
 
 # Environment


colbymchenry:
Reviewed and merging. Anchoring to root is the correct fix — unanchored matches at any depth, which is exactly the silent-swallow footgun you flagged. Thanks.

PR #128, opened 2026-04-28
fix(cli): surface lock-acquisition errors and silence Emscripten Aborted() spam

andreinknv:
## Summary

Two small, unrelated CLI cosmetic bugs that show up when the indexer is under load. Both are actively misleading rather than just noisy.

### 1. Lock-acquisition error hidden behind "No files found to index"

`printIndexResult` falls through to `'No files found to index'` whenever `filesIndexed === 0 && filesErrored === 0`. The lock-acquisition path in `indexAll`/`sync` returns `{ success: false, errors: [{ message: 'Could not acquire file lock ...', severity: 'error' }] }`, but `filesErrored` counts only file-level parse failures — so the lock-failure message never gets shown. The user sees "No files found to index" while another indexer is actively running.

Added a check at the top of `printIndexResult` that surfaces the first generic error when `result.success === false`.

### 2. Emscripten Aborted() spam leaking to the parent terminal

When tree-sitter WASM crashes on a pathological file (large generated C/C++ headers, etc.), the parse worker correctly catches the `RuntimeError: memory access out of bounds` and exits with code 1 — but Emscripten itself prints `Aborted()` (and a follow-up `Build with -sASSERTIONS for more info` line) directly to stderr *before* the JS handler runs. Those lines leak to the parent's terminal even though the JS layer recovers cleanly. Re-indexing a repo with WASM-prone files produces dozens of `Aborted()` lines.

Installed a startup-time `process.stderr.write` filter in `parse-worker.ts` that drops only those two specific Emscripten internal lines. Everything we log ourselves (via `console.*` / `parentPort.postMessage`) passes through unchanged.

## Test plan

Verified live against ollama/ollama@v0.22.0 (1.2 GB repo, 1,179 indexed files including vendored llama.cpp / MLX-C):

- [x] `codegraph index` while another indexer is running now shows `Could not acquire file lock - another process may be indexing` (was: "No files found to index")
- [x] Re-index of a repo that had previously triggered WASM-abort lines: 0 `Aborted()` lines (was: 68+)
- [x] `npx tsc --noEmit` passes

### DIFF
--- a/src/bin/codegraph.ts
+++ b/src/bin/codegraph.ts
@@ -265,6 +265,19 @@ type IndexResult = {
 function printIndexResult(clack: typeof import('@clack/prompts'), result: IndexResult, projectPath?: string): void {
   const hasErrors = result.filesErrored > 0;
 
+  // Surface non-file-level failures (e.g. lock-acquisition failure
+  // when another indexer is running) before the file-count branches.
+  // Without this the CLI falls through to "No files found to index",
+  // which is actively misleading — the index DID run, it just couldn't
+  // get the lock.
+  if (!result.success && !hasErrors && result.filesIndexed === 0) {
+    const generic = result.errors.find((e) => e.severity === 'error');
+    if (generic) {
+      clack.log.error(generic.message);
+      return;
+    }
+  }
+
   if (result.filesIndexed > 0) {
     if (hasErrors) {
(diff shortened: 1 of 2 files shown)


colbymchenry:
Reviewed and merging.

The stderr filter is well-scoped to the parse-worker process (no risk of swallowing main-process output) and the comments explicitly call out the per-call-match and substring-exactness caveats — exactly the right things to flag. The CLI fix only kicks in for the precise misreported state. Auto-merges with main, full suite passes. Thanks.

PR #129, opened 2026-04-28
fix(extraction): drop duplicate export-var nodes and honour maxFileSize in bulk path

andreinknv:
## Summary

Two correctness bugs in the core extraction pipeline, surfaced by stress-testing against an adversarial corpus (5k synthetic export-const declarations plus an 8MB single-line file).

### 1. Every `export const` produces two duplicate nodes

`export const X = ...` was producing **two** nodes for the same symbol:
- one with `kind: 'variable'` from `extractExportedVariables`
- one with `kind: 'constant'` from `extractVariable` (invoked when the walker descended into the `export_statement` child)

Stress-testing showed 100% duplication across 5,003 `export const` declarations.

The dedicated `extractVariable` dispatch is the correct one — it picks `kind` from `isConst`, captures the initializer signature, and walks type annotations. The `extractExportedVariables` helper was redundant because each language extractor's `isExported` predicate already walks the parent chain to detect the export wrapper. Removed the `export_statement` branch from the walker dispatch (children are descended into normally) and dropped the private helper.

### 2. `maxFileSize` silently ignored on the bulk-index path

`extractFile()` (single-file API) checked `stats.size > config.maxFileSize`, but the bulk `indexAll()` path read each file's stats and never compared. Vendored generated files (multi-MB headers, minified bundles) were indexed regardless of the user's cap. Mirrored the single-file behaviour: emit a `size_exceeded` warning, count the file as skipped, advance progress, continue.

## Test plan

Verified live against a stress workspace (5,005 synthetic files; 50,000 fns in one 3MB file; 8MB single-line file; 5,000 `export const` declarations):

| Metric | Before | After |
|--------|--------|-------|
| Duplicate var/const node sets | 5,003 (100%) | **0** |
| Files >1MB indexed despite `maxFileSize: 1MB` | 2 | **0** |
| Total nodes | 65,014 | 10,008 |

- [x] `npx vitest run` — **380 passed** (was 374 passed, 6 failed before this fix; the 6 failing tests asserted the duplicate behavior, now updated to match the correct `kind: 'constant'`)
- [x] `npx tsc --noEmit` passes
- [x] Live `codegraph index` against stress corpus produces the expected node counts and `size_exceeded` warnings

### DIFF
--- a/__tests__/extraction.test.ts
+++ b/__tests__/extraction.test.ts
@@ -376,7 +376,7 @@ export const useUIStore = create<UIState>((set) => ({
 `;
     const result = extractFromSource('store.ts', code);
 
-    const varNode = result.nodes.find((n) => n.kind === 'variable' && n.name === 'useUIStore');
+    const varNode = result.nodes.find((n) => n.kind === 'constant' && n.name === 'useUIStore');
     expect(varNode).toBeDefined();
     expect(varNode?.isExported).toBe(true);
   });
@@ -390,7 +390,7 @@ export const config = {
 `;
     const result = extractFromSource('config.ts', code);
 
-    const varNode = result.nodes.find((n) => n.kind === 'variable' && n.name === 'config');
+    const varNode = result.nodes.find((n) => n.kind === 'constant' && n.name === 'config');
     expect(varNode).toBeDefined();
     expect(varNode?.isExported).toBe(true);
   });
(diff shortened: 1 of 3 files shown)


colbymchenry:
Reviewed and merging.

Both bugs are real and the fix is the obviously-right shape — removing the redundant `export_statement` dispatch (the inner declaration's extractor handles it correctly with `isExported` propagating via parent-walk) and mirroring the single-file path's `maxFileSize` check on the bulk path. Net negative LOC, no regressions, auto-merges cleanly with main. Thanks.

PR #130, opened 2026-04-28
feat(resolution): tsconfig path aliases + re-export chain following

andreinknv:
## Summary

Two related correctness improvements that unlock accurate import resolution on modern JS/TS codebases — the single biggest gap in resolution quality identified during the recent quality audit.

### 1. tsconfig/jsconfig path aliases

The resolver had a small hard-coded alias list (`@/`, `~/`, `src/`, `app/`) and **ignored** project-defined `compilerOptions.paths`. Every import through `@components/Foo`, `@lib/utils`, `~api/auth`, etc. on Vite / Next / Nuxt / Nest / NestJS / TanStack-Start projects silently failed to resolve.

Adds `src/resolution/path-aliases.ts`:
- Reads `tsconfig.json`, falls back to `jsconfig.json`
- Honours `compilerOptions.baseUrl` and `compilerOptions.paths`
- Supports the `*` wildcard (the only TS-supported wildcard)
- Multiple replacement targets per alias, tried in tsconfig priority order
- JSONC-tolerant: strips `//` and `/* */` comments + trailing commas before parsing (tsconfigs in the wild routinely contain those, which `JSON.parse` rejects)

`ResolutionContext.getProjectAliases()` lazily loads + caches the result. `resolveAliasedImport` consults it **before** the legacy fallback list, so old projects keep working unchanged.

### 2. Re-export chain following

`import { Foo } from './barrel'` where `barrel.ts` only re-exports (`export { Foo } from './real'` or `export * from './real'`) used to fail because the resolver only looked for declarations *in* the resolved file. The barrel pattern is universal in modern JS/TS — every `index.ts` aggregator broke resolution for everything it forwarded.

Adds:
- `extractReExports(content, language)` — recognises named (`export { a, b as c } from '…'`), wildcard (`export * from '…'`), and namespace (`export * as ns from '…'`) forms
- `ResolutionContext.getReExports(filePath, language)` — per-file cached
- A new recursive `findExportedSymbol(...)` helper with depth cap (8) and visited-set cycle protection

`resolveViaImport` now follows the chain whenever the symbol isn't directly declared in the imported file.

## Test plan

- [x] **Path alias verification** — synthetic project with `@utils/*` and `@lib` aliases:
  ```
  resolved calls from src/main.ts:
    { caller: 'main', callee: 'formatDate', callee_file: 'src/utils/format.ts' }
    { caller: 'main', callee: 'libCore',    callee_file: 'src/lib/index.ts' }
  unresolved calls (should be empty): []
  ```
- [x] **Re-export chain verification** — 3-hop chain `main.ts → all.ts (wildcard) → index.ts (named) → auth.ts (declaration)`:
  ```
  resolved edges from src/main.ts (lands in src/services/auth.ts):
    { caller: 'go', callee: 'signIn', callee_file: 'src/services/auth.ts' }
  unresolved (should be empty for signIn/Session): []
  ```
- [x] `npx vitest run` — **380 passed**, no regressions

### DIFF
--- a/__tests__/resolution.test.ts
+++ b/__tests__/resolution.test.ts
@@ -607,4 +607,75 @@ function main(): void {
       expect(result.stats.total).toBeGreaterThanOrEqual(0);
     });
   });
+
+  describe('tsconfig path aliases', () => {
+    it('resolves an aliased import to the alias-mapped file (not a same-named file elsewhere)', async () => {
+      // Two same-named exports in different directories. Without alias
+      // resolution, name-matcher would pick whichever it finds first;
+      // with alias resolution, the import path uniquely picks one.
+      fs.mkdirSync(path.join(tempDir, 'src/utils'), { recursive: true });
+      fs.mkdirSync(path.join(tempDir, 'src/legacy'), { recursive: true });
+      fs.writeFileSync(
+        path.join(tempDir, 'src/utils/format.ts'),
+        `export function pickMe(): number { return 1; }\n`
+      );
+      fs.writeFileSync(
+        path.join(tempDir, 'src/legacy/format.ts'),
+        `export function pickMe(): number { return 99; }\n`
+      );
+      fs.writeFileSync(
+        path.join(tempDir, 'src/main.ts'),
+        `import { pickMe } from '@utils/format';\nexport function go(): number { return pickMe(); }\n`
+      );
+      fs.writeFileSync(
+        path.join(tempDir, 'tsconfig.json'),
+        JSON.stringify({
+          compilerOptions: {
+            baseUrl: './src',
+            paths: { '@utils/*': ['utils/*'] },
+          },
+        })
+      );
+
+      cg = await CodeGraph.init(tempDir, { index: true });
+      cg.resolveReferences();
+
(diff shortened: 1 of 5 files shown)


colbymchenry:
Reviewed and merging — but with a fix on top.

The path-aliases half is solid: state-machine JSONC parser, correct wildcard semantics, specificity sort matches TS, replacements that escape projectRoot are skipped.

While reviewing the re-export half I wrote two tests for the chain logic. The wildcard one passed but only via the existing name-matcher fallback (the call site name matched a real declaration). The renamed-export test (`import { login }` from a barrel doing `export { signIn as login } from './auth'`) failed — the call site name `login` has no project-wide declaration and the resolver's fast pre-filter in `resolveOne()` bails before `resolveViaImport` runs, so chain-following never gets a chance.

Pushed a small commit on your branch:

- `matchesAnyImport()` escape so the pre-filter doesn't false-negative when the ref matches a local import
- the two re-export tests (3-hop wildcard chain, named-rename) to lock in the behavior

Both halves of this PR now do what the description claims. Thanks for the contribution — path aliases especially are a big quality win for modern JS/TS codebases.

PR #131, opened 2026-04-28
feat(search): field-qualified queries (kind:/lang:/path:/name:) + fuzzy typo fallback

andreinknv:
## Summary

Two UX improvements that turn free-text search into something a user can drive precisely.

### 1. Field-qualified queries

A new query parser splits the raw query into structured filters and a free-text remainder:

```
kind:function name:auth path:src/api authenticate
```

becomes:

```js
{ kinds: ['function'], nameFilters: ['auth'],
  pathFilters: ['src/api'], text: 'authenticate' }
```

Filters compose with the `SearchOptions` arg (intersection). Unknown prefixes pass through as plain text so `query "TODO:"` keeps working. Quoted values (`path:"my dir"`) handle whitespace. When the user supplies only filters with no text, the search uses a filter-only candidate scan instead of bailing out.

Recognised fields:
| Prefix | Value |
|--------|-------|
| `kind:` | any `NodeKind` value (`function`, `method`, `class`, ...) |
| `lang:` (alias `language:`) | any `Language` value |
| `path:` | case-insensitive substring of `file_path` |
| `name:` | case-insensitive substring of `node.name` |

### 2. Fuzzy typo fallback

When **both** FTS and LIKE return nothing AND the text is at least 3 chars, scan the distinct-name set with a bounded edit distance (≤2 for ≥5-char queries, ≤1 for 4-char). Bounded edit distance early-exits once the row min exceeds `maxDist`, so the per-query cost stays O(distinct-names × avg-name-length) with a very low constant.

## Test plan

Verified live against ollama/ollama@v0.22.0:

| Query | Result |
|-------|--------|
| `kind:function auth` | only function-kind hits |
| `lang:go path:server route` | Go files under `server/` |
| `getUssr` (typo) | finds `getUser`, `SetUser` |
| `confg` (typo) | finds `Config` |

- [x] `npx vitest run` — **380 passed**

### DIFF
--- a/__tests__/search-query-parser.test.ts
+++ b/__tests__/search-query-parser.test.ts
@@ -1,0 +1,142 @@
+/**
+ * Unit tests for the field-qualified query parser and bounded
+ * edit distance — the two algorithms behind `kind:`/`lang:`/`path:`/
+ * `name:` filtering and the fuzzy typo fallback.
+ */
+
+import { describe, it, expect } from 'vitest';
+import { parseQuery, boundedEditDistance } from '../src/search/query-parser';
+
+describe('parseQuery', () => {
+  it('returns plain text for a query with no field prefixes', () => {
+    const r = parseQuery('authenticate user');
+    expect(r.text).toBe('authenticate user');
+    expect(r.kinds).toEqual([]);
+    expect(r.languages).toEqual([]);
+    expect(r.pathFilters).toEqual([]);
+    expect(r.nameFilters).toEqual([]);
+  });
+
+  it('extracts kind: filter and removes it from text', () => {
+    const r = parseQuery('kind:function auth');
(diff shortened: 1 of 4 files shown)


colbymchenry:
Reviewed and merging. Pushed a small polish commit:

- Derived `KIND_VALUES` / `LANGUAGE_VALUES` from new `NODE_KINDS` / `LANGUAGES` `as const` arrays in `types.ts` so the parser stays in sync if a new kind or language gets added (e.g. `lang:unknown` now works because the type already had it).
- Fixed the `path:` JSDoc — it claimed prefix match but the implementation is substring.

The field-qualified syntax and the bounded-edit-distance fuzzy fallback are both clean, well-tested wins. Thanks for the contribution.

PR #125, opened 2026-04-28
feat(biomarkers): static-analysis Code Health findings + codegraph_biomarkers MCP tool

andreinknv:
> **🔗 Stacked on #118 (file-based migrations).** Until that lands, this PR's diff transitively includes the refactor commit. The intrinsic addition for *this* PR is the new files in `src/biomarkers/`, `src/index-hooks/biomarkers.ts`, `src/db/migrations/019-*.ts`, `src/mcp/tools/biomarkers.ts`, the QueryBuilder methods, and the tests — see "Files changed" below. After #118 lands, the diff auto-cleans.

## Summary

Five static-analysis biomarkers run on every indexed function/method and persist to a queryable `code_health_findings` table. The killer query an AI agent calls before refactoring:

```
mode=ranked, minSeverity=warning, minCentrality=0.001
```

— "high-impact code with structural problems" in one MCP call. Currently the agent has to read the function and eyeball its complexity. Now it gets the answer as structured data.

## What ships

**Biomarkers (McCabe-style metrics):**

| Biomarker | What it measures | Thresholds (info / warning / error) |
|---|---|---|
| **Large Method** | Inclusive line count of the function span | 50 / 100 / 200 |
| **Complex Method** | Cyclomatic complexity (branching nodes + 1) | 10 / 15 / 25 |
| **Nested Complexity** | Max nesting depth | 3 / 4 / 6 |
| **Complex Conditional** | Operands inside the most complex conditional in the body | 4 / 6 / 8 |
| **Brain Method** | All three of (large, complex, nested) at warning+ — fires as `error` | — |

**Code Health score (1-10):** aggregated from findings; -2/-1/-0.5 per error/warning/info, floored at 1.

**Plumbing:**
- `src/biomarkers/{engine,lang-map,index,types}.ts` — engine + per-language node-kind maps for TS/JS, Python, Go, Java (other languages skipped gracefully).
- `src/index-hooks/biomarkers.ts` — runs after indexAll (full scan) and afterSync (changed files only).
- Migration 019: `code_health_findings` keyed on `(node_id, biomarker)`, FK CASCADE.
- QueryBuilder methods: `replaceFindingsForFile`, `getFindingsForNode`, `getFindingsRanked`, `getFindingsStats`.
- MCP tool `codegraph_biomarkers` with modes `symbol`/`ranked`/`stats` and filters.
- `enableBiomarkers` config flag (default on).

## The agent-first angle

[...]

PR #132, opened 2026-04-28
feat(biomarkers): four new rules — unused_export, long_parameter_list, magic_number, hardcoded_url

andreinknv:
## Summary

Extends the biomarker engine (PR #125) with four new rules. Three operate on metrics gathered during the per-symbol AST walk; one is a cross-file graph rule.

### Per-symbol additions to `SymbolMetrics`
- `paramCount` — counted from the function/method's parameter list at AST walk time
- `magicNumberCount` — numeric literals not in the trivial allow-list (`0`, `1`, `-1`, `2`, plus float variants)
- `hardcodedUrlCount` — string literals matching `http(s)|ws(s)|ftp|s3|gs` URLs

### New rules
| Rule | Severity tiers |
|------|----------------|
| `long_parameter_list` | info=4 / warning=5 / error=7 (Clean Code thresholds) |
| `magic_number` | info=3 / warning=5 / error=8 |
| `hardcoded_url` | info on first / warning at 2 / error at 3+ |
| `unused_export` | warning — single SQL query: nodes with `is_exported=1` whose only incoming edges are `contains`/`exports`/`imports`/`tests` (or come from the same file) |

### DB additions
- `findUnusedExports()` — single graph query over the whole index
- `appendFindings(...)` — insert without delete (needed by cross-file rules so they don't clobber per-file rule output)
- `clearFindingsByKind(...)` — wipe one biomarker kind globally before re-emitting
- `replaceFindingsForFile` updated to **preserve** `unused_export` findings on per-file replace, so sync runs that touch a file don't silently lose its cross-file findings without recomputing them

### Edge-kind exclusions for `unused_export`
"Used" means a real consumer edge from outside the symbol's file. Excluded as not-real-use:
- `contains` — structural; every method is contained by its class
- `exports` — re-export barrels (`export { foo } from './a'`) forward but don't consume
- `imports` — the import statement itself isn't a use; real use is `calls`/`references`/`instantiates`/`extends`/etc.
- `tests` — convention-derived test→subject edges aren't a semantic API call

## Test plan

- [x] **Live verification** — synthetic stress fixture with one 7-param function using 5 magic numbers and 3 dead exports:
  ```
  fetchAll → long_parameter_list error (7), magic_number error (10)
  unused1, unused2, consumeFetch → unused_export warning
  ```
- [x] **Unit tests added** — 6 new `evaluateRules` cases covering each new rule's trigger thresholds (`long_parameter_list` warning at 5 + clean at 3, `magic_number` error at 8 + clean at 2, `hardcoded_url` info/warning/error tiers + clean at 0). Plus a `metrics()` helper so future metric additions don't bloat the test suite.
- [x] `npx vitest run` — **830 passed**

## Notes

- Pre-existing `clean code = 0 findings` test fixture changed to use non-exported helpers (semantically those WERE unused exports under the new rule — calibration artefact, not a regression)
- `unused_export` is skipped on partial scans (sync with `filePaths`) because it depends on global graph state; `replaceFindingsForFile` preserves existing `unused_export` rows on touched files so they remain valid until the next full pass
- `appendFindings` + `clearFindingsByKind` are wrapped in a single `queries.transaction(...)` so concurrent readers never see the intermediate empty state


colbymchenry:
Closing — leaf rules on #125's biomarker framework, which I'm also closing for architectural-scope reasons (see #125 for the full reasoning).

PR #133, opened 2026-04-28
feat(biomarkers): graph-aware god_class + feature_envy rules

andreinknv:
## Summary

Two more cross-file biomarker rules that lean on the graph rather than per-symbol AST metrics. Both are SQL-only and run during the cross-file pass after the per-symbol walk.

### `god_class`
Class-like nodes (`class`/`struct`/`interface`/`trait`/`protocol`) whose `contains`-edge subtree has at least N method/property/field children. Includes Go's `interface`-with-methods and Rust's `trait`-with-default-impls patterns.

| Tier | Members |
|------|---------|
| info | 15 |
| warning | 25 |
| error | 40 |

Thresholds from Lanza/Marinescu's *Object-Oriented Metrics in Practice* and CodeScene's published WMC ranges, picked at the conservative end so we under-flag rather than spam findings.

### `feature_envy`
Methods whose outbound `calls` edges land in OTHER files at least 5 times AND at least 2× more often than into their own file AND have at least 1 same-file call (so pure aggregators / facade methods that legitimately only call externals aren't flagged). "Same class" is approximated by "same file" — accurate for the dominant one-class-per-file convention across our supported languages.

| Tier | External calls |
|------|---------------|
| info | 5 |
| warning | 20 |

### Implementation notes
- New `runCrossFileRule(queries, kind, produce, onError)` helper wraps `clearFindingsByKind` + `appendFindings` in `queries.transaction` for atomicity (mirrors the `unused_export` pattern).
- New `CROSS_FILE_BIOMARKERS` constant in `src/biomarkers/types.ts` is the single source of truth — `replaceFindingsForFile` imports it so future cross-file rules only need to update one place.
- Edge dedup at insert (UNIQUE source/target/kind) means `externalCalls` counts distinct callee nodes, not invocation frequency. Documented in the helper's JSDoc.

## Test plan

- [x] **Live verification** on a synthetic fixture:
  - `GodClass` (16 methods) → `god_class` info, metric=16
  - `envious()` (5 distinct externals + 2 same-file) → `feature_envy` info, metric=5
  - `aggregate()` (5 externals + 0 same-file) → no finding (pure aggregator guard)
- [x] **3 new e2e tests** in `__tests__/biomarkers.test.ts` covering each scenario above
- [x] `npx vitest run` — **856 passed** (was 853, +3 new)

## Depends on

- PR #125 (`feat/biomarkers`) — biomarker framework
- PR #132 (`feat/biomarkers-more-rules`) — `unused_export` pattern + `appendFindings` / `clearFindingsByKind` infrastructure


colbymchenry:
Thanks, but closing — this PR's diff is +26k lines across 191 files, not the "two new rules" the description suggests. It's stacked on #125 and #132, both unmerged, and inherits their entire footprint (biomarkers engine, centrality/churn/cochange/coverage subsystems, 18 new migrations, HCL extractor, etc.).

The architectural decision lives in #125 — that's where to have the conversation about whether codegraph should ship a built-in biomarkers framework at all. Reviewing leaf rules in isolation isn't possible when the trunk hasn't been agreed to. Closing #133 (and likely #132) as part of that.

PR #134, opened 2026-04-28
feat(extraction): instantiates + decorates graph edges

andreinknv:
## Summary

Two new structural edges that fill significant gaps in the call graph for modern JS/TS / Java / C# / Python / Kotlin codebases.

### `instantiates` edges from `new Foo(...)`

Previously the extractor only recognised `call_expression`; `new_expression` (and equivalents in other grammars) was silently ignored, so constructor invocations produced **zero** graph edges.

Adds an `INSTANTIATION_KINDS` set covering `new_expression`, `object_creation_expression`, and `instance_creation_expression`. Wired into both the top-level `visitNode` dispatcher AND the per-function-body `visitForCallsAndStructure` walker — `new` calls inside function bodies were being missed by the body walker even after the top-level dispatch was hooked up.

Handles three subtleties:
- **Generic types**: `new Map<K, V>()` strips the angle-bracket suffix to produce `'Map'`, so resolution can match the class node.
- **Qualified constructors**: `new ns.Foo()` keeps the trailing identifier (`'Foo'`) — same shape as the existing `extractCall` does for member-call.
- **Nested calls**: children are still walked, so `new Foo(bar())` produces both an `instantiates` ref and a `calls` ref.

### `decorates` edges from `@Decorator` annotations

Tree-sitter places decorator nodes BEFORE the symbol they apply to, so a naive walk-time dispatch saw the wrong nodeStack head (file/class instead of class/method). The fix runs decorator extraction from inside `extractClass` / `extractFunction` / `extractMethod` / `extractProperty` / `extractField`, after the symbol's node id is known.

Looks for decorator nodes in two places:
- Direct named children of the declaration (method/property style).
- Preceding siblings in the parent (TypeScript class style: `@Foo class X {}` parses as `parent { decorator, class_decl }`).

Handles two subtleties:
- **Sibling boundary**: walks BACKWARD from the declaration and stops at the first non-decorator separator. Without that stop, `@A class Foo {} @B class Bar {}` would attribute `@A` to `Bar`. Caught by reviewer; covered by a regression test.
- **Tree-sitter wrapper identity**: `parent`/`namedChild` returns fresh JS wrapper objects, so `sibling === declNode` is unreliable — uses `startIndex` comparison instead.

Also recognises Java's `marker_annotation` (no-args annotations like `@Override`/`@Deprecated`) alongside `decorator`/`annotation`.

## Test plan

Verified live on a synthetic NestJS-shape fixture:

```
src              kind         tgt
UserController   decorates    Controller   (class decorator)
list             decorates    Get          (method decorator)
bootstrap        instantiates UserService  (new ...())
bootstrap        instantiates UserController
```

- [x] **6 new extraction tests** covering `new Foo()` / generic-stripping / qualified-new / `@Foo class X` / two-adjacent-decorated-classes regression / `@Foo method()`
- [x] `npx vitest run` — **386 passed** (was 380, +6 new)

### DIFF
--- a/__tests__/extraction.test.ts
+++ b/__tests__/extraction.test.ts
@@ -3079,3 +3079,101 @@ describe('Directory Exclusion', () => {
     expect(files.every((f) => !f.includes('vendor'))).toBe(true);
   });
 });
+
+describe('Instantiates + Decorates edge extraction', () => {
+  it('emits an instantiates ref for `new Foo()`', () => {
+    const code = `
+class Foo {}
+function bootstrap() { return new Foo(); }
+`;
+    const result = extractFromSource('app.ts', code);
+    const ref = result.unresolvedReferences.find(
+      (r) => r.referenceKind === 'instantiates' && r.referenceName === 'Foo'
+    );
+    expect(ref).toBeDefined();
+  });
+
+  it('strips type-argument suffix from generic constructors', () => {
+    const code = `
+class Container<T> { constructor(_: T) {} }
+function go() { return new Container<string>('x'); }
+`;
+    const result = extractFromSource('app.ts', code);
+    const ref = result.unresolvedReferences.find(
+      (r) => r.referenceKind === 'instantiates'
+    );
+    expect(ref).toBeDefined();
+    // Container<string> must be normalised to "Container" — otherwise
+    // resolution can never match the class node.
+    expect(ref!.referenceName).toBe('Container');
+  });
+
+  it('keeps trailing identifier from qualified `new ns.Foo()`', () => {
+    const code = `
+const ns = { Foo: class {} };
+function go() { return new ns.Foo(); }
+`;
+    const result = extractFromSource('app.ts', code);
+    const ref = result.unresolvedReferences.find(
+      (r) => r.referenceKind === 'instantiates'
+    );
+    // We can't always resolve which Foo, but the name should be the
+    // simple identifier so name-matching has a chance.
+    expect(ref?.referenceName).toBe('Foo');
+  });
+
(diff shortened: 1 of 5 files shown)


colbymchenry:
Reviewed and approved. Pushed two follow-ups before merging (commit `63a4bbe`):

**1. Kind-aware scoring in `name-matcher.ts`**

The existing scorer only had a kind bonus for `calls` (preferring `function`/`method`). With the two new ref kinds, an `instantiates` ref like `new Logger()` could tie or pick a same-named function over the actual class when both exist across modules. Added:
- `instantiates` → +25 for `class`/`struct`/`interface`
- `decorates` → +25 for `function`/`method`, +15 for `class`/`interface` (Python class decorators, Java annotation interfaces)

**2. Python instantiation promotion in `createEdges`**

The PR description listed Python, but Python (and Ruby) have no `new` keyword — `Foo()` is the standard instantiation syntax and looks identical to a function call at extraction time. Resolution can tell the difference once the target is known, so I mirrored the existing `extends`→`implements` promotion: when a `calls` ref resolves to a `class`/`struct`, kind is promoted to `instantiates`. This covers Python without any extractor changes.

**Known behavior change (left as-is, intentionally):**

Dart's `new Widget()` previously emitted a `calls` edge via `extractBareCall`. The new `INSTANTIATION_KINDS` branch wins ahead of that path in `visitForCallsAndStructure`, so it now emits `instantiates` instead. Semantically more accurate, but a breaking change for any consumer querying "what calls Widget" on Dart code. Worth a release-note line.

Verified: 386 → 389 passing (+3 tests covering the kind biases and the Python promotion). `tsc --noEmit` clean.

Merging.

PR #135, opened 2026-04-28
feat(search): signature search + callers-of / callees-of graph qualifiers

andreinknv:
## Summary

Three new search query qualifiers extending PR #131's field-qualified parser. All compose with the existing `kind:` / `lang:` / `path:` / `name:` filters.

### `sig:` / `signature:` — substring of node.signature
Find functions by what they accept or return:
```
query "sig:Promise<User> kind:function"
query "sig:context.Context"
query "sig:io.Reader sig:io.Writer"   # OR semantics
```

### `callers-of:NAME` — nodes that call NAME
```
query "callers-of:authenticate"
query "callers-of:authenticate name:Handler"
query "callers-of:Forward kind:function"
```

### `callees-of:NAME` — nodes called BY NAME
```
query "callees-of:bootstrap"
```

## Implementation notes

- New `nodesCallingAny(names)` and `nodesCalledByAny(names)` SQL helpers; `nodesByIds(ids)` for bulk node fetching when seeded from the graph.
- Graph-seeded candidate set when callers-of/callees-of is given without text — otherwise the filter-only path scans 50 arbitrary name-ordered nodes that rarely intersect with the graph result.
- `searchAllByFilters` accepts `signatureLike: string[]` and OR's all entries in SQL — matches the OR semantics of `path:` / `name:` post-filters.
- All IN-clause helpers chunk inputs to 900 per query (under SQLite's 999 variable cap) so pathological inputs can't trigger `SQLITE_ERROR: too many SQL variables`.
- Graph-seeded path applies `kinds`/`languages` post-filter to the seed set so `callers-of:foo kind:method` correctly narrows.
- Reviewer-driven: also memoises `callersSet()` / `calleesSet()` closures so the seed and the filter pass don't re-query the same names.

## Test plan

Verified live on ollama/ollama@v0.22.0:

| Query | Result |
|-------|--------|
| `sig:context.Context kind:function` | `BuildLauncherState`, `Chat`, ... |
| `callers-of:Forward kind:function` | `audioFeedForward`, `forwardConvBlock`, ... |
| `auth sig:error` | `getAuthorizationToken` (text + sig: composes) |

- [x] **5 new integration tests** in `__tests__/search-qualifiers.test.ts` covering `sig:`, multi-`sig:` OR, `callers-of:`, `callees-of:`, and `callers-of: + kind:` composition
- [x] **2 new parser tests** for `sig:` and `callers-of:` / `callees-of:` parsing
- [x] `npx vitest run` — **410 passed** (was 405)

## Depends on

- PR #131 (`feat/search-fields-and-fuzzy`) — parser foundation + chunking infrastructure

### DIFF
--- a/__tests__/search-qualifiers.test.ts
+++ b/__tests__/search-qualifiers.test.ts
@@ -1,0 +1,120 @@
+/**
+ * Integration tests for search qualifiers that go through the SQL
+ * layer: sig:, callers-of:, callees-of:.
+ *
+ * Mirrors the synthetic-fixture style of resolution.test.ts — small
+ * temp project, real CodeGraph index, query the result. These cover
+ * the SQL paths (signatureLike pre-filter, nodesCallingAny / nodesCalledByAny)
+ * that the parser-only tests can't exercise.
+ */
+
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as os from 'os';
+import { CodeGraph } from '../src';
+
+describe('Search qualifiers — DB-level integration', () => {
+  let tempDir: string;
+  let cg: CodeGraph | null = null;
+
+  beforeEach(() => {
+    tempDir = fs.mkdtempSync(path.join(os.tmpdir(), 'codegraph-qualifiers-'));
+  });
+
+  afterEach(() => {
+    if (cg) {
+      cg.destroy();
+      cg = null;
+    } else if (fs.existsSync(tempDir)) {
+      fs.rmSync(tempDir, { recursive: true });
+    }
+  });
+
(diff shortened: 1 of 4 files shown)


colbymchenry:
Thanks for the contribution! Closing this one — the direction overlaps too much with what's already shipped rather than for any code-quality reason.

`callers-of:` / `callees-of:` duplicate the `codegraph_callers` and `codegraph_callees` MCP tools, which are the canonical entry points for graph traversal. Exposing the same traversals through a string DSL gives LLM consumers two ways to do the same thing, and the dedicated tools are easier to call correctly.

`sig:` is genuinely new and useful — if you'd like to land that piece, a smaller standalone PR (just signature filtering, rebased on main once #131 merges) would be very welcome.

Also noting that this PR is stacked on #131, which isn't merged yet — so it can't land in its current form regardless.

PR #136, opened 2026-04-29
fix: add @clack/prompts transitive deps to fix npx installation

mfrancime:
## Summary

- Adds `fast-wrap-ansi`, `fast-string-width`, and `sisteransi` as direct dependencies

## Problem

When installed via `npx`, npm's flat `node_modules` cache fails to hoist ESM-only transitive dependencies from `@clack/prompts` → `@clack/core`. This causes:

```
Cannot find package '/root/.npm/_npx/.../node_modules/fast-wrap-ansi/index.js'
imported from /root/.npm/_npx/.../node_modules/@clack/core/dist/index.mjs
```

**Reproduces on:** Node 24.11.0 + npm 11.6.1 with `npx @colbymchenry/codegraph@0.7.3`

## Root cause

`@clack/core@1.2.0` imports `fast-wrap-ansi` (ESM-only, no `exports` map, just `"main": "lib/main.js"`). In a normal `npm install`, dependency hoisting puts `fast-wrap-ansi` in the top-level `node_modules/`. But in an `npx` cache directory, the flat structure sometimes fails to resolve these transitive ESM deps, causing a `Cannot find package` error at runtime.

## Fix

Promote the three transitive dependencies (`fast-wrap-ansi`, `fast-string-width`, `sisteransi`) to direct dependencies in `package.json`. This guarantees they're always installed at the top level regardless of installation method. These are already required at runtime — this just makes the dependency explicit.

## Test plan

- [ ] `npx @colbymchenry/codegraph --help` works on Node 24 after clearing `~/.npm/_npx`
- [ ] `npm install -g @colbymchenry/codegraph` still works
- [ ] Interactive installer launches without import errors

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,14 +7,15 @@
     "": {
       "name": "@colbymchenry/codegraph",
       "version": "0.7.2",
-      "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
-        "@clack/prompts": "^1.2.0",
-        "@xenova/transformers": "^2.17.0",
+        "@clack/prompts": "^1.3.0",
         "commander": "^14.0.2",
+        "fast-string-width": "^3.0.2",
+        "fast-wrap-ansi": "^0.2.0",
         "node-sqlite3-wasm": "^0.8.30",
         "picomatch": "^4.0.3",
+        "sisteransi": "^1.0.5",
         "tree-sitter-wasms": "^0.1.11",
         "web-tree-sitter": "^0.25.3"
       },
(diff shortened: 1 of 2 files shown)


colbymchenry:
Thanks for the diagnosis — verified the npm hoisting issue is real and the transitive-deps were exactly as you identified.

Pushed a follow-up commit to your branch bumping the pins to match `@clack/prompts@1.3.0` (which shipped after this PR was opened and changed the transitive versions to `fast-wrap-ansi@^0.2.0` and `fast-string-width@^3.0.2`). Merging now.

PR #151, opened 2026-05-11
test+feat: add cargo workspace crate resolution for rust resolver

miketheman:
Hi there! Thanks for building codegraph - it's a fun project, and useful in a variety of languages I use.
I noticed that it didn't handle [Cargo Workspace](https://doc.rust-lang.org/cargo/reference/workspaces.html) traversals yet.

If you'd prefer to have your own Claude drive the solution, that's cool too - but I figured this would at the very least help start the conversation.

---

`codegraph` indexed Rust files in workspace repos but did not resolve crate-style references (e.g. `mytool_core`) across workspace members under `crates/*`. This PR adds workspace manifest awareness so Rust reference resolution can map crate names to member crate roots.

- **Workspace manifest parsing**
  - Added `src/resolution/frameworks/cargo-workspace.ts` to parse:
    - `[workspace].members` from root `Cargo.toml`
    - `[package].name` from each member `Cargo.toml`
  - Builds crate alias map with both Cargo and Rust forms (`mytool-core` and `mytool_core`).

- **Rust resolver integration**
  - Extended `src/resolution/frameworks/rust.ts` module resolution to check workspace crate roots when resolving module-like references:
    - `<member>/src/lib.rs`
    - `<member>/src/main.rs`
  - Preserves existing local `src/<name>.rs` and `src/<name>/mod.rs` behavior.

- **Performance and safety refinements**
  - Added per-context caching of parsed workspace crate map (`WeakMap`) to avoid reparsing TOML for each reference.
  - Hardened parser handling for quoted values, bracket parsing, and regex key escaping.

- **TDD coverage for new behavior**
  - Added framework-level tests validating crate-name resolution from workspace members for both `lib.rs` and `main.rs` paths.

```rust
// crates/mytool-fetcher/src/main.rs
use mytool_core::selector::pick;
//            ^ resolves to crates/mytool-core/src/lib.rs via workspace map
```

### DIFF
--- a/__tests__/frameworks.test.ts
+++ b/__tests__/frameworks.test.ts
@@ -260,6 +260,138 @@ describe('rustResolver.extract', () => {
   });
 });
 
+describe('rustResolver.resolve cargo workspace crates', () => {
+  it('resolves crate name from workspace member lib.rs', () => {
+    const workspaceCargo = `
+[workspace]
+members = ["crates/mytool-core", "crates/mytool-fetcher"]
+`;
+    const coreCargo = `
+[package]
+name = "mytool-core"
+version = "0.1.0"
+`;
+    const libNode: Node = {
+      id: 'module:crates/mytool-core/src/lib.rs:mytool_core:1',
+      kind: 'module',
+      name: 'mytool_core',
+      qualifiedName: 'crates/mytool-core/src/lib.rs::mytool_core',
+      filePath: 'crates/mytool-core/src/lib.rs',
+      language: 'rust',
+      startLine: 1,
+      endLine: 1,
+      startColumn: 0,
+      endColumn: 0,
+      updatedAt: Date.now(),
+    };
+
+    const context = {
+      getNodesInFile: (fp: string) => (fp === 'crates/mytool-core/src/lib.rs' ? [libNode] : []),
+      getNodesByName: () => [],
+      getNodesByQualifiedName: () => [],
+      getNodesByKind: () => [],
+      fileExists: (p: string) => (
+        p === 'Cargo.toml' ||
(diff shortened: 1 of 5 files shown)


colbymchenry:
Thanks @miketheman — clean PR. Pulled it down and pushed one follow-up commit (`dc895ef`) on top before merging:

**Glob expansion for `members`.** The original parser only handled literal member paths, so workspaces using `members = ["crates/*"]` (uv, ruff, bevy, polkadot-sdk, …) silently no-op'd. Added picomatch-based expansion via a new optional `ResolutionContext.listDirectories` method (kept the abstraction — no raw fs in resolvers), with a depth cap and the usual skip-dirs (`target`, `node_modules`, `.git`, etc.).

**Bumped Pattern 4 confidence to 0.95 on workspace hits.** While validating end-to-end against `astral-sh/uv` I found that even with the workspace map populated, zero cross-crate `imports` edges were materializing. The framework result (0.6) was losing to name-matcher's self-file match (0.7) — every file with `use foo::...` has its own import node named `foo`, so it self-resolved. Cargo's manifest is an unambiguous crate-name → crate-root mapping, so trusting it above name-matcher self-matches is the right call. The original isolated tests still pass; added a confidence assertion to lock in the new behavior.

**Validation:**
- `astral-sh/uv` (`members = ["crates/*"]`, 67 crates, 567 .rs files): **1,969 cross-crate `imports` edges reaching 60 distinct crate roots — up from 0.**
- `helix-editor/helix` (literal members, exercises the pre-existing path): 343 cross-crate edges to 10 crate roots.
- Full suite: 519/519 passing.

Merging.

Issue #168, opened 2026-05-18
Terminal output garbled/mojibake on Windows (PowerShell) during codegraph index


starkleek:
### Environment
- OS: Windows 11
- Node: v20.19.5
- codegraph: 0.7.9
- Terminal: ( Cursor integrated terminal / Windows Terminal / cmd)

### Steps to Reproduce
1. `codegraph init -i`
2. `codegraph index`

### Expected Behavior
Terminal output shows clean progress with colored text.

### Actual Behavior
Output contains garbled characters (mojibake), for example:

<img width="546" height="451" alt="Image" src="https://github.com/user-attachments/assets/70d56a15-3718-4317-88f9-44952006f17d" />

zhaorx:
me too

colbymchenry:
Fixed in #178, will ship in **0.7.10**.

**Root cause:** the shimmer progress renderer runs in a worker thread and writes to stdout via `fs.writeSync(1, …)` to keep the animation smooth while the main thread is blocked in SQLite. That path bypasses Node's TTY-aware UTF-8→codepage conversion on Windows, so glyphs like `│ ◆ —` hit your console as raw UTF-8 bytes and got reinterpreted by the active OEM codepage (CP437 / CP936 / …) — producing `鋍?[0m 鉒?[0m Scanning files 鈥?N found`. Clack's output in the same screenshot rendered fine because clack writes through the main thread's TTY-aware path.

**Fix:** CodeGraph now picks an ASCII glyph set on Windows by default (`| * -` instead of `│ ◆ —`). The shimmer animation, status icons, and tree drawing all participate.

**Escape hatches:**
- `CODEGRAPH_UNICODE=1` to opt back into Unicode (e.g. pwsh 7 where you've set `[Console]::OutputEncoding` to UTF-8)
- `CODEGRAPH_ASCII=1` on any platform to force ASCII (useful for log collectors and non-TTY pipelines)

**Workaround until 0.7.10 ships** — for current versions on Windows PowerShell, set `[Console]::OutputEncoding = [System.Text.Encoding]::UTF8` and `chcp 65001` before running `codegraph index`. That fixes the console codepage so UTF-8 bytes render correctly.

Thanks for the report, and thanks to @Bortlesboat for the initial PR (#174).

PR #174, opened 2026-05-19
fix: avoid mojibake in Windows terminal status output

Bortlesboat:
## Summary
- replace unicode status glyphs with ASCII markers ([OK], [ERR], [INFO], [WARN]) in shared CLI helpers
- replace em dash in verbose progress output with ASCII - for safer rendering across terminals
- keep behavior unchanged except output formatting

## Why
Issue #168 reports garbled output on Windows PowerShell. This keeps status text readable in non-UTF terminal paths and log collectors.

## Validation
- 
pm ci`n- 
pm test (existing upstream failures in installer-target tests + Node OOM; unrelated to this formatting-only change)

### DIFF
--- a/src/bin/codegraph.ts
+++ b/src/bin/codegraph.ts
@@ -32,7 +32,7 @@ async function loadCodeGraph(): Promise<typeof import('../index')> {
     return await import('../index');
   } catch (err) {
     const msg = err instanceof Error ? err.message : String(err);
-    console.error('\x1b[31m✗\x1b[0m Failed to load CodeGraph modules.');
+    console.error('\x1b[31m[ERR]\x1b[0m Failed to load CodeGraph modules.');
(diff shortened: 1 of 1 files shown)


colbymchenry:
Thanks for the patch! Superseded by #178, which keeps the spirit of your fix (ASCII markers for status icons) but extends coverage to `src/ui/shimmer-worker.ts` — the worker thread is the actual source of the mojibake in #168's screenshot because it uses `fs.writeSync(1, …)` and bypasses Node's TTY-aware UTF-8→codepage conversion on Windows. It also adds a `supportsUnicode()` check so macOS/Linux terminals keep the Unicode rendering.

You're credited in the CHANGELOG entry that ships in 0.7.10. Closing this one in favour of the merged fix.

PR #182, opened 2026-05-19
feat!: migrate from npm/better-sqlite3 to bun + bun:sqlite

AngerM:
Switches the runtime from Node + better-sqlite3 (with a node-sqlite3-wasm fallback) to Bun + the built-in bun:sqlite. Drops the optional native build entirely — bun:sqlite ships with the runtime, so there's no longer a slow WASM fallback to chase.

Notable user-visible changes:
- engines.bun >= 1.1; engines.node removed
- bin/codegraph shebang is now #!/usr/bin/env bun
- WASM-fallback banner, getBackend(), and "Backend:" line in status are gone
- Node 25 OOM block is gone (irrelevant under Bun)
- Test runner switched from vitest run to `bun test`
- Build uses tsconfig.build.json so tsc keeps the web-tree-sitter type override paths mapping while tsconfig.json (read by Bun and the IDE) stays free of it

Tooling:
- npm/npx removed from package.json scripts, publish.js, scripts/*
- package-lock.json -> bun.lock
- New @types/bun devDep; @types/better-sqlite3 dropped

Adapter notes:
- src/db/sqlite-adapter.ts wraps bun:sqlite with the same surface the rest of CodeGraph was written against (prepare/run/get/all, named- param translation, pragma() with {simple: true}, transaction()).
- New src/installer/targets/home.ts because Bun caches os.homedir() at startup and does not re-read process.env.HOME; the installer-targets tests mutate HOME and need a fresh read.

Tests: 591 pass under `bun test`. Watcher tests gained a 100ms settle after watcher.start() — Bun's recursive fs.watch needs a brief grace period before per-file events arrive.


AngerM:
Meant to keep this on my private fork for now.

PR #201, opened 2026-05-20
fix(extraction): index independent nested git repos

### LINKED ISSUE #193: `codegraph init -i` does not detect source files when run from a top-level CMake workspace containing multiple nested repos
### Description
Our project uses **CMake** and is organized as a top-level workspace that contains several independent Git repositories. These are **not** Git submodules.

Example structure:

```text
root/
├── coding/
├── CMakeLists.txt
├── sub_repo1/
│   ├── CMakeLists.txt
│   └── src/
└── sub_repo2/
    ├── CMakeLists.txt
    └── src/
```

The top-level directories (`root/` and `coding/`) do not contain any source files directly. All source code is located inside the nested repositories such as `sub_repo1/` and `sub_repo2/`.

### Problem
When I run:

```bash
codegraph init -i
```

from either `root/` or `coding/`, it reports:

```text
No files found to index
```

However, if I run the same command inside `sub_repo1/`, it works correctly.

### Expected behavior
I expect `codegraph init -i` to discover and index source files in nested repositories when run from the top-level project directory.

### Actual behavior
`codegraph init -i` only works when executed inside each individual sub-repository. It does not work from the main project root, which is the directory we normally open and work in.

### Why this matters
In our workflow, developers usually open the entire workspace from the top-level directory. Currently, indexing has to be done separately for each nested repository, which is inconvenient and makes project-wide indexing difficult.

sjh9714:
## Summary

- Fixes `codegraph init -i`/scan discovery for top-level workspaces that contain independent nested Git repos.
- Expands untracked nested-repo directory entries from `git ls-files -o --exclude-standard` by enumerating that repo’s own visible files.
- Adds regression coverage for issue #193 while preserving existing Git submodule scan behavior.

## Why

Before this change, a parent Git repo saw an unregistered nested repo as a single `sub_repo1/` untracked directory. That directory did not match source include patterns, so top-level workspace scans could report no files even though the nested repo contained source files.

## Testing

- `npx vitest run __tests__/extraction.test.ts -t "independent nested git"` — failed before the fix with `expected [] to include 'sub_repo1/src/main.cpp'`
- `npx vitest run __tests__/extraction.test.ts -t "independent nested git|git submodules"`
- `npx vitest run __tests__/extraction.test.ts`
- `npm run build`
- `npm test`
- `git diff --check`

Fixes #193

Note: I used Codex to help inspect the scan path and draft the regression test, then reviewed the final diff and ran the listed verification commands locally.

### DIFF
--- a/__tests__/extraction.test.ts
+++ b/__tests__/extraction.test.ts
@@ -3130,6 +3130,37 @@ describe('Git Submodules', () => {
     expect(files).toContain('app.ts');
     expect(files).toContain('libs/lib/lib.ts');
   });
+
+  it('should index files inside independent nested git repositories (issue #193)', async () => {
+    const { execFileSync } = await import('child_process');
+    const git = (cwd: string, ...args: string[]) =>
+      execFileSync('git', args, { cwd, stdio: 'pipe' });
+
+    const mainDir = path.join(tempDir, 'main');
+    fs.mkdirSync(mainDir, { recursive: true });
+    git(mainDir, 'init', '-q');
+    git(mainDir, 'config', 'user.email', 'test@test.com');
+    git(mainDir, 'config', 'user.name', 'Test');
+    fs.writeFileSync(path.join(mainDir, 'CMakeLists.txt'), 'add_subdirectory(sub_repo1)\n');
+    git(mainDir, 'add', 'CMakeLists.txt');
+    git(mainDir, 'commit', '-q', '-m', 'root init');
+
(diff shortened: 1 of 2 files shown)


sjh9714:
Thanks for the review window here. Upstream has now merged #217, which fixes #193 with the same nested non-submodule git repository behavior and regression coverage, so this PR is obsolete rather than worth rebasing through the conflict. Closing to avoid duplicating the merged fix.

PR #202, opened 2026-05-20
Enhance shimmer progress with performance metrics

krataratha:
Added optional fields for estimated time remaining and items per second in IndexProgress interface. Enhanced createShimmerProgress function to calculate and pass performance metrics.

### DIFF
--- a/src/ui/shimmer-progress.ts
+++ b/src/ui/shimmer-progress.ts
@@ -12,6 +12,9 @@ export interface IndexProgress {
   phase: string;
   current: number;
   total: number;
+  // Added optional fields for the new metrics
+  estimatedTimeRemainingMs?: number;
+  itemsPerSecond?: number;
 }
 
 export interface ShimmerProgress {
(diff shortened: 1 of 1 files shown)


colbymchenry:
Thanks for the idea, @krataratha — surfacing ETA and items/sec during indexing is a nice touch. Unfortunately this PR can't be merged as-is:

- **It doesn't compile.** The diff closes `createShimmerProgress` partway through `onProgress` (an early `};}` plus a second `stop()`), leaving the original body — `lastPhase`, the percent/count logic, and the real `stop()` — dangling at the top level. `tsc` fails with 6 syntax errors, and there end up being two `stop()` methods.
- **The metrics never render.** It posts `{ type: 'progress-update', ... }` to the worker, but `shimmer-worker.ts` only handles `update` / `finish-phase` / `stop` — so the computed numbers are dropped, and `phaseName` there is out of scope. Actually showing ETA/items-per-sec needs changes in the worker's `render()` (where the progress line is drawn), which this PR doesn't touch.

The salvageable part is just the two optional fields on `IndexProgress`. Since making this real means implementing the rendering (and validating it on Windows, where the worker writes raw bytes to fd 1 and relies on `getGlyphs()` ASCII fallbacks), it's better done as a fresh, focused change than patched onto this one.

Closing as not-mergeable — but the feature idea is welcome if you'd like to take another run at it. 🙏

PR #219, opened 2026-05-20
feat: Nix support

vic:
Tested locally by indexing a Nix project, and also via the codegraph cli query.


src/types.ts
```diff
@@ -75,26 +75,18 @@ export const LANGUAGES = [
...
   'scala',
-  'lua',
-  'luau',
-  'objc',
-  'r',
-  'yaml',
-  'twig',
-  'xml',
-  'properties',
+  'nix',
```
ruinshe, line 89: `nix` is currently added after `dart`, the second one looks redundant and should be removed.

ruinshe:
I tested the patch locally against a real Nix-heavy repo, and it worked correctly in my setup.

What I verified:

- `codegraph status .` reports `.nix` files under "Files by Language"
- file-level lookup works for Nix files (`codegraph node flake.nix --file flake.nix`)
- symbol lookup works for Nix bindings (`codegraph query opencode-pinned.url`)
- imports are extracted as graph nodes as well

Concrete numbers from my local test repo:

- 442 indexed files total
- 396 indexed `.nix` files
- 9,056 total nodes
- 8,733 Nix-language nodes
  - 396 `file`
  - 8,232 `variable`
  - 105 `import`

Sample extracted imports from the local graph:

- `./nix` from `flake.nix`
- `./helper.nix` from `flakes/agent-skills/anthropic.nix`
- `./default.nix` from `flakes/agent-skills/flake.nix`

vic:
Closed by #1190

PR #226, opened 2026-05-20
docs(readme): add codegraph_explore to MCP Tools table and bump Node badge to 20+

AimoreRRD:
Two small README mismatches between the docs and the current project state:

### 1. `codegraph_explore` is missing from the MCP Tools table

`src/mcp/tools.ts` exports 9 tools, but the table near the bottom of the README lists only 8 — `codegraph_explore` is absent. The tool is referenced prominently elsewhere in the same README (lines 96, 239, 243, 246, 247) including being singled out as a primary exploration tool, so a reader who skims the table can miss that it exists.

Added a row between `codegraph_node` and `codegraph_files` to match the source-file order. Description matches the opening sentence of the tool's own `description` field in `src/mcp/tools.ts`.

### 2. Node.js badge is stale (says 18+, should be 20+)

`package.json` `engines` has required `>=20.0.0` since commit 9b06b0e (`fix(db): require better-sqlite3 ^12.4.1 so Node 24 gets the native backend`), which explicitly raised the floor because Node 18 was EOL and dropped from better-sqlite3 12.x prebuilds. The badge still claims "Node.js 18+", which misleads users.

Bumped the badge to `20+`. (Left the upper bound alone — the engines pin `<25.0.0` is being addressed by PR #222 and a one-shot badge change shouldn't pre-empt that.)

### Scope

- Only `README.md` touched.
- 1 line added, 1 line changed.
- No code, no tests, no behaviour change.


colbymchenry:
Merged — thanks @AimoreRRD! 🎉 Took **Part 1** (the `codegraph_explore` row): good catch, `tools.ts` exports 9 tools but the table listed 8. I added it to the full 'MCP server exposes these tools' table only — not the curated 'lightweight tools the main session may use directly' table above it, since explore is intentionally the heavy/delegate-it tool there.

Dropped **Part 2** (Node 18+→20+ badge): that badge no longer exists — it was replaced by 'Node.js bundled · none required' when CodeGraph became self-contained (bundles its own runtime), so there's no user-facing version floor to advertise anymore. Credited to you via squash. 🙏

PR #233, opened 2026-05-21
fix(config): validate languages from canonical list

kietdev27:
## Summary

- Validate `config.languages` against the canonical `LANGUAGES` list from `src/types.ts`
- Remove the stale hard-coded language allowlist in `src/config.ts`
- Add regression coverage for supported languages that were previously rejected

## Why

`validateConfig()` only allowed a small older subset of languages. Supported languages such as `php`, `tsx`, `vue`, `liquid`, `pascal`, and `scala` could be rejected even though they are valid CodeGraph languages.

## Test plan

- [x] `npx vitest run __tests__/foundation.test.ts`
- [x] `npm run build`
- [x] `npx vitest run __tests__/foundation.test.ts __tests__/search-query-parser.test.ts __tests__/extraction.test.ts -t "supported languages|Configuration|parseQuery"`

Note: `npm test` full suite was attempted on Windows/Node v24.14.1 but failed in unrelated installer/MCP temp cleanup tests and ended with a V8 OOM worker crash.

### DIFF
--- a/__tests__/foundation.test.ts
+++ b/__tests__/foundation.test.ts
@@ -10,7 +10,7 @@ import * as path from 'path';
 import * as os from 'os';
 import { CodeGraph } from '../src';
 import { DEFAULT_CONFIG, Node, Edge } from '../src/types';
-import { loadConfig, saveConfig } from '../src/config';
+import { loadConfig, saveConfig, validateConfig } from '../src/config';
 import { isInitialized, getCodeGraphDir, validateDirectory } from '../src/directory';
 import { DatabaseConnection, getDatabasePath } from '../src/db';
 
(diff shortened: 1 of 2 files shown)


colbymchenry:
Thanks for this, @kietdev27 — and sorry for the timing. This was a legitimate fix when you opened it: `validateConfig()` did carry a stale, narrower language allowlist than the canonical `LANGUAGES` list, so configs naming `php`/`tsx`/`vue`/etc. could be wrongly rejected.

It's since been overtaken by #285 ("zero-config indexing driven by .gitignore"), merged May 21, which removed `src/config.ts` and `validateConfig()` entirely. CodeGraph no longer has a config file or a `languages` allowlist — indexing is now driven by `.gitignore`, so there's nothing left to validate against, and the file this PR patches no longer exists on `main`.

Closing as obsolete — this reflects the refactor, not the quality of the change. The regression test was a nice touch; hope you'll send more PRs. 🙏

PR #240, opened 2026-05-21
Fix: Increase SQLite busy_timeout to 5 minutes to mitigate locking errors

### LINKED ISSUE #238: SQLite "database is locked" error on concurrent MCP tool calls
## Problem

When multiple CodeGraph MCP tools are invoked in parallel (e.g., `codegraph_callees` and `codegraph_callers` in the same request batch), some calls intermittently fail with:

```
Error: Tool execution failed: database is locked
```

## Environment

- CodeGraph index: 298 files, 3073 nodes, 6.68 MB SQLite database
- Called from an AI agent that batches independent tool calls concurrently

## Reproduction

Invoke two or more read-only CodeGraph MCP tools in the same message (parallel tool calls). For example:

```
codegraph_callees("DeployAndLaunchAsync")
codegraph_callers("PlayAsync")
```

At least one will fail with `database is locked`.

## Suggestion

Since these are all read-only queries, enabling SQLite WAL mode (`PRAGMA journal_mode=WAL`) or adding connection retry/busy timeout (`PRAGMA busy_timeout=5000`) would allow concurrent reads without issues.

9chait9:
Fixes #238. Increased SQLite busy_timeout from 2 minutes (120000ms) to 5 minutes (300000ms) in src/db/index.ts to provide more buffer for concurrent read operations, especially when the WASM fallback is active and WAL mode is not available.

### Test Plan
- Unit tests have been added in `src/db/__tests__/index.test.ts`.
- The tests cover the `initialize` and `open` methods of `DatabaseConnection` to verify `busy_timeout` is set to 300000ms.
- Basic functionality of `DatabaseConnection` class such as `getDb`, `getPath`, `isOpen`, `close`, `transaction`, `getSize`, and `optimize` are also covered by unit tests.

### DIFF
--- a/src/db/__tests__/index.test.ts
+++ b/src/db/__tests__/index.test.ts
@@ -1,0 +1,185 @@
+
+import { DatabaseConnection, getDatabasePath, DATABASE_FILENAME } from '../index';
+import { createDatabase, SqliteDatabase, SqliteBackend } from '../sqlite-adapter';
+import * as fs from 'fs';
+import * as path from 'path';
+
+// Mock fs and path modules
+jest.mock('fs', () => ({
+  existsSync: jest.fn(),
+  mkdirSync: jest.fn(),
+  readFileSync: jest.fn(),
+  statSync: jest.fn(),
+}));
+
+jest.mock('path', () => ({
+  dirname: jest.fn(),
+  join: jest.fn(),
+}));
+
+// Mock sqlite-adapter
+jest.mock('../sqlite-adapter', () => ({
+  createDatabase: jest.fn(),
+  WASM_FALLBACK_FIX_RECIPE: 'test-recipe',
+}));
+
+// Mock migrations
+jest.mock('../migrations', () => ({
+  runMigrations: jest.fn(),
+  getCurrentVersion: jest.fn(),
+  CURRENT_SCHEMA_VERSION: 1,
+}));
+
+describe('DatabaseConnection', () => {
+  let mockDb: SqliteDatabase;
+  let mockBackend: SqliteBackend;
+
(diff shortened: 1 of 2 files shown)


colbymchenry:
Thanks for the PR and the detailed test plan, @9chait9 — but this is already handled on main, and raising the timeout would undo a deliberate fix.

**#238 is already fixed** (and closed as completed) by #282 / commit `ac52fd7`, the node:sqlite consolidation. The root cause in #238 was concurrent **read-only** MCP calls hitting `database is locked`. The fix the reporter themselves suggested — **WAL mode** — is now always on, because node:sqlite supports WAL on every platform. Under WAL, readers never block on a writer, so the original symptom is gone without relying on any timeout.

**On the magnitude:** that same change deliberately *lowered* `busy_timeout` from 120000ms to **5000ms** and reordered it to be set *first* (before `journal_mode`) so a connection waits out a lock instead of throwing immediately. The `configureConnection` comment spells out why: the old 2-minute wait 'presented as a frozen, hung agent.' A 5-minute timeout makes that worse — up to 5 minutes of an apparently-hung agent on the rare cross-process write contention (e.g. a git-hook `codegraph sync` racing the MCP server's writes). With WAL handling all read concurrency, 5s is ample cushion for the write-write case.

**Also:** the rationale here ('when the WASM fallback is active and WAL mode is not available') describes a state that no longer exists — #282 dropped the better-sqlite3/wasm backends entirely, so node:sqlite is the only backend and always has WAL. And the PR currently conflicts with main, and the added tests use the `jest` API in a `vitest` project (they wouldn't run).

Closing as already-resolved — appreciate you taking a crack at it. 🙏

PR #265, opened 2026-05-21
fix(config): validate languages from shared registry

okbexx:
## Summary
- replace the stale hard-coded language allowlist in `validateConfig` with the shared `LANGUAGES` registry
- add regression coverage that every registered language validates and unknown languages are still rejected

## Why
`LANGUAGES` already includes languages such as `tsx`, `jsx`, `c`, `cpp`, `csharp`, `vue`, `liquid`, `pascal`, and `scala`, but `validateConfig` had a smaller local allowlist. A user config that explicitly listed one of those supported languages could be rejected as invalid.

## Test plan
- [x] `npm test -- --run __tests__/foundation.test.ts -t "supported language|unknown languages"`
- [x] `npm test -- --run __tests__/foundation.test.ts`
- [x] `npm run build`
- [x] `npm test`

### DIFF
--- a/__tests__/foundation.test.ts
+++ b/__tests__/foundation.test.ts
@@ -9,8 +9,8 @@ import * as fs from 'fs';
 import * as path from 'path';
 import * as os from 'os';
 import { CodeGraph } from '../src';
-import { DEFAULT_CONFIG, Node, Edge } from '../src/types';
-import { loadConfig, saveConfig } from '../src/config';
+import { DEFAULT_CONFIG, LANGUAGES } from '../src/types';
+import { loadConfig, saveConfig, validateConfig } from '../src/config';
 import { isInitialized, getCodeGraphDir, validateDirectory } from '../src/directory';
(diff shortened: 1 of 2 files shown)


colbymchenry:
Thanks for this, @okbexx — and sorry for the timing. This was a legitimate fix when you opened it: `validateConfig()` did carry a stale, narrower language allowlist than the shared `LANGUAGES` registry, so configs naming `tsx`/`vue`/`scala`/etc. could be wrongly rejected. (#233 was a parallel attempt at the same fix.)

It's since been overtaken by #285 ("zero-config indexing driven by .gitignore"), merged May 21, which removed `src/config.ts` and `validateConfig()` entirely. CodeGraph no longer has a config file or a `languages` allowlist — indexing is now driven by `.gitignore`, so there's nothing left to validate against, and the file this PR patches no longer exists on `main`.

Closing as obsolete — this reflects the refactor, not the quality of the change. Appreciate the contribution and the regression coverage; hope you'll send more. 🙏

PR #272, opened 2026-05-21
Enhance query parser with negation and exclude filters

krataratha:
Updated query parser to support negation for filters and added excludeFilters.

### DIFF
--- a/src/search/query-parser.ts
+++ b/src/search/query-parser.ts
@@ -3,7 +3,7 @@
  *
  * Splits a raw query like
  *
- *     kind:function name:auth path:src/api authenticate
+ * kind:function name:auth path:src/api authenticate -lang:python
  *
  * into structured filters (kind=function, name="auth", path prefix
  * "src/api") plus the free-text portion ("authenticate") that goes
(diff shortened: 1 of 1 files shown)


colbymchenry:
Thanks for the contribution, and for thinking about negation/exclude — it's a reasonable feature idea. Unfortunately I can't take this as-is:

1. **It doesn't compile.** The new `parseQuery`/`ParsedQuery` is inserted ahead of the existing interface, leaving the original members (`kinds: NodeKind[]`, `languages`, `pathFilters`, `nameFilters`) orphaned after the new function's closing brace, plus a duplicate `parseQuery`.
2. **It breaks the consumer.** It reshapes `ParsedQuery` (arrays → single optional fields) without updating `src/db/queries.ts`, which reads `parsed.kinds/.languages/.pathFilters/.nameFilters` across ~20 query methods.
3. **Feature regression.** `kinds: NodeKind[]` → `kind?` drops multi-value OR'd filtering, a shipped/tested capability (`__tests__/search-query-parser.test.ts` asserts `kind:function kind:method` → `['function','method']`). Those tests aren't updated and would fail.
4. **The headline feature is inert.** `excludeFilters` is parsed but never threaded into the SQL builder, so negation is a no-op.

Doing this properly means keeping the array shape, adding `excludeKinds`/`excludeLanguages`/etc., wiring `NOT IN`/`NOT LIKE` through `queries.ts`, and adding tests — effectively a rewrite — so I'm closing rather than iterating here. If you'd like to tackle the full version with the consumer + tests wired up, I'd be glad to review a fresh PR. Appreciate the effort!

Issue #276, opened 2026-05-21
fs.watch(root, {recursive: true}) registers inotify watch per directory with no exclude — exhausts kernel budget on large repos


mattmorganpdx:
## Bug

On Linux, a single `codegraph serve --mcp` instance can hold hundreds of thousands of inotify watches in a typical monorepo, because the recursive watcher does not exclude any directories at registration time. Two concurrent sessions in the same repo are enough to exhaust the kernel's `fs.inotify.max_user_watches` budget (default 1,048,576 on Ubuntu), after which any other tool that uses inotify (Next.js dev server, JetBrains IDE, etc.) fails with `ENOSPC` / "OS file watch limit reached."

## Environment

- OS: Ubuntu 24.04, kernel 6.17
- Node: v22.21.1
- CodeGraph: 0.7.4 (installed globally via npm; reproduced behavior is identical in current 0.8.0 source)
- Project: large TS + Python monorepo (~200k tracked files including `node_modules/`, `.next/`, `.git/objects/`)

## Repro

1. `codegraph init` at the root of a large repo (one with `node_modules/`, `.git/`, `.next/`, build output, etc.).
2. Start `codegraph serve --mcp` in two sibling AI-agent sessions pointing at the same root.
3. From a third terminal:

```bash
for pid in $(pgrep -f 'codegraph serve --mcp'); do
  total=0
  for fdinfo in /proc/$pid/fdinfo/*; do
    fdnum=\$(basename \"\$fdinfo\")
    [ \"\$(readlink /proc/\$pid/fd/\$fdnum 2>/dev/null)\" = \"anon_inode:inotify\" ] || continue
    w=\$(grep -c '^inotify ' \"\$fdinfo\" 2>/dev/null) || w=0
    total=\$((total+w))
  done
  echo \"\$total \$pid\"
done
```

In my repo each instance held ~440k watches. With three stale instances around (see related orphan-process report) I had 1,048,435 / 1,048,576 watches in use and `next dev` died with:

```
Error [TurbopackInternalError]: Unable to watch …
- OS file watch limit reached.
```

## Root cause

`src/sync/watcher.ts` (compiled at `dist/sync/watcher.js:89`):

```js
this.watcher = fs.watch(this.projectRoot, { recursive: true }, (_eventType, filename) => {
    if (!filename || this.stopped) return;
    // …
    if (!shouldIncludeFile(normalized, this.config)) return;   // filter runs HERE
    // …
});
```

[...]

PR #286, opened 2026-05-21
fix(mcp): reap serve --mcp child when parent is SIGKILL'd (#277)

### LINKED ISSUE #2: git-hook potential issue when codegraph is not installed globally
https://github.com/colbymchenry/codegraph/blob/f0ddfccf47e69737a36b7f4b8681d51e8167e961/src/sync/git-hooks.ts#L46

I guess it should be `npx @colbymchenry/codegraph sync`

Question: Would it make sense to launch `sync` with claude hooks? (Stop or SessionEnd)

### LINKED ISSUE #277: serve --mcp is not reaped when the parent Claude Code process is SIGKILL'd (Linux)
## Bug

When the parent agent process (Claude Code in my case, but any MCP host would do) exits ungracefully — SIGKILL'd by the kernel OOM killer, force-quit, container torn down — the spawned `codegraph serve --mcp` child keeps running indefinitely. Over a few sessions these stale children accumulate, each one holding its own inotify watch set (see companion report on watch-budget exhaustion).

I had three orphaned `codegraph serve --mcp` processes from three different prior sessions still running with combined uptime in the days, until I noticed and killed them manually.

## Environment

- OS: Ubuntu 24.04
- Node: v22.21.1
- CodeGraph: 0.7.4
- Parent: Claude Code CLI (stdio MCP transport)

## Repro

1. Start any MCP host that spawns `codegraph serve --mcp` over stdio (Claude Code, opencode, etc.).
2. Confirm the child is running: `pgrep -af 'codegraph serve --mcp'`.
3. SIGKILL the parent: `kill -9 <parent-pid>`.
4. Check again: the codegraph child is still running, now reparented to PID 1, with stdin still nominally open.

The graceful path (closing the host normally, or SIGTERM'ing it) works correctly — the existing `stdin.on('end' | 'close')` handlers fire and the child exits.

## Root cause

`dist/mcp/index.js:116-121` installs the expected handlers:

```js
process.on('SIGINT',  () => this.stop());
process.on('SIGTERM', () => this.stop());
// When the parent process (Claude Code) exits, stdin closes.
process.stdin.on('end',   () => this.stop());
process.stdin.on('close', () => this.stop());
```

This is correct for graceful parent shutdown, but on Linux, when a parent is SIGKILL'd:

- The child does **not** receive a signal — Linux does not propagate parent death to children by default.
- The child is reparented to PID 1 (or the user's systemd `--user` instance).
- The stdio pipe is half-closed from the parent side; whether the child sees an `end`/`close` event depends on whether the *write* side of the child's stdin was the only writer. In practice, in my reproductions, the child's stdin sits readable and idle and neither handler fires.

Node does not expose `prctl(PR_SET_PDEATHSIG)`, which is the standard Linux mechanism to ask the kernel to deliver a signal when the parent dies.

## Suggested fixes

Pick one or combine:

1. **PPID watchdog.** Cheapest and pure-JS. After startup, `setInterval(() => { if (process.ppid === 1) this.stop(); }, 5000)`. (Or 1000 if you want faster cleanup.) `process.ppid` reflects reparenting on Linux. Works on macOS too. Costs nothing when the parent is alive.
2. **Native `PR_SET_PDEATHSIG` helper.** Tiny addon (or a `prebuildify`'d one) that calls `prctl(PR_SET_PDEATHSIG, SIGTERM)` once at startup. Kernel signals the child the instant the parent dies. Linux-only.
3. **Read-end detection.** Set a heartbeat — if no JSON-RPC traffic from stdin in N seconds *and* `process.ppid === 1`, exit. This is a defensive layer on top of either of the above.

For a stdio-based MCP server I'd recommend (1) as the primary fix — three lines of code, no native build complexity, works everywhere, and the failure mode you're guarding against is rare enough that 5s polling latency is fine.

## Why this matters

- Each orphan retains its inotify watches, file descriptors, and DB connection — they don't just sit idle.
- Without the watchdog the user has no in-band signal that they have accumulated stale instances; they only find out when something else (a dev server, an IDE) fails with `ENOSPC`.
- Force-kills of the parent are not exotic — OOM, laptop sleep timeouts, container restarts, and `pkill -9 claude` all hit this path.

evanclan:
## Summary

Adds a `process.ppid` watchdog to `MCPServer.start()` so a `codegraph serve --mcp` child terminates when its MCP host is force-killed. Resolves [#277](https://github.com/colbymchenry/codegraph/issues/277).

## Problem

The existing shutdown path (`src/mcp/index.ts`) leans entirely on signal handlers and stdin close events:

\`\`\`75:79:src/mcp/index.ts
    process.on('SIGINT', () => this.stop());
    process.on('SIGTERM', () => this.stop());

    // When the parent process (Claude Code) exits, stdin closes.
    // Detect this and shut down gracefully to prevent orphaned processes.
\`\`\`

On Linux that's not enough when the host (Claude Code, opencode, …) is SIGKILL'd by the OOM killer / a \`kill -9\` / a container teardown:

- The kernel does **not** propagate parent death to children.
- The child gets reparented to init/systemd.
- Whether the half-closed stdio pipe surfaces \`end\` / \`close\` to Node depends on whether anyone else still holds the write end. In the reporter's environment it didn't fire — three orphan \`codegraph serve --mcp\` processes were pinned across sessions, each holding ~440k inotify watches (which then transitively tripped [#276](https://github.com/colbymchenry/codegraph/issues/276)'s watch-budget exhaustion in Next.js / IDEs).

## Solution

Capture \`process.ppid\` once at construction, then \`setInterval\` (default 5s, \`.unref()\`'d) to check it. The moment it diverges from the baseline, we know the original parent has died and we tear down cleanly:

\`\`\`text
[CodeGraph MCP] Parent process exited (ppid 9177 -> 1); shutting down.
\`\`\`

Cross-platform: reparenting changes \`process.ppid\` on Linux and macOS; on Windows the value drops to 0 once the parent is gone, which also trips the check.

Knobs:

- \`CODEGRAPH_PPID_POLL_MS\` env var overrides the poll interval (default 5000ms).
- \`CODEGRAPH_PPID_POLL_MS=0\` disables the watchdog entirely — escape hatch for embedded scenarios where the parent deliberately re-parents the server.

\`stop()\` is now guarded by an idempotency flag so the watchdog can't race the existing stdin-close handlers and double-close the SQLite handle / transport.

## Out of scope

- The companion inotify-watch exhaustion ([#276](https://github.com/colbymchenry/codegraph/issues/276)) — this PR removes the orphan-server _multiplier_ on the watch budget, but the underlying \`fs.watch({ recursive: true })\` registering-everything behavior is its own change.
- \`prctl(PR_SET_PDEATHSIG)\` (issue's suggested fix #2) — pure-JS polling is enough for a failure mode that's already rare, no native build dependency needed.

## Test plan

- New \`__tests__/mcp-ppid-watchdog.test.ts\` stands up a four-tier process tree (vitest → wrapper → {stdin-holder, codegraph}) and SIGKILL's the wrapper. The stdin-holder is a long-lived sibling whose \`stdout\` pipe is dup'd into codegraph's \`stdin\`, so the wrapper's death does **not** transitively close codegraph's stdin. That isolates the watchdog from the pre-existing stdin-close path — confirmed by temporarily disabling the watchdog (test fails) vs leaving it in (test passes).
- 5/5 consecutive local runs at ~925ms each, no flakes.
- \`npx vitest run __tests__/mcp-ppid-watchdog.test.ts __tests__/mcp-initialize.test.ts __tests__/mcp-roots.test.ts\` — 6/6 passing.
- Full \`npm test\`: 718 passing, 5 failing in \`git-hooks.test.ts\` / \`watcher.test.ts\` — both pre-existing on \`main\` (verified by re-running on \`origin/main\` HEAD) and unrelated to this change.
- \`npx tsc --noEmit\` clean.

## Related

- #277 — original report (with the suggested PPID-watchdog fix this PR implements)
- #276 — sibling watch-budget exhaustion (orphan servers were the multiplier)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,18 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixed
+- **Orphaned `codegraph serve --mcp` processes after a parent SIGKILL.** When
+  the MCP host (Claude Code, opencode, …) was force-killed — OOM killer, a
+  `kill -9`, a container teardown — the child kept running indefinitely on
+  Linux, holding inotify watches, file descriptors, and the SQLite WAL. The
+  kernel doesn't propagate parent death to children, and the stdin
+  `end`/`close` handlers we relied on don't always fire. The MCP server now
+  polls `process.ppid` and shuts down the moment it changes from the value
+  observed at startup; the poll interval is `CODEGRAPH_PPID_POLL_MS` (default
+  `5000`, `0` disables). Resolves
+  [#277](https://github.com/colbymchenry/codegraph/issues/277).
+
 ### Added
 - **Framework support: Drupal 8/9/10/11** — CodeGraph now detects Drupal
   projects (via a `drupal/*` dependency in `composer.json`) and adds three
--- a/__tests__/mcp-ppid-watchdog.test.ts
+++ b/__tests__/mcp-ppid-watchdog.test.ts
@@ -1,0 +1,168 @@
+/**
+ * PPID watchdog regression test (#277).
+ *
+ * On Linux, when an MCP host (Claude Code, opencode, …) is SIGKILL'd by the
+ * OOM killer / a force-quit / a container teardown, the kernel does NOT
+ * propagate the death to its `codegraph serve --mcp` child. The child gets
+ * reparented to init/systemd, its stdin stays half-open in some
+ * configurations, and the existing `stdin.on('end' | 'close')` handlers
+ * never fire — the server lingers indefinitely, holding inotify watches,
+ * file descriptors, and the SQLite WAL.
+ *
+ * `src/mcp/index.ts` polls `process.ppid` and shuts down the moment it
+ * diverges from the value observed at startup. This test stands up a
+ * four-tier process tree (vitest → wrapper → {stdin-holder, codegraph}) and
+ * SIGKILL's the wrapper. The stdin-holder is a long-lived sibling whose
+ * `stdout` pipe is dup'd into codegraph's `stdin`. After the wrapper dies
+ * the pipe stays open (stdin-holder still owns the write-end), so the
+ * existing stdin close handlers do **not** fire — the only thing that can
+ * terminate codegraph then is the PPID watchdog.
+ *
+ * Windows is excluded — `process.kill(pid, 'SIGKILL')` does not actually
+ * deliver SIGKILL there, and the per-OS reparenting semantics the watchdog
+ * relies on are POSIX-specific.
+ */
+import { describe, it, expect, afterEach } from 'vitest';
+import { spawn, ChildProcessWithoutNullStreams } from 'child_process';
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+
+const BIN = path.resolve(__dirname, '../dist/bin/codegraph.js');
+
+function isAlive(pid: number): boolean {
+  try {
+    process.kill(pid, 0);
+    return true;
+  } catch {
+    return false;
+  }
+}
+
+function waitForExit(pid: number, timeoutMs: number): Promise<boolean> {
+  return new Promise((resolve) => {
+    const start = Date.now();
+    const tick = () => {
+      if (!isAlive(pid)) return resolve(true);
+      if (Date.now() - start > timeoutMs) return resolve(false);
+      setTimeout(tick, 100);
+    };
+    tick();
+  });
+}
+
+describe.skipIf(process.platform === 'win32')('MCP PPID watchdog (#277)', () => {
+  let wrapper: ChildProcessWithoutNullStreams | null = null;
+  let childPid: number | null = null;
+  let stdinHolderPid: number | null = null;
+
+  afterEach(() => {
+    if (wrapper && !wrapper.killed) {
+      try { wrapper.kill('SIGKILL'); } catch { /* already gone */ }
+    }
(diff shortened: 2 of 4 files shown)


colbymchenry:
Merged — thank you @evanclan! 🎉 Genuinely one of the best-documented PRs I've reviewed: the four-tier process-tree test (dup'ing a sibling's stdout into codegraph's stdin so SIGKILL'ing the wrapper *doesn't* close stdin) isolates the watchdog perfectly.

One thing surfaced during validation that needed a fix on your branch before merge: main moved under the PR. Since you opened it, main gained a `--liftoff-only` re-exec (the V8 turboshaft WASM workaround), which inserts an intermediate process between the MCP host and the server. That intermediate outlives the host (it's blocked in `spawnSync`), so the server's own `process.ppid` never changes when the host dies — and the watchdog never fired. Your regression test caught it cleanly: it **failed** on the merged state.

Fix (commit on your branch, credited to you via squash): propagate the host PID across the re-exec via `CODEGRAPH_HOST_PPID`, and have the watchdog poll that PID for liveness in addition to the `process.ppid` divergence check (which still covers the bundled/direct-launch path). Your test now passes *through* the re-exec, so it guards both paths.

Validation on macOS (reparent-to-launchd mirrors the Linux case):
- Your test: green on the merged state.
- Independent A/B (same binary, watchdog on vs off): orphan reaped only with it on; survives 4s+ with it off — confirming the watchdog is precisely what fixes it.
- Full suite: 776 passing, 0 failing. (The git-hooks/watcher failures you noted were environment-specific — they pass on the merged state.)

Thanks again — great contribution. 🙏

PR #97, opened 2026-04-26
docs: cookbook for adding a new language extractor

andreinknv:
> **🔄 Rebased onto refactor #116 (per-language registry).** The cookbook now teaches the post-refactor flow: one new file in `src/extraction/languages/` instead of the previous 3-file mutation across `types.ts` / `grammars.ts` / `CLAUDE.md`. Sections 1, 2, 4, 6, 7, 8 and the reference table are unchanged.

---

## Summary

Adds [`docs/ADDING-A-LANGUAGE.md`](https://github.com/andreinknv/codegraph/blob/docs/adding-a-language/docs/ADDING-A-LANGUAGE.md) — a cookbook for contributors adding a new language extractor. Closes #55.

The doc was prompted by @cfournel's question on #55 (they published `tree-sitter-mql5` and asked how to plug it in), but applies equally to anyone adding HCL/Terraform, R, SQL, dbt, Scala, Vue, or any of the other language requests in the issue tracker. Now there's a single self-serve walkthrough.

## What it covers

1. **Sourcing the wasm grammar** — three real-world paths: already in `tree-sitter-wasms`, pre-built in a GitHub release / npm tarball, or built from source via `tree-sitter-cli`'s bundled wasi-sdk (no Docker / local emcc needed).
2. **Probing the AST** with a 15-line scratch script before writing any extractor code.
3. **Registering the language** — one new file (`src/extraction/languages/<name>.ts` exporting a `LanguageDef`) plus a 2-line registry update; reflects the per-file registry pattern from #116.
4. **Type-check before extraction logic** — catches wiring bugs early.
5. **Two extractor patterns** plugged into the same `LanguageDef`:
   - `LanguageExtractor` config (procedural / OO — Python, Ruby, R)
   - Self-contained extractor class (declarative / template / non-OO — HCL, SQL, Liquid)
6. **NodeKind / EdgeKind mapping table** so contributors don't introduce new kinds unnecessarily (those are cross-cutting).
7. **Test patterns** with `extractFromSource` + the end-to-end CLI smoke recipe.
8. **PR description checklist** including grammar-source provenance, sha256 for vendored wasms, and being honest about constructs the grammar can't parse.

Each section points at concrete extractors in the repo as worked examples — R for the OO path, HCL / SQL / Liquid for the custom-class path, Pascal + DFM for the cross-format case.

## Files changed

| File | Change |
|---|---|
| `docs/ADDING-A-LANGUAGE.md` | New cookbook (~460 lines) |
[...]

PR #288, opened 2026-05-22
feat(julia): language extractor with vendored tree-sitter grammar

hexu2:
## Summary

Adds first-class **Julia (`.jl`)** indexing to CodeGraph using the [tree-sitter-julia](https://github.com/tree-sitter/tree-sitter-julia) grammar.

This PR **extends and supersedes** the groundwork in #244 by @kongdd, with additional handling for common Julia AST shapes seen in real-world packages (including Julia 1.11):

- **Structs without `block` wrappers** — `struct_definition` may use `typed_expression` children instead of `body`/`block`; `extractStruct` now uses `resolveBody` (consistent with interface extraction).
- **Functions without `block` wrappers** — many files use a flat statement list after `signature`; `resolveBody` returns the function node so nested calls are still indexed.
- **One-line definitions** — e.g. `f(x) = expr` via `assignment` handling (not covered in #244).
- **`module` nodes** — `module_definition` mapped to `module` kind for namespacing.
- **`macrocall_expression`** — included in `callTypes`.
- **Vendored WASM** — `tree-sitter-julia.wasm` is committed under `src/extraction/wasm/` (the grammar is **not** in `tree-sitter-wasms@0.1.11`; npm-only loading would fail in CI and offline installs).

Also adds the optional `getName` hook on `LanguageExtractor` (as proposed in #244) and wires it through `extractName`.

## Test plan

- [x] `npm run build`
- [x] `npm test -- --run -t "Julia Extraction"` (11 tests)
- [x] Manual: `extractFromSource` on representative fixtures (structs, one-liners, modules, call sites)
- [ ] `codegraph index` on a Julia project; `codegraph query <symbol>` returns expected definitions and references

## Related

- Supersedes / builds on: #244 (happy to co-author with @kongdd)
- Upstream doc: adding-a-language cookbook (if #97 lands)

## Notes for maintainers

- `tree-sitter-julia` is a **devDependency** only for rebuilding WASM; runtime uses the committed `src/extraction/wasm/tree-sitter-julia.wasm` binary (same pattern as lua/luau/pascal/scala vendoring).


hexu2:
Closing to resubmit with cleaned commit history and description.

PR #291, opened 2026-05-22
fix: add shell: true to spawnSync on Windows for .cmd launcher

### LINKED ISSUE #289: EINVAL
codegraph -v
codegraph: spawnSync D:\Application\dev-app\nvm\v24.11.0\node_modules\@colbymchenry\codegraph\node_modules\@colbymchenry\codegraph-win32-x64\bin\codegraph.cmd EINVAL

ayu-exorcist:
Node.js `spawnSync` requires `shell: true` when executing `.cmd` batch files on Windows; without it the npm thin-installer shim throws `EINVAL` when running `codegraph -v` (or any command) via `npm i -g` or `npx`.

Fixes #289

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,13 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   exports are gone. Existing `.codegraph/config.json` files are simply ignored.
   The `.codegraphignore` marker is no longer supported — use `.gitignore`.
 
+### Fixed
+- **Windows `EINVAL` via npm/npx.** The thin-installer shim used `spawnSync` to
+  launch the bundled `codegraph.cmd` on Windows, which throws `EINVAL` when
+  `shell: true` is omitted (Node.js requires a shell to execute `.cmd` batch
+  files). Windows installs via `npm i -g` or `npx` now work. Resolves
+  [#289](https://github.com/colbymchenry/codegraph/issues/289).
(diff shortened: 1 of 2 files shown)


colbymchenry:
Thanks a lot for this @ayu-exorcist — your diagnosis is spot on, and you caught the `.cmd`/`EINVAL` root cause exactly right. 🙏

We ended up fixing this in #292 (now merged, shipping in 0.9.2) with a slightly different approach, so I'm closing this one as superseded.

**Why we didn't go with `shell: true`:** on Windows, `spawnSync` with `shell: true` and an args array doesn't auto-quote the arguments — it concatenates them into a `cmd.exe` command line. That breaks as soon as a path or arg contains a space (e.g. `C:\Users\me\My Project`) or a shell metacharacter, and routing user-supplied args back through `cmd.exe` re-opens part of the very injection surface the CVE-2024-27980 hardening (the thing throwing `EINVAL`) was added to close.

Instead, #292 skips the `.cmd` entirely on Windows and spawns the bundled `node.exe` directly against the app entry (`lib/dist/bin/codegraph.js`) — which is exactly what the `.cmd` does internally, so there's no behavior change and no shell involved at all. It's verified end-to-end against a real win32-x64 bundle (`npm install` + `codegraph init -i`/`status` on the bundled Node 24).

Really appreciate you taking the time to report and fix this — hope to see more PRs from you. 🙌

PR #299, opened 2026-05-22
feat: add Solidity language support

guiltylotus:
## Summary

- Adds `.sol` file indexing via a vendored `tree-sitter-solidity` WASM (v1.2.13)
- Extracts contracts and libraries as `class`, interfaces as `interface`, structs, enums, functions, modifiers, events, constructors, fallback/receive, state variables, constants, and imports — all with full signature capture
- Registers `solidity` in `LANGUAGES`, `EXTENSION_MAP`, `WASM_GRAMMAR_FILES`, the vendored WASM path check, and the display name map

## What gets indexed

| Solidity construct | Graph kind |
|---|---|
| `contract_declaration`, `library_declaration` | `class` |
| `interface_declaration` | `interface` |
| `struct_declaration` | `struct` |
| `enum_declaration` / `enum_value` | `enum` / `enum_member` |
| `function_definition`, `modifier_definition`, `event_definition` | `method` / `function` |
| `constructor_definition`, `fallback_receive_definition` | `function` (via `visitNode`) |
| `state_variable_declaration` | `variable` with type signature |
| `constant_variable_declaration` | `constant` |
| `import_directive` | `import` |

## Benchmark

| Language | Cost | Tokens | Time | Tool calls |
|---|---|---|---|---|
| Solidity · ~343 files | 8% cheaper | 25% fewer | 10% faster | 29% fewer |

## Notes

- WASM vendored from `tree-sitter-solidity@1.2.13` (same vendoring pattern as Lua, Luau, Pascal, Scala)
- `constructor_definition` and `fallback_receive_definition` have no `name` field in the AST — handled via the `visitNode` hook, naming them `"constructor"`, `"fallback"`, or `"receive"` respectively
- Solidity `external` visibility is mapped to `public` in the graph (it means callable from outside the contract)


colbymchenry:
Closing — Solidity support landed via #667/#1170 (`1441933`), which covered a superset of this PR (inheritance `extends`/`implements` edges, emit/revert/modifier call refs, parameter-aware signatures, tests + real-repo validation) without vendoring a wasm (`tree-sitter-wasms` already ships the ABI-14 Solidity grammar, so the 508 KB binary and its provenance-verification burden weren't needed).

Thank you for opening the first PR here, @guiltylotus — it staked out the mapping (contract/library → class, interface, struct/enum handling) that the landed implementation converged on too, and it's what put Solidity on the roadmap's front burner. If you're interested in another language from the tracker (#648), PRs are very welcome — the `/add-lang` cookbook makes the second one much easier than the first.

Issue #155, opened 2026-05-16
Support for Git Worktree: Share or auto-inherit .codegraph across worktrees


JonasGao:
## Problem

When using **git worktree**, each worktree is an independent working directory. Currently, CodeGraph requires running `codegraph init -i` in each worktree directory separately.

This creates friction for developers who use worktrees for:
- Parallel development on multiple branches
- AI coding agents working on different branches simultaneously
- Quick context switching between features/PRs

## Current Behavior

- `.codegraph/` directory is created per working directory
- It's NOT tracked by git (typically in .gitignore)
- Each new worktree needs manual initialization
- Indexing large projects multiple times is time-consuming

## Desired Behavior

**Option A: Shared Index Repository-wide**
- Store `.codegraph/` in the main git repository directory (not worktree-specific)
- All worktrees reference the same index
- MCP server detects worktree context and uses shared index
- Trade-off: Index reflects one branch's code structure

**Option B: Auto-inherit on Worktree Creation**
- When detecting git worktree context, automatically copy or symlink `.codegraph/` from parent/main worktree
- Or provide a `--inherit` flag for `codegraph init`

**Option C: Git-tracked Index**
- Officially support committing `.codegraph/` to git
- Provide guidance on handling branch differences
- New worktrees automatically inherit via git checkout

## Use Case Example

```bash
# Main repo with CodeGraph initialized
cd ~/project-main
codegraph init -i

# Create worktree for feature branch
git worktree add ~/project-feature feature-123

# Currently: need to manually init again
cd ~/project-feature
codegraph init -i  # <-- This is the friction point

# Desired: worktree auto-detects and reuses/inherits index
```

## Environment

- Git worktree feature (Git 2.5+)
- Common in AI agent workflows (Claude Code, OpenClaw, etc.)

## Related

This is particularly relevant for AI coding agents that use worktrees to work on multiple tasks/branches in parallel.

colbymchenry:
Shipped in #312 (commit 4a4a37d), thanks to @shinju4n.

**Approach:** rather than the originally-proposed shared / auto-inherit / git-tracked options — a single index can't correctly represent multiple branches at once — codegraph now makes the silent failure visible.

When a worktree is nested inside the main checkout (the common case for agent tools that place worktrees under gitignored paths like `.claude/worktrees/<name>/`), running CodeGraph from that worktree used to walk *up* to the main checkout's `.codegraph/` and silently return that tree's code (usually a different branch). Now:

[...]

PR #312, opened 2026-05-22
feat: warn when a query borrows another git worktree's index

shinju4n:
# feat: warn when a query borrows another git worktree's index

Refs #155

## Problem

CodeGraph resolves a project's index by walking up to the nearest `.codegraph/`
(`findNearestCodeGraphRoot`). That walk is unaware of git worktrees.

When a worktree is created **inside** the main checkout — which is exactly what
several agent tools do (e.g. placing worktrees under gitignored paths like
`.claude/worktrees/<name>/`) — a command run from the worktree walks up and
silently resolves the **main checkout's** index. Every query then returns
results from the main tree's code (usually a different branch) rather than the
worktree the user is actually editing. Symbols added or changed only in the
worktree are invisible, and nothing tells the user this is happening.

This is the silent footgun behind #155. (Note: a single shared index *cannot*
be correct for multiple worktrees at once — each can be on a different branch —
so the right primitive is to make the situation visible, then let users opt into
a worktree-local index.)

## What this PR does

Adds **detection + a non-fatal warning**, scoped intentionally small:

- **`src/sync/worktree.ts`** — `detectWorktreeIndexMismatch(startPath, indexRoot)`
  compares the caller's git working-tree root against the resolved index root via
  `git rev-parse --show-toplevel`. Returns `null` (no warning) when: not a git
  repo / git unavailable, the index lives in the caller's own tree, or the index
  root is a plain (non-worktree) ancestor directory. So non-worktree and
  monorepo-subdir layouts are unaffected.
- **`codegraph status`** (CLI) — prints the warning, and adds a
  `worktreeMismatch` field to `--json` output.
- **`codegraph_status`** (MCP tool) — prepends the warning to the status text.

Example (`status` run inside a nested worktree):

```
Project: /repo
⚠ This CodeGraph index belongs to a different git working tree.
  Running in: /repo/wt
  Index from: /repo
Results reflect that tree's code (often a different branch), not this worktree —
symbols changed only here are missing. Run "codegraph init -i" in this worktree
for a worktree-local index.
```

## Tests

`__tests__/worktree-detection.test.ts` — drives **real `git worktree add`**
(no mocking, matching the existing `mcp-roots.test.ts` style):

- flags a nested worktree borrowing the main index
- returns null for same-tree, subdirectory, non-git, and plain-ancestor cases
- `gitWorktreeRoot` reports each tree distinctly
- warning text names both trees and the fix

`tsc` build passes. Verified end-to-end with the bundled FTS5 Node runtime:
the warning fires in a worktree and is absent in the main checkout.

## Deliberately out of scope (happy to follow up per your preference)

- Surfacing the warning on **every** tool (search/context/…), not just status.
- An **opt-in auto-init** path so a nested worktree builds its own local index
  (the natural next step once the detection primitive lands).

I kept the surface area minimal to make review easy — glad to extend in whatever
direction you'd prefer for #155.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,20 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   standalone until it idles out — they never mix versions over the socket.
 
 ### Fixed
+- **Git worktrees no longer silently borrow another tree's index (#155).**
+  When a worktree is nested inside the main checkout — exactly what agent
+  tools that place worktrees under gitignored paths like
+  `.claude/worktrees/<name>/` do — running CodeGraph from that worktree used
+  to walk *up* to the main checkout's `.codegraph/` and silently return that
+  tree's code (usually a different branch). Symbols changed only in the
+  worktree were invisible and nothing told you. Now `codegraph status` (CLI +
+  MCP) calls out the conflict explicitly, and every MCP read tool
+  (`codegraph_search`/`context`/`trace`/`callers`/`callees`/`impact`/
+  `explore`/`node`/`files`) prefixes a one-line notice naming the borrowed
+  index and the fix (`codegraph init -i` in the worktree). Detection is
+  best-effort (no git / not a repo / monorepo subdir → no warning) and runs
+  once per session per start path, so it never costs more than a single pair
+  of `git rev-parse` invocations.
 - **The file watcher no longer exhausts the OS file-watch budget on large
   repos (#276).** It used to register a recursive watch over the *entire*
   project — `node_modules/`, build output, caches and all — and filter only
--- a/__tests__/worktree-detection.test.ts
+++ b/__tests__/worktree-detection.test.ts
@@ -21,6 +21,8 @@ import {
   worktreeMismatchWarning,
   gitWorktreeRoot,
 } from '../src/sync/worktree';
+import CodeGraph from '../src/index';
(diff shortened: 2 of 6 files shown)


colbymchenry:
Merged 🎉 — thanks for this, @shinju4n.

The design judgment is what made it land cleanly. #155 proposed three options (shared index, auto-inherit, git-tracked), and you correctly recognized that the shared/tracked options are *semantically wrong* — a single index can't represent multiple branches at once — and reached for the one primitive that's unambiguously correct: make the silent failure visible. The "indexRoot must itself be a worktree root" guard is a nice touch too — it's what keeps monorepo subdirs and non-git ancestors from generating false warnings, and it's the kind of thing that would be easy to miss.

I extended on top to surface the notice on every MCP read tool (search/context/trace/callers/callees/impact/explore/node/files), with the detection cached once per session per start path so it costs at most one pair of `git rev-parse` spawns per project no matter how many tool calls. The reason: an agent in a worktree almost never calls `status` — it calls search/context/trace — so the dangerous case (codegraph returning confidently-wrong results that the agent trusts and doesn't fall back from) needed to ride inline with the result the agent is actually reading.

Validated on macOS / Linux (Docker) / Windows (Parallels VM); 11/11 worktree tests green on all three.

Really appreciated the explicit "deliberately out of scope" list in the PR description — it made it easy to see exactly where to extend without second-guessing the surface area you'd already considered.

jiwony933:
LGTM~ 👍

PR #326, opened 2026-05-22
docs(readme): link support badges to sections

Bortlesboat:
The support badges at the top of the README currently point to #.

This links them to new Supported Platforms / Supported Agents sections so they work as navigation without changing any claims.


colbymchenry:
Merged — thanks @Bortlesboat! 🎉 Good catch on the dead `(#)` badge links.

I made one small adjustment on your branch before merging: the new sections were landing in the middle of **Get Started** (which pushed *Initialize Projects* / *Uninstall* under them), so I moved them down next to **Supported Languages** to form a tidy Supported Platforms / Agents / Languages cluster, and fleshed them out a bit — a platform/arch table (x64 + arm64, matching the release build matrix) and the agent list (matching the installer registry). Your badge-anchor fix is the core of it and is preserved as-is. The squash commit is authored to you. 🙏

Issue #293, opened 2026-05-22
Fatal process out of memory: Zone


weinierfei:
`
codegraph --version
0.9.1
`

<img width="1135" height="260" alt="Image" src="https://github.com/user-attachments/assets/45ed0a71-9702-4824-ae17-c74419499c28" />

<img width="1026" height="164" alt="Image" src="https://github.com/user-attachments/assets/5a255a39-472b-4aaa-8654-f63efb70026d" />

weinierfei:
Codegraph version 0.9.2 also has the same problem

IreinaMaid150700:
im also getting the same error

colbymchenry:
I will take a look at this and have it fixed tomorrow. Can you guys provide me system specs? Memory specifically. What OS?

weinierfei:
> I will take a look at this and have it fixed tomorrow. Can you guys provide me system specs? Memory specifically. What OS?

<img width="471" height="90" alt="Image" src="https://github.com/user-attachments/assets/bc6632dc-5cf0-4603-b6f1-58badd560ad1" />

quangbahoa:
Yea me too.
Macbook air m3
```
codegraph index  
┌  Indexing project
│
│  ◆ Scanning files — 1,793 found
│  · Parsing code  ██░░░░░░░░░░░░░░░░░░░░░░░  8%

#
# Fatal process out of memory: Zone
#
----- Native stack trace -----

 1: 0x100e7f388 node::NodePlatform::GetStackTracePrinter()::$_0::__invoke() [/Users/me/.codegraph/versions/v0.9.2/node]
 2: 0x102734250 v8::base::FatalOOM(v8::base::OOMType, char const*) [/Users/me/.codegraph/versions/v0.9.2/node]
 3: 0x10103e410 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, v8::OOMDetails const&) [/Users/me/.codegraph/versions/v0.9.2/node]
 4: 0x10177dc84 v8::internal::Zone::Expand(unsigned long) [/Users/me/.codegraph/versions/v0.9.2/node]
[...]

PR #331, opened 2026-05-22
feat: add Nix flake for native Nix/NixOS installation

### LINKED ISSUE #324: Nix flake please?
Installation scripts don't work under Nix, could it be possible to add distribute this tool through a flake?

uxtechie:
## Summary

Resolves #324 — the `curl|sh` install scripts don't work under Nix (FHS paths like `~/.local/bin` aren't on `$PATH`, and Nix's immutable store is incompatible with imperative installers).

This adds a `flake.nix` so Nix/NixOS users can install codegraph natively.

## Changes

| File | Description |
|---|---|
| `flake.nix` | Nix flake with `buildNpmPackage` derivation + dev shell |
| `flake.lock` | Auto-generated lock file (nixpkgs-unstable + flake-utils) |
| `.gitignore` | Added `result` (Nix build output symlink) |

## Details

- Uses **Node.js 24** (matching `build-bundle.sh`'s pinned `v24.16.0`)
- Wrapper injects `--liftoff-only` to prevent the V8 turboshaft WASM Zone OOM (issues #293/#298), matching the existing bundle launcher
- Supports **linux** and **darwin** on both **x86_64** and **aarch64**
- No native addons — pure JS + WASM, so builds on any platform without cross-compilation
- Includes a `devShells.default` for contributors

## Usage

```bash
# Run directly
nix run github:colbymchenry/codegraph

# Install to profile
nix profile install github:colbymchenry/codegraph

# Use in another flake / NixOS config
inputs.codegraph.url = "github:colbymchenry/codegraph";

# Development shell
nix develop github:colbymchenry/codegraph
```

## Verification

- [x] `nix build .#default` — builds successfully
- [x] `./result/bin/codegraph --version` → `0.9.3`
- [x] `./result/bin/codegraph --help` → full command list
- [x] `--liftoff-only` present in wrapper script
- [x] `nix flake check` — all systems evaluate cleanly

## Note for future releases

When `package-lock.json` changes, the `npmDepsHash` in `flake.nix` must be updated. Workflow: set hash to `""`, run `nix build`, copy the correct hash from the error.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,6 @@ test-languages/
 
 nul
 release/
+
+# Nix
+result
--- a/flake.lock
+++ b/flake.lock
@@ -1,30 +1,12 @@
 {
   "nodes": {
-    "flake-utils": {
-      "inputs": {
-        "systems": "systems"
-      },
-      "locked": {
-        "lastModified": 1731533236,
-        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
-        "owner": "numtide",
-        "repo": "flake-utils",
-        "rev": "11707dc2f618dd54ca8739b309ec4fc024de578b",
-        "type": "github"
-      },
-      "original": {
-        "owner": "numtide",
-        "repo": "flake-utils",
-        "type": "github"
-      }
-    },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1779414690,
-        "narHash": "sha256-gOTcX/9MZVMUE0Xvb4IEcv+0TQJkZFNEnL757ljU360=",
+        "lastModified": 1779536132,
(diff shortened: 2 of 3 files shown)


Myzel394:
Can you add this to the official nixpkgs repo?

pshirshov:
Please don't. Keep it as a flake. The tool is evolving fast, nixpkgs takes a lot of time to get something available. Flake solves the problem and does not create heavy maintenance burden.

uxtechie:
Both valid points!

@pshirshov  Agree — a flake is the right approach for a fast-moving project like this. nixpkgs would lag behind constantly.

@Myzel394  A nixpkgs package could make sense down the road, but given the current pace of changes, keeping the hash in sync would be a constant chore. The flake in this PR is self-contained and easy to maintain — whoever wants to use it on Nix can point directly to this repo once it's merged.

For now, this PR is the simplest path: zero changes to existing code, just a flake.nix + lock + one .gitignore line.

pshirshov:
Any chance you could merge that? I use this directly from the P/R and it seems to be working.

PR #333, opened 2026-05-22
feat(cli): extend status JSON output for CI usage

### LINKED ISSUE #329: Add --json flag to codegraph status for CI/scripting
## Problem
The `codegraph status` command outputs human-readable text, making it difficult to use in CI pipelines or scripting contexts where machine-parseable output is needed.

## Solution
Add a `--json` flag to `codegraph status` that outputs structured JSON with fields like:
- `initialized`: boolean
- `indexPath`: string (path to index directory)
- `lastIndexed`: ISO timestamp or null
- `agentCount`: number
- `version`: string

This is consistent with other tools in the agent ecosystem (e.g., `ctx7 skills list --json`) and enables scripting use cases.

## Use case
CI scripts can parse \`codegraph status --json\` to check whether a project is properly initialized before running agents, or to verify index freshness.

12122J:
Closes #329

## Summary
- Extends the existing `codegraph status --json` output with `indexPath`, `lastIndexed`, `agentCount`, and `version`
- Preserves existing JSON fields so current consumers keep working
- Reports uninitialized projects with the same machine-readable CI fields
- Uses a database aggregate for `lastIndexed` instead of scanning file records in the CLI

## Test plan
- `npm run build`
- `node dist/bin/codegraph.js status /tmp/codegraph-status-json-smoke --json`
- `npx vitest run __tests__/sync.test.ts` attempted locally, but this machine is on Node v20.20.0 and the suite requires Node 22.5+ for `node:sqlite`; Git-backed cases also hit sandboxed `git` spawn EPERM here

### DIFF
--- a/__tests__/sync.test.ts
+++ b/__tests__/sync.test.ts
@@ -49,6 +49,14 @@ describe('Sync Module', () => {
     });
 
     describe('getChangedFiles()', () => {
+      it('should report the most recent indexed timestamp', () => {
+        const lastIndexed = cg.getLastIndexedAt();
+
+        expect(lastIndexed).toEqual(expect.any(Number));
+        expect(lastIndexed).toBeGreaterThan(0);
+        expect(new Date(lastIndexed!).toISOString()).toMatch(/^\d{4}-\d{2}-\d{2}T/);
+      });
+
       it('should detect added files', () => {
         // Add a new file
         fs.writeFileSync(
--- a/src/bin/codegraph.ts
+++ b/src/bin/codegraph.ts
@@ -247,6 +247,39 @@ function createVerboseProgress(): (progress: { phase: string; current: number; t
   };
 }
 
+function toIsoTimestamp(timestamp: number | null | undefined): string | null {
+  if (timestamp == null) {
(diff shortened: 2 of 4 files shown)


fjgbue:
Hi @colbymchenry, tagging you in case this got lost. Let me know if you have any feedback!

colbymchenry:
Thanks for this, @12122J — and for engaging on the issue to clarify scope. The `status --json` CI fields just landed in #655 (squashed as `7b62356`): `version`, `indexPath`, and an ISO-8601 `lastIndexed`, plus a `CodeGraph.getLastIndexedAt()` library method. I reused your `MAX(indexed_at)` approach, and you're credited as a co-author on the squash commit.

One change from your PR: we dropped `agentCount`. It didn't map cleanly to the issue's own CI use case ("check init + freshness before running agents") and had two conflicting meanings (configured integrations vs. live daemon sessions), so we left it out rather than guess — happy to revisit it as a focused follow-up if a concrete consumer turns up.

Closing as superseded by #655. Thank you for the contribution! 🙏

PR #338, opened 2026-05-22
fix: allow a drive/filesystem root as the project root

JavierusTk:
## Problem

  `codegraph init W:\` (or any drive root) fails immediately — every file
  is reported as "could not be read". With a large tree the CLI ends with
  e.g. `Indexing failed - all 34,498 files had errors`.

  ## Cause

  `validatePathWithinRoot` and `isPathWithinRoot` (`src/utils.ts`) build the
  containment prefix as `normalizedRoot + path.sep`. A drive root (`W:\`,
  `C:\`) or POSIX root (`/`) already ends with a separator, so the prefix
  becomes `W:\\` / `//`, which no real path starts with. Every file under a
  drive-root project is therefore rejected as a path-traversal escape. In
  the batch reader (`src/extraction/index.ts`) that surfaces as a
  `read_error` ("files could not be read").

  ## Fix

  Add a `rootPrefix()` helper that appends a separator only when one is
  missing, and route all three containment checks through it
  (`validatePathWithinRoot`, `isPathWithinRoot`, and the realpath branch of
  `isPathWithinRootReal`). Behavior is byte-identical for non-drive-root
  paths.

  ## Tests

  New containment-check suite in `__tests__/security.test.ts` covering
  drive/filesystem roots (win32- and POSIX-gated), normal nested roots,
  traversal escapes, and sibling-prefix paths (`cg-root` vs `cg-root-evil`).


JavierusTk:
Will make another PR updated to current repo state

PR #346, opened 2026-05-23
fix(watcher): use chokidar with .gitignore filtering to prevent inotify exhaustion

### LINKED ISSUE #276: fs.watch(root, {recursive: true}) registers inotify watch per directory with no exclude — exhausts kernel budget on large repos

7emotions:
## Summary

Replace `fs.watch(root, {recursive: true})` with **chokidar**, which supports an `ignored` callback that filters directories **before** registering inotify watches. Closes #276.

## Problem

On Linux, `fs.watch({recursive: true})` walks the entire directory tree and registers one inotify watch per directory — including `node_modules/`, `.git/`, `dist/`, `.next/`, etc. Filtering happens only inside the callback, after the watch is already registered. On a monorepo this can consume hundreds of thousands of watches per instance, exhausting the kernel watch budget (`fs.inotify.max_user_watches`) and causing other tools (Next.js, IDEs) to fail with ENOSPC.

## Solution

- **chokidar** replaces raw `fs.watch`. It is a mature (40M+ weekly downloads), pure-JS library with zero native compilation — compatible with CodeGraph's self-contained, no-compile philosophy.
- A `.gitignore`-aware `ignored` callback is passed to chokidar. It loads all `.gitignore` files from the project root upward (using the existing `ignore` package dependency) and hardcodes exclusion of `.codegraph/` and `.git/`.
- chokidar only registers underlying `fs.watch` calls on directories that pass the filter, dropping per-instance watch count from hundreds of thousands to hundreds on monorepos.
- The existing `watch-policy.ts` (WSL2 `/mnt/*` detection, `CODEGRAPH_NO_WATCH`) is preserved.

## Changes

- `package.json`: add `chokidar` dependency (v4, CommonJS compatible)
- `src/sync/watcher.ts`: core fix — replace `fs.watch` with `chokidar.watch` + `ignored` callback

### Unit tests

```
vitest src/sync/watcher.test.ts      → 9/9 ✅
vitest src/sync/watch-policy.test.ts → 8/8 ✅
tsc --noEmit                         → clean ✅
```

### Inotify watch measurement

Real inotify watch counts measured from `/proc/{pid}/fdinfo/` (not directory traversal estimates). Method: register old watcher (`fs.watch` recursive) in-process, wait 3s for kernel registration, read `/proc/self/fdinfo/`; then close, register new watcher (chokidar), wait for `ready`, read again.

**Medium project** (pnpm + Astro blog, ~9,000 dirs):

| | Before | After |
|---|---|---|
| inotify watches | 48,368 | 175 |
| Saved | — | 48,193 (99.6%) |

**Small project** (codegraph repo itself, ~250 dirs):

| | Before | After |
|---|---|---|
| inotify watches | 2,045 | 209 |
| Saved | — | 1,836 (90%) |

Kernel limit on the test machine: 65,536. The old behavior on the blog project alone consumed 74% of the total budget.

### Test limitations (honest)

1. **Single runs, no variance calculation.** If async fs.watch registration did not complete within the 3s timeout, OLD counts may be underestimates.
2. **Only two projects tested.** Scale difference is ~20× between them — far from the ~440k-watch monorepo in #276. Absolute savings will be larger on bigger repos, but the reduction ratio may not be linear.
3. **No repeat trials per project.** One measurement each, no averaging.
4. **No concurrency test.** Single-instance only. Two watcher instances running simultaneously on the same root were not tested (though theoretically no additional resource contention is expected).
5. **Linux only.** macOS (FSEvents) and Windows (ReadDirectoryChangesW) use different kernel APIs and were not tested — but neither platform has the inotify budget problem this PR addresses.
6. **Effectiveness depends entirely on `.gitignore` quality.** If a project does not list `node_modules/` in `.gitignore`, the fix provides little benefit. This is by design — the watcher mirrors git's own exclusion semantics.

### Recommendation for maintainers

If merged, consider a note in the release notes:

> On Linux, the file watcher now respects `.gitignore` to avoid consuming inotify watches on excluded directories (e.g. `node_modules/`, `dist/`, `.git/`). Ensure your `.gitignore` covers these directories. If the watcher still causes issues on very large repos, set `CODEGRAPH_NO_WATCH=1`.

## Related

- Closes #276
- PR #286 (orphan process watchdog) addressed the multiplier but not the root cause

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,20 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   already attached to the old daemon keep using it while new sessions run
   standalone until it idles out — they never mix versions over the socket.
 
+### Fixed
+- **The file watcher no longer exhausts the OS file-watch budget on large
+  repos (#276).** It used to register a recursive watch over the *entire*
+  project — `node_modules/`, build output, caches and all — and filter only
+  after the fact. On Linux that meant hundreds of thousands of inotify watches
+  per project; enough that a second project, or codegraph alongside your editor
+  / `next dev`, could hit the per-user ceiling and fail with "OS file watch
+  limit reached." The watcher now excludes the same directories the indexer
+  ignores (the built-in default-ignore set **plus** your `.gitignore`) *before*
+  registering a watch — so on a repo with a 900-directory `node_modules` the
+  watch count drops from ~1,200 to ~14, even when the project has no
+  `.gitignore`. (Stacks with the shared daemon from #411: one watcher across
+  agents, and now that watcher is small.)
+
 ## [0.9.5] - 2026-05-25
 
 ### Fixed
--- a/__tests__/watcher.test.ts
+++ b/__tests__/watcher.test.ts
@@ -166,6 +166,36 @@ describe('FileWatcher', () => {
 
       watcher.stop();
     });
+
+    it('should not watch node_modules even without a .gitignore (#276/#417)', async () => {
+      // No .gitignore in testDir — exclusion relies on the built-in
+      // default-ignore set the indexer uses (buildDefaultIgnore), which a
+      // .gitignore-only filter would miss.
+      fs.mkdirSync(path.join(testDir, 'node_modules', 'dep', 'lib'), { recursive: true });
+      fs.writeFileSync(path.join(testDir, 'node_modules', 'dep', 'index.ts'), 'export const dep = 1;');
+
+      const syncFn = vi.fn().mockResolvedValue({ filesChanged: 0, durationMs: 0 });
+      const watcher = new FileWatcher(testDir, syncFn, { debounceMs: 200 });
+      watcher.start();
+
+      // Let the watcher settle past any residual crawl events.
+      await new Promise((r) => setTimeout(r, 400));
+      syncFn.mockClear();
+
+      // A source-extension edit INSIDE node_modules must NOT trigger a sync —
+      // the directory was never watched.
+      fs.writeFileSync(path.join(testDir, 'node_modules', 'dep', 'lib', 'extra.ts'), 'export const e = 2;');
+      await new Promise((r) => setTimeout(r, 600));
+      expect(syncFn).not.toHaveBeenCalled();
+
+      // Positive control: a real source edit still triggers sync, proving the
+      // watcher is live (not merely inert).
+      fs.writeFileSync(path.join(testDir, 'src', 'live.ts'), 'export const live = 3;');
+      await waitFor(() => syncFn.mock.calls.length > 0, 5000);
+      expect(syncFn).toHaveBeenCalled();
+
+      watcher.stop();
+    });
   });
 
   describe('callbacks', () => {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1460,6 +1460,7 @@
       "integrity": "sha512-o5a9xKjbtuhY6Bi5S3+HvbRERmouabWbyUcpXXUA1u+GNUKoROi9byOJ8M0nHbHYHkYICiMlqxkg1KkYmm25Sw==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "esbuild": "^0.21.3",
         "postcss": "^8.4.43",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,15 +1,16 @@
 {
   "name": "@colbymchenry/codegraph",
-  "version": "0.9.3",
+  "version": "0.9.4",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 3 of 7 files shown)


colbymchenry:
Thank you, @7emotions! 🙏 chokidar with a pre-registration `ignored` filter was exactly the right call for #276 — pure-JS, no native build, fits the bundled philosophy perfectly.

Merged into `main` 🎉 I built on your branch with a couple of tweaks:

- **Reused the indexer's ignore set** (`buildDefaultIgnore` = built-in default-ignore dirs + your `.gitignore`) instead of a `.gitignore`-only chain — so `node_modules`/build/cache dirs are excluded even when a repo has no `.gitignore` (the #407 default-ignore case), and the watcher can never drift from index scope.
- Used chokidar's passed `stats` instead of a per-path `statSync`, and added a behavioral test (node_modules edits don't sync; source edits do).

Measured the payoff on Linux: on a repo with a 900-directory `node_modules` and no `.gitignore`, the watch count drops from **~1,200 to ~14**. It stacks really nicely with the shared daemon (#411) — one watcher across all agents, and now that watcher is small. Validated on macOS, Linux, and Windows. Your authorship is preserved on the squashed commit.

Thanks again for the fix!

PR #372, opened 2026-05-24
feat(extraction): add Terraform / HCL support

derekclair:
Adds .tf and .tfvars to the extension map and a new block-driven extractor. Each top-level Terraform block becomes a graph node:

  resource T N      -> class       qualified: resource.T.N
  data     T N      -> class       qualified: data.T.N
  module   N        -> module      qualified: module.N
  variable N        -> variable    qualified: var.N
  output   N        -> export      qualified: output.N
  provider N        -> namespace   qualified: provider.N
  terraform { ... } -> namespace   qualified: terraform
  locals  { K=... } -> one constant per K, qualified: local.K

Attribute expressions are walked to emit unresolved references for var.*, local.*, data.*.*, module.*, and provider_type.name chains. Each module block's source string is recorded as an imports reference, enabling module-graph traversal.

The grammar is vendored from @tree-sitter-grammars/tree-sitter-hcl v1.2.0 (ABI 14). Both tree-sitter-terraform.wasm and tree-sitter- hcl.wasm are bundled; copy-assets ships them in dist/.

Sanity-checked on a 240-file Terraform monorepo: 1924 nodes / 1683 edges in 414ms, and a 936-file deployments repo: 3662 nodes / 2723 edges in 871ms.


vladfr:
Nice, I was looking for this.

derekclair:
> Nice, I was looking for this.

I fixed my merge conflicts, sorry for taking so long.

kedar-1:
@derekclair  Please check #1173 , it looks TF support is now included.

PR #400, opened 2026-05-25
fix(cli): resolve visual disconnect on Windows modern terminals

### LINKED ISSUE #398: Title: Inconsistent vertical line characters (| vs │) in codegraph index output
**Description:**
When running the `codegraph index` command, the terminal output uses a mix of standard ASCII pipe characters (`|`) and Unicode box-drawing characters (`│`). This creates a slight visual disconnect in the CLI UI.

**Steps to Reproduce:**
1. Install `codegraph`.
2. Run the `codegraph index` command in a project directory.
3. Observe the indexing progress and completion output in the terminal.

**Actual Behavior:**
The top section of the output uses standard pipes (`|`), while the bottom section switches to Unicode box-drawing characters (`│`).

```text
┌  Indexing project                                        
|                                            
|  * Scanning files - 8 found
|  * Parsing code - done
|  * Resolving refs - done
│
◆  Indexed 8 files
│
●  40 nodes, 32 edges in 258ms
│
└  Done
```

Expected Behavior:
The CLI should use consistent characters for the vertical lines. Ideally, it should use the Unicode box-drawing character │ (U+2502) throughout the entire output to perfectly match the other stylistic characters being used (like ┌, ◆, ●, and └).

Environment:

OS: [Windows 11]

Terminal: [powershell]

Codegraph Version: [0.9.4]

fyuo863:
The CLI output previously forced an ASCII fallback on Windows to prevent mojibake caused by `writeSync(1, ...)` bypassing TTY codepage conversion (as noted in historical issue #168). This caused a visual disconnect between the UI library's Unicode borders and the worker's ASCII lines.
This commit updates `writeStdout` to dynamically detect modern Windows terminals (like Windows Terminal or VS Code) via `WT_SESSION` and `TERM_PROGRAM`. When detected, it routes output through `process.stdout.write`, safely rendering perfect Unicode glyphs (e.g., `│`, `◆`) without mojibake. Legacy Windows terminals and other platforms continue to use the high-performance `fs.writeSync` path.
Fixes #398

### DIFF
--- a/src/ui/shimmer-worker.ts
+++ b/src/ui/shimmer-worker.ts
@@ -13,7 +13,14 @@ import type { ShimmerWorkerMessage } from './types';
 // so UTF-8 bytes hit the console raw and mojibake on OEM codepages.
 // `getGlyphs()` returns ASCII fallbacks on Windows to avoid this (#168).
 function writeStdout(s: string): void {
-  writeSync(1, s);
+  if (
+    process.platform === 'win32' &&
+    (process.env.WT_SESSION || process.env.TERM_PROGRAM === 'vscode')
+  ) {
+    process.stdout.write(s);
+  } else {
+    writeSync(1, s);
+  }
 }
 
 const G = getGlyphs();


inth3shadows:
This detects the right terminals, but it changes the write path without changing the glyphs — so the bars won't actually switch to Unicode.

The worker renders `G = getGlyphs()` (`shimmer-worker.ts:19`), and `getGlyphs()` → `supportsUnicode()` (`glyphs.ts:21`) returns `false` for all `win32` — this PR doesn't touch that. Confirmed on Windows: with `WT_SESSION` set, `getGlyphs().rail` is still `|` (0x7c), while `@clack/prompts` (whose own detection *does* check `WT_SESSION`) renders `S_BAR` as `│` (U+2502). So the frame prints `│` and the worker prints `|` — the #398 mismatch stays.

The write-path change here is needed, but only as the companion to switching the glyphs. `supportsUnicode()` needs the matching branch on the same signal:

```ts
if (process.platform === 'win32') {
  // clack enables Unicode on these; match it so the frame and the bars agree
  return !!(process.env.WT_SESSION || process.env.TERM_PROGRAM === 'vscode');
}
```

`WT_SESSION` / `TERM_PROGRAM==='vscode'` cover the common cases; clack's own list is broader (`CI`, `TERM==='xterm-256color'`, `alacritty`, ConEmu, …), so mirroring it keeps the frame and the bars in lockstep everywhere rather than just in Windows Terminal / VS Code.

With both changes: `getGlyphs()` returns Unicode on modern Windows terminals, and your `writeStdout` change keeps those multi-byte glyphs from mojibake-ing through `writeSync` — legacy consoles still fall back to ASCII (matching clack, which also downgrades there). Either change alone is a no-op or reintroduces #168; they're coupled.

Issue #389, opened 2026-05-25
Enhance enterprise Java/Spring flow coverage: MyBatis XML, config keys, and field-injected bean traces


hajimilaowu:
## Summary

CodeGraph is already very useful on large Java/Spring repositories. In a multi-module Spring MVC + Service + DAO project, it reduced the amount of grep/read exploration substantially. While testing v0.9.4-style flows, I found three related areas where enterprise Java call chains still require falling back to textual search.

This is not a request to replace the current Java/Spring support. The Java/Kotlin interface -> implementation synthesis added in v0.9.4 is useful. These suggestions are about extending that same idea to common enterprise Spring/MyBatis patterns so agents can trace from HTTP entry points through service code to persistence and configuration.

### Current gap

For MyBatis projects, CodeGraph can see the Java DAO/mapper interface and implementation, but the chain stops before the XML SQL statement. A typical call chain is only partially represented:

```text
Controller -> BO/Service -> UserDAO.updateUser
```

The remaining important part still requires grep:

```text
UserDAO.updateUser -> UserDAOMapper.updateUser -> <update id="updateUser"> SQL
```

### Requested behavior

Parse MyBatis mapper XML files and create symbols for statements such as:

```xml
<mapper namespace="com.example.user.dao.UserDAOMapper">
  <select id="getById" resultType="User">...</select>
  <update id="updateUser">...</update>
</mapper>
```

Potential graph nodes/edges:

```text
com.example.user.dao.UserDAO.updateUser
  -> com.example.user.dao.UserDAOMapper.updateUser
  -> SQL statement node / mapper XML location
```

Useful support would include:

- `<select>`, `<insert>`, `<update>`, `<delete>` statement nodes
- `namespace + id` resolution
- `<sql id="...">` fragments and `<include refid="...">` relationships
- table names and maybe affected columns as metadata, where feasible
- `codegraph_trace` and `codegraph_impact` including mapper XML statements

### Why this matters

For Java backend code review, the call chain is incomplete without the mapper XML. Many behavioral and security risks live in the SQL, not just the Java interface.

### Current gap

Spring applications often fail because a configuration key is missing, renamed, or no longer bound to the new configuration model. Today, resolving this requires textual search across Java, YAML, properties, and sometimes external config dumps.

Example Java usage:

```java
@Value("${app.cache.name.user-token}")
private String tokenCacheName;
```

Example YAML:

```yaml
app:
  cache:
    name:
      user-token: "example-service:auth:token"
[...]

Issue #314, opened 2026-05-22
feat: Add Java import resolution to extractImportMappings for cross-module disambiguation


ZanXusV:
## Background

While analyzing CodeGraph's resolution layer for a Java microservices project (1,500+ files / 44,000+ symbols across 8+ Maven modules), I noticed that `src/resolution/import-resolver.ts` handles import parsing for TypeScript/JavaScript, Python, Go, and PHP — but has no corresponding `extractJavaImports` branch for Java.

## Current Behavior

`extractImportMappings` returns `[]` for any Java file, which means:

1. `resolveViaImport` always returns `null` for Java references.
2. Every Java cross-file symbol lookup falls through to `findBestMatch` in `name-matcher.ts`, which uses **file-path proximity as a heuristic** (up to +80 score for shared directory segments).
3. Confidence for multi-candidate Java resolutions is capped at 0.4–0.7, versus 0.9 for languages with import-based resolution.

## Impact: Java Same-Name Symbol Collisions Are Resolved Heuristically

In a real Maven multi-module project, same-name classes in different packages are common — especially converter/mapper/DTO classes that follow naming conventions like `XxxConverter` or `XxxMapper`. A quick query on the project's `.codegraph/codegraph.db` confirms:

| qualified_name | node count | modules spanned |
|---|---|---|
| `FooConverter::convert` | 11 | `dao/converter` + `service/converter` |
| `BarConverter::convert` | 10 | same pattern |
| `BazConverter::convert` | 9 | same pattern |

When a service-layer class calls `FooConverter.convert()`, CodeGraph picks the right node *most of the time* due to path proximity — but this is coincidental, not semantic. If the caller is in a cross-cutting module equidistant from both converters, the wrong node gets chosen, producing incorrect call edges.

## Root Cause

Java import statements carry the full FQN:

```java
import com.example.project.dao.converter.FooConverter;
```

This is exactly the disambiguation signal needed. A simple regex over the caller's source file resolves the ambiguity with certainty — no AST change required, just a new `extractJavaImports` branch parallel to `extractGoImports`.

## Why No Java Import Parsing Today?

My hypothesis (please correct if wrong): Java's import resolution differs architecturally from JS/TS in that Java imports are **type declarations at the top of the file** and don't map to a resolvable *file path* within the project the way `import { Foo } from './foo'` does in JS. The `resolveImportPath` function expects a relative or alias-based path, not a FQN like `com.example.project.dao.converter.FooConverter`.

[...]

PR #412, opened 2026-05-25
feat(jvm): resolve Java/Kotlin imports by fully-qualified name

arttttt:
## Summary

- Wrap top-level declarations of `.kt` / `.java` files in an implicit `namespace` node carrying the file's `package`, so a class `Bar` in `package com.example.foo` is indexed with qualifiedName `com.example.foo::Bar`.
- New `resolveJvmImport` strategy (gated to Java/Kotlin, `imports`-kind refs only) looks up FQNs through the qualifiedName index — `import com.example.foo.Bar` now resolves by package, not by filename, with confidence 0.95.
- Fix `hasAnyPossibleMatch` pre-filter so it doesn't strip JVM FQNs (`com.example.foo.Bar`) before resolution gets a chance — the previous logic only consulted the first segment.

## Motivation

Java's existing filesystem-based path lookup (`import com.example.Foo` → `com/example/Foo.java`) works only because Java enforces class-name = filename. Kotlin doesn't — `Bar` can live in `Models.kt`, `Utils.kt`, or any file under `package com.example`. Top-level functions and extension functions have no class to name a file after at all. The fallback path (name-matcher) finds *some* match for collisions but cannot disambiguate same-name classes across packages — the central failure mode in multi-module Spring / Android codebases (see the `mall` / `halo` notes in the dynamic-dispatch playbook).

This change makes `imports` edges land on the exact target the package declares, which feeds `callers` / `callees` / `context` precision and lays groundwork for binding-precise Dagger2 / Hilt resolution later.

## What's covered

| Case | Test |
|---|---|
| `package com.example.foo` → class qualifiedName `com.example.foo::Bar` | `extraction.test.ts` (Kotlin / Java) |
| Single-segment package | `extraction.test.ts` |
| No package → qualifiedName unchanged (regression guard) | `extraction.test.ts` |
| FQN lookup of class import | `resolution.test.ts` (unit, mocked context) |
| Top-level function import | unit + integration |
| Cross-language Kotlin importing a Java class | unit + integration |
| Name collision across packages — each caller resolves to ITS `Bar` | unit + integration |
| Wildcard `com.example.*` deliberately falls through to name-matcher | unit |
| Non-JVM languages skip the strategy entirely | unit |
| Non-`imports` reference kinds skip the strategy | unit |
| `imports` edge end-to-end, filename ≠ class name | `frameworks-integration.test.ts` |

19 new tests; all existing Kotlin (8), Java (2), Spring (3), pr19 (33), symbol-lookup (9), resolution (23), frameworks (86), frameworks-integration (8) suites stay green.

## Out of scope

- **Wildcard imports** (`import com.example.*`) — name-matcher still handles these.
- **Aliased imports** (`import com.example.Bar as Baz`) — the Kotlin extractor doesn't currently surface the alias; orthogonal to this PR.
- **Companion-object property imports**, **typealias chains** — left for a follow-up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,20 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+- **Java / Kotlin imports now resolve by fully-qualified name.** Extraction
+  wraps every top-level declaration of a `.kt` / `.java` file in a `namespace`
+  node carrying the file's `package` (so a class `Bar` in
+  `package com.example.foo` is indexed with qualifiedName
+  `com.example.foo::Bar`), and `import com.example.foo.Bar` looks the target
+  up through that index — regardless of whether the class lives in `Bar.kt`,
+  `Models.kt`, or a top-level function. Disambiguates same-name classes
+  across packages (the central failure mode of the previous name-matcher
+  fallback in multi-module Spring / Android codebases), works across the
+  Java↔Kotlin interop boundary, and lays groundwork for binding-precise
+  Dagger2 / Hilt resolution. Wildcard imports (`com.example.*`) still go
+  through name-matcher.
+
 ### Fixed
 - **`codegraph index` / `init -i` summary now reports the true edge count.**
   The per-file counter in the orchestrator only saw extraction-phase edges,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,21 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   Java↔Kotlin interop boundary, and lays groundwork for binding-precise
   Dagger2 / Hilt resolution. Wildcard imports (`com.example.*`) still go
   through name-matcher.
+- **Java / C# anonymous classes (`new T() { ... }`) are now extracted as
+  first-class class nodes with their overrides.** Previously, an anonymous
+  subclass returned from a factory or lambda — `return new BaseIter() {
+  @Override int separatorStart(int s) { ... } };` — produced only an
+  `instantiates` edge: the override methods were invisible to the graph and
+  Phase 5.5 interface-impl synthesis had no class to bridge. The anon class
+  now lands as `<TypeName$anon@line>` with an `extends` reference to the
+  named base/interface, scoped under the enclosing method, and its
+  `method_declaration` members become normal method nodes. The interface→impl
+  synthesizer then bridges the base's abstract methods to the anonymous
+  overrides automatically. Concrete effect on `google/guava` (3,227 .java
+  files): 3,608 anonymous classes extracted, +2,534 interface-impl edges
+  reach overrides hidden in `new T() { ... }` blocks (including lambda
+  bodies). An agent investigating `Splitter.SplittingIterator.separatorStart`
+  now sees the four anonymous overrides in its trail without a Read.
 
 ### Fixed
 - **`codegraph index` / `init -i` summary now reports the true edge count.**
--- a/__tests__/extraction.test.ts
+++ b/__tests__/extraction.test.ts
@@ -814,6 +814,38 @@ public class Calculator {
     expect(methodNode).toBeDefined();
     expect(methodNode?.isStatic).toBe(true);
   });
+
+  it('wraps top-level declarations in a namespace from package_declaration', () => {
+    const code = `
+package com.example.foo;
+
+public class Bar {
+    public String greet() { return "hi"; }
+}
+`;
+    const result = extractFromSource('Bar.java', code);
+
+    const ns = result.nodes.find((n) => n.kind === 'namespace');
+    expect(ns?.name).toBe('com.example.foo');
+
+    const cls = result.nodes.find((n) => n.kind === 'class' && n.name === 'Bar');
+    expect(cls?.qualifiedName).toBe('com.example.foo::Bar');
+
+    const greet = result.nodes.find((n) => n.kind === 'method' && n.name === 'greet');
+    expect(greet?.qualifiedName).toBe('com.example.foo::Bar::greet');
+  });
+
+  it('does not wrap when no package is declared', () => {
+    const code = `
+public class Bar {
+    public String greet() { return "hi"; }
+}
+`;
+    const result = extractFromSource('Bar.java', code);
+    expect(result.nodes.find((n) => n.kind === 'namespace')).toBeUndefined();
+    const cls = result.nodes.find((n) => n.kind === 'class' && n.name === 'Bar');
+    expect(cls?.qualifiedName).toBe('Bar');
+  });
 });
 
 describe('C# Extraction', () => {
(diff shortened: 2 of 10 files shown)


arttttt:
## Before/after validation on two Kotlin/Java repos

Indexed both repos with **`main`-build** and this branch, compared SQLite snapshots. Probe-only (deterministic), no agent A/B — adoption-gap caveat from the dispatch playbook still applies.

### Coverage of in-project import statements

| Repo | Files | In-project `import` statements¹ | main resolved | branch resolved | Coverage |
|---|---:|---:|---:|---:|---:|
| `spring-petclinic/spring-petclinic-kotlin` | 38 .kt | 16 | **0** | **16** | 0% → **100%** |
| `macrozheng/mall` | 524 .java | 893 | **0** | **860** | 0% → **96.3%** |

¹ Ground truth: `grep -c "^import <project-prefix>\." src --include="*.{kt,java}"`.

The mall result is the more meaningful one — it's the central failure mode this PR addresses: multi-module Maven where filenames don't sit at `com/macro/mall/Foo.java` paths, so the existing filesystem-based Java lookup misses them entirely.

### Collision disambiguation (mall)

mall has aggressive name collisions — MyBatis-generated examples:

| Name | Variants across packages |
|---|---:|
| `Criteria` | 76 |
| `Criterion` | 76 |
| `GeneratedCriteria` | 76 |
| `OmsOrderDetail` | 2 (admin `dto` vs portal `domain`) |

Sample of 5 collision-targeted imports — each caller-package resolved to the *correct* target-package (no cross-package leak):

```
portal.controller  → portal.domain::OmsOrderDetail   ✓
service.impl       → bo::AdminUserDetails             ✓
portal.service     → portal.domain::OmsOrderDetail   ✓
portal.dao         → portal.domain::OmsOrderDetail   ✓
dao (admin)        → dto::OmsOrderDetail              ✓
```

Spot-check 20 random resolved imports on mall: 20 / 20 plausibly correct on package-flow inspection (mapper→model, service→mapper, controller→dto/common.api).

### Delta accounting — no edge inflation

Edge counts per kind on mall before/after (this is the part I want a sanity check on — the only thing that should change is `contains` for the new namespace nodes):

| kind | main | branch | Δ |
|---|---:|---:|---:|
| `contains` | 20 047 | 20 571 | **+524** (= namespaces, one per packaged file) |
| `references` | 11 036 | 11 036 | 0 |
| `calls` | 10 974 | 10 974 | 0 |
| `imports` | 2 878 | 2 878 | 0 (same count, **target shifted**) |
| `instantiates` | 550 | 550 | 0 |
| `extends` / `implements` | 94 / 50 | 94 / 50 | 0 |
| heuristic `interface-impl` | 310 | 310 | 0 |
| **total** | 45 629 | 46 153 | **+524** |

`imports` edge *count* doesn't change — what changes is `target`: it shifts from a stub `import`-node to the actual project class/interface. The dynamic-dispatch synthesizer (`interface-impl`, 310 edges — matches the playbook's reported `mall` number) is unchanged.

### Concrete `codegraph_node` improvement

For `interface VisitRepository` in petclinic-kotlin:

- **main**: Trail is empty — graph has zero incoming edges to the interface (calls go to its methods, not the interface). Agent would Read files to find usages.
- **branch**: Trail surfaces 4 incoming sources (`OwnerController.kt`, `VisitController.kt`, and their tests), letting the agent answer "who uses VisitRepository?" without Read.

### Caveat

This is a deterministic-correctness measurement. It does not, by itself, demonstrate an agent A/B win — see the trace-adoption discussion in `docs/benchmarks/call-sequence-analysis.md`. The mechanism here improves `callers` / `context` / `trace` precision unconditionally; whether that converts to fewer agent Reads is adoption-gated.

### Side note (not a regression in this PR)

While running these probes I noticed the CLI's post-index report (`X nodes, Y edges in Zms`) undercounts edges — it sums only extraction-phase results (`src/extraction/index.ts:836`), so resolution-phase + synthesizer edges aren't included. On mall, CLI reports `20 047 edges` while the DB has `45 629`. Pre-existing on `main`, orthogonal to this PR — flagging as worth a separate fix.

---

Repro: `https://github.com/spring-petclinic/spring-petclinic-kotlin` + `https://github.com/macrozheng/mall` cloned at depth=1, indexed with the same builds. Raw metrics in `/tmp/eval/{petclinic,mall}-{main,feat}.txt` locally — happy to attach if useful.

arttttt:
Followup on the CLI-undercount note from the previous comment — opened a separate fix: #413.

That PR is off `main` (not stacked on this one — the bug is pre-existing, unrelated to FQN imports). It snapshots `(nodes, edges)` before/after the full `indexAll` pipeline so the post-index summary reports the true totals, including resolution + synthesizer edges. Concrete impact on `macrozheng/mall`: CLI was reporting `20 047 edges`, actual graph had `45 629`.

Regression test included (`__tests__/integration/full-pipeline.test.ts` — verified to fail without the fix, pass with it). Self-contained scope: only `indexAll` is touched; `indexFiles` and `sync` use different fields and weren't affected.

colbymchenry:
Rebased onto main (conflicts in `src/resolution/index.ts` + `__tests__/resolution.test.ts` from C/C++ resolver merge, `__tests__/frameworks-integration.test.ts` from #389 Spring bean trace, CHANGELOG under [Unreleased]). Force-pushed back.

Also fixed a pre-existing test the namespace wrapping broke: `__tests__/resolution.test.ts > Java import disambiguates same-name classes across modules (#314)` was asserting `qualifiedName === 'Handler::use'`, the wrapping moves this to `com.example.web::Handler::use`. Updated the lookup (separate commit so the original author/date stays clean).

## Validation

**Deterministic correctness on 3 real JVM repos:**

| Repo | Files | Namespace nodes | imports→Java/Kotlin class/iface/method |
|---|---:|---:|---:|
| spring-petclinic-kotlin (small) | 38 .kt | 38 | 16 |
| macrozheng/mall (medium) | 524 .java | 524 | 862 |
| google/guava (large) | 3227 .java | 3224 | 5094 |

Disambiguation spot-check on mall: 76 distinct `Criteria` classes preserved, 2 distinct `OmsOrderDetail` (`com.macro.mall.dto::OmsOrderDetail` vs `com.macro.mall.portal.domain::OmsOrderDetail`) correctly separated.

**Agent A/B per the dynamic-dispatch coverage playbook (3 repos × 3 flow prompts × 2 runs/arm × 2 arms = 36 runs, claude-opus, headless, `--strict-mcp-config`):**

| Repo | Arm | dur(s) | tools | Read | Grep | CG |
|---|---|---:|---:|---:|---:|---:|
| petclinic-kt | **with** | **25.8** | **4.2** | **0.0** | **0.0** | 3.2 |
| petclinic-kt | without | 35.3 | 7.2 | 3.2 | 1.8 | 0 |
| mall | **with** | **32.5** | **4.0** | **0.2** | **0.0** | 2.8 |
| mall | without | 47.7 | 7.0 | 2.5 | 1.2 | 0 |
| guava | with | 50.3 | 5.5 | 1.8 | 0.2 | 2.5 |
| guava | without | 44.3 | 6.7 | 2.0 | 2.2 | 0 |

Playbook pass bar (~0 Read/Grep, faster than without): **met on Spring** (the target). Guava is mixed — the residual reads aren't FQN-import related, they're a functional-interface (lambda → SAM) coverage gap: `Splitter.splittingIterator → strategy.iterator(...) → <lambda body>` has a dynamic-dispatch hop that's orthogonal to this PR. See follow-up issue.

## Follow-up commit on top

Added `feat(extraction): surface Java/C# anonymous-class overrides as nodes` (4c7229c). Without it, the lambda-returned anon `new BaseIter() { @Override int separatorStart(int s) {...} }` produced only an `instantiates` edge — Phase 5.5 interface-impl synthesis had no class to bridge. The anon now lands as `<TypeName$anon@line>` with an `extends` reference; its override methods become normal method nodes; the synthesizer bridges automatically.

Concrete effect on guava: 3608 anon classes extracted, +2534 synthesized `calls` edges to overrides hidden in `new T() { ... }` blocks. Agent A/B on guava q2 (CacheBuilder.build) went from 1.5 → 0 Read with this on top.

Tried explore-output ranking tweaks (symbol diversification in `findRelevantContext` + fair-share per-file budget for multi-entry-point clusters) to also close the Splitter q3 gap. They surfaced the right symbols in the explore output but the *agent* did 1 extra Read on q2 anyway — more context was net-negative. Reverted.

**Tests:** 1069 passed, 2 skipped on the rebased tip.

## Verdict

Lands cleanly on Spring (the explicit target of this PR). Guava's residual is a separate lambda-SAM coverage gap — filing a follow-up issue. Squash-merging.

PR #413, opened 2026-05-25
fix(cli): include resolution + synthesizer edges in indexAll report

arttttt:
## Summary

The `X nodes, Y edges` line printed after `codegraph init -i` / `codegraph index` undercounts the graph — often by more than half on repos with heavy cross-file resolution. Cause: the orchestrator's per-file counter (`totalEdges` in `src/extraction/index.ts`) only sees extraction-phase edges; resolution-phase edges (`calls`, `imports`, `extends`, `implements`, `instantiates`, `references`) and synthesizer edges (`interface-impl`, callbacks, etc.) are added afterwards and were invisible to the report.

Fix: snapshot `(nodes, edges)` counts in `indexAll` before/after the full pipeline (extraction → resolution → synthesis → maintenance) and write the true delta back to `result.nodesCreated` / `result.edgesCreated`. Adds a lightweight `QueryBuilder.getNodeAndEdgeCount()` so the snapshot is one round-trip without pulling in `getStats()`' per-kind breakdowns.

## Concrete impact

Indexing `macrozheng/mall` (524 .java files):

| | Reported by CLI | Actually in DB |
|---|---:|---:|
| Before | 20 817 nodes · **20 047 edges** | 20 817 · 45 629 |
| After | 21 341 · **46 153** | 21 341 · 46 153 |

(The node-count delta here is from a separate, already-merged feature wrapping packaged Kotlin/Java declarations in namespace nodes — unrelated; included for completeness.)

The bug is most visible on JVM repos where resolution wires up dense cross-file call / interface→impl / import graphs.

## Scope

- Only `indexAll` is changed. `indexFiles` doesn't run the resolution phase, so its `edgesCreated` was already correct. `sync` reports `nodesUpdated`/`filesAdded`/`filesModified`/`filesRemoved` (different fields), unaffected.
- No public API change. `result.nodesCreated` / `result.edgesCreated` semantics shift from "extraction-phase additions" to "total additions across all phases" — which is what the field name implies and what every consumer (CLI summary, any test asserting on it) actually wants.

## Test

`__tests__/integration/full-pipeline.test.ts` — new case `reports edgesCreated including resolution + synthesizer phases`. Synthetic 30-file cross-file project; without the fix the test fails (`expected 241, received 121` — extraction sees ~half the eventual graph). Verified to pass with the fix applied. All 45 tests in the integration + db-perf + concurrent-locking + foundation suites stay green.

## How I found this

Noticed during the validation runs for #412 — indexed `mall` with the same build twice (different branches), the CLI reported a 524-edge delta while the DB showed the same 524 plus all the resolved edges. Sanity-checking that gap surfaced this pre-existing under-count. Unrelated to #412 — submitting separately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,15 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixed
+- **`codegraph index` / `init -i` summary now reports the true edge count.**
+  The per-file counter in the orchestrator only saw extraction-phase edges,
+  so resolution and synthesizer edges (often >50% of the graph on
+  cross-file-heavy repos like Spring multi-module Java) were missing from
+  the `X nodes, Y edges` line. Snapshotting the DB before/after the full
+  pipeline now reports the actual additions. Example: indexing
+  `macrozheng/mall` previously reported `20 047 edges` while the DB held
+  `45 629`.
 
 ## [0.9.6] - 2026-05-27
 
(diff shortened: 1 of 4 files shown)


colbymchenry:
Rebased onto main (CHANGELOG conflict resolved — entry now under [Unreleased] above the promoted [0.9.6] block) and force-pushed back to the head branch.

**Tests:** 1048 passed, 2 skipped on the rebased tip. The new regression test (`__tests__/integration/full-pipeline.test.ts > reports edgesCreated including resolution + synthesizer phases`) asserts `result.nodesCreated/edgesCreated` match `getStats()` DB counts plus a sanity check that resolution actually contributed beyond bare `contains` edges.

**Scope check:** only `indexAll` touched; `indexFiles` (no resolution phase) and `sync` (uses `nodesUpdated`, not `nodesCreated`) unaffected.

Skipping the agent A/B playbook — pure CLI-summary fix, no retrieval impact. Lands as a `### Fixed` entry under [Unreleased]. Squash-merging.

PR #415, opened 2026-05-25
fix: accept files under Windows drive-root projects (Z:\, C:\)

Kamilslaoui93:
## Summary

`codegraph index` fails on any project located at a Windows drive root (e.g. `C:\`, `Z:\`, or a mapped network drive). Every file is rejected by the path-traversal check, indexing produces zero results, and the log fills with `Path traversal blocked in batch reader` for every file in the project.

## Root cause

`path.resolve('Z:\')` returns `Z:\` — a path that **already** ends with `path.sep`. The existing check appended another separator:

```ts
resolved.startsWith(normalizedRoot + path.sep)
```

…producing `Z:\` for drive roots. No real file path under the drive root starts with `Z:\`, so the check fails for **every** file. For non-root projects (`C:\Users\foo\proj`) the appended separator is correct and the check works; the bug is specific to roots whose resolved form already terminates with `path.sep`.

Same bug existed in `isPathWithinRoot` and `isPathWithinRootReal`.

## Fix

Extract a small `ensureTrailingSep()` helper that appends `path.sep` only when missing, and use it at all three call sites in `src/utils.ts`. Behaviour is unchanged for every non-drive-root project.

## Reproduction

```powershell
PS Z:\> npx @colbymchenry/codegraph index
# ...
[CodeGraph] Path traversal blocked in batch reader { filePath: 'configuration.yaml' }
[CodeGraph] Path traversal blocked in batch reader { filePath: 'automations.yaml' }
# ... repeated for every file, "Indexed 0 files"
```

With this patch applied locally I get:
```
Indexed 768 files
20,783 nodes, 20,015 edges in 2m 51s
```

## Tests

Added 4 cases to `__tests__/security.test.ts` under a new `Path-within-root helpers — Windows drive-root regression` describe block:
- Two Windows-gated cases that exercise the drive-root path (`Z:\`), skipped on POSIX and on Windows runners where `Z:\` isn't mounted.
- Two cross-platform cases asserting that legitimate traversal attempts (`../escape.txt`) are still rejected — guarding against the obvious failure mode of an over-eager fix.

Full `__tests__/security.test.ts` passes locally on Windows (41 passed, 2 POSIX-gated skipped).

## Test plan

- [x] `npx vitest run __tests__/security.test.ts` — 41 passed
- [x] `npx tsc --noEmit` — clean
- [x] End-to-end: `npx codegraph index` on a project at `Z:\` (mapped network drive) — 768 files indexed, 0 path-traversal warnings

### DIFF
--- a/__tests__/security.test.ts
+++ b/__tests__/security.test.ts
@@ -12,7 +12,12 @@ import { describe, it, expect, beforeEach, afterEach } from 'vitest';
 import * as fs from 'fs';
 import * as path from 'path';
 import * as os from 'os';
-import { FileLock, validateProjectPath } from '../src/utils';
+import {
+  FileLock,
+  validateProjectPath,
+  validatePathWithinRoot,
+  isPathWithinRoot,
+} from '../src/utils';
 import CodeGraph from '../src/index';
 import { ToolHandler, tools } from '../src/mcp/tools';
 import { scanDirectory, isSourceFile } from '../src/extraction';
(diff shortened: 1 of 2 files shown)


ZyphrZero:
PR describes all three functions fixed, but only two were actually changed:
- validatePathWithinRoot ✓
- isPathWithinRoot ✓
- isPathWithinRootReal(src/utils.ts:142) not changed — still realRoot + path.sep

isPathWithinRootReal internally calls isPathWithinRoot (fixed), so the logical path check for the drive root will pass; However, the subsequent realpath secondary verification (realPath.startsWith(realRoot + path.sep)) will still reproduce the same bug on the drive root and does not close the project scenario with symlink.

PR #419, opened 2026-05-25
feat(mcp): share one serve --mcp per project across MCP clients (#411)

### LINKED ISSUE #411: Share one `serve --mcp` per project across concurrent MCP clients to avoid N× inotify + N× index cost
## Problem

Running multiple MCP clients against the same project — two Claude Code sessions in different terminals, an agent in a worktree alongside the main checkout, `/loop` plus an interactive session, parallel sub-agents — spawns one `codegraph serve --mcp` per client. Each child independently:

- registers its own recursive inotify watch set on the project root,
- holds its own SQLite connection,
- maintains its own in-memory state,
- duplicates the cost of any incremental re-sync work.

The inotify cost is the load-bearing one. Per #276, a single instance in a monorepo holds ~440k watches; the kernel default budget is 1,048,576 per user. Two concurrent sessions in the same repo already get close to the ceiling; three is reliably over it. Even after #276's per-instance fix lands (prune at registration time, drop recursive `fs.watch`), you still pay N× the (reduced) cost — N agents on the same repo is increasingly the *normal* workflow for AI tooling, not the edge case.

This is also adjacent to #277 (orphan accumulation) and #238 (SQLite lock contention on concurrent tools): both get easier when there is exactly one writer per project.

## What I'd like

One `codegraph` daemon process per project root. Subsequent `serve --mcp` invocations against the same root detect the daemon and act as thin JSON-RPC proxies over a local socket, instead of spinning up their own watcher / DB connection / index.

Concretely:

1. **Lockfile + socket in `.codegraph/`** — e.g. `.codegraph/daemon.sock` (Unix) / named pipe (Windows) and `.codegraph/daemon.pid`.
2. **`serve --mcp` startup**:
   - If lockfile present and the PID is alive and the socket accepts a handshake: become a proxy. Forward stdio JSON-RPC to the daemon socket; forward responses back. No watcher, no DB open, no parser warm-up.
   - Otherwise: become the daemon. Write the lockfile, listen on the socket, do the existing watcher + indexer work.
3. **Reference counting** — daemon tracks attached clients. When the last client disconnects, optionally linger for `CODEGRAPH_DAEMON_IDLE_TIMEOUT` (default e.g. 300s) before exiting, so back-to-back sessions don't re-pay startup cost.
4. **Health / takeover** — proxy handshake includes a version check; if the daemon is from an older binary, proxy can either refuse (forcing the user to restart) or politely ask the daemon to drain and hand off. Stale lockfile (PID gone, or socket connect refused) → take over.
5. **Opt-out** — `CODEGRAPH_NO_DAEMON=1` for users who prefer the current per-client model (or for debugging).

This is roughly the model `gopls`, `rust-analyzer`, and the various LSP servers settled on for the same reason: one process per project root, many editors attached.

## Why this over just landing #276

#276's pruning fix is necessary and I'm not arguing against it — it'll drop per-instance watch counts from ~440k to single-digit thousands, which is a huge win on its own. But:

- N× small is still linear in N. Three agents on the same monorepo is currently a normal config, not an exotic one, and growing.
- DB contention (#238) is independent of watch count and only goes away with a single writer.
- Cold-start indexing is paid N times today.
- Orphan accumulation (#277) compounds: an orphan from a SIGKILL'd parent isn't *just* a leaked process — it's a leaked watch set, FD set, and DB handle that the next session can't reuse.

A shared daemon collapses all four of those into "pay once."

## Out of scope (deliberately)

- **Cross-project daemon.** Keep it per-project; the lockfile lives in `.codegraph/` and is scoped to that root. No global registry to reason about.
- **Worktree sharing.** #155 is the right place for "one index across N worktrees." Composable with this — if/when #155 lands, the daemon naturally serves all worktrees that point at the shared index.
- **Network sockets.** Local socket / named pipe only. No auth surface, no remote access.

## Environment

- OS: Ubuntu 24.04, kernel 6.17 (same as #276 / #277)
- CodeGraph: 0.9.x
- Workflow: typically 2–3 concurrent Claude Code sessions per repo (interactive + background agents + worktrees)

## Related

- #276 — per-instance watch exhaustion (this proposal stacks with that fix, not in place of it)
- #277 — orphan reaping (becomes simpler with single daemon + PPID-of-last-client check)
- #238 — SQLite lock contention on concurrent tools (resolved structurally by single writer)
- #155 — shared index across worktrees (composable: shared index *and* shared process)

mattmorganpdx:
## Summary

- Adds a per-project shared daemon for `codegraph serve --mcp`. The first invocation per project root listens on `.codegraph/daemon.sock` (named pipe on Windows); subsequent invocations attach as thin stdio↔socket proxies. One file watcher, one SQLite handle, one tree-sitter warm-up no matter how many agents you point at the repo.
- Refactors `MCPServer` into transport-pluggable layers (`engine` / `session` / `transport`) without changing the public API: `new MCPServer(path).start()` still works exactly as before. Daemon-vs-proxy-vs-direct is decided at `start()` time.
- `CODEGRAPH_NO_DAEMON=1` opts out (restores the pre-change per-client behavior).
- `CODEGRAPH_DAEMON_IDLE_TIMEOUT_MS` (default 300s) tunes the linger window after the last client disconnects.

Closes #411. Composable with the per-instance fixes in #276 (pruning) and the SQLite-contention angle in #238 (one writer per project naturally), but neither is required.

## Design

```
serve --mcp
  └─► daemon-opt-out?          ── yes ─► direct (today's behavior)
  └─► .codegraph/ reachable?   ── no  ─► direct
  └─► tryAcquireDaemonLock     ── acquired? ─► become daemon
                               └─ taken?  ─► runProxy(socket)
                                              ├─ proxied  ── stdio↔socket pipe
                                              ├─ version mismatch ── fall back to direct
                                              └─ unreachable + pid dead ── clear stale, retry
```

The daemon emits a single JSON hello line (`{codegraph, pid, socketPath, protocol}`) on every connection so a proxy can verify the version before piping any application bytes. Same-version → transparent pipe (server-initiated `roots/list` requests flow through unchanged). Mismatch → proxy falls back to direct mode, daemon stays untouched for older clients.

The daemon refcounts connected sessions (its stdio launcher + N socket-attached proxies). When all disconnect, idle timer arms; new connections cancel it. PPID watchdog is deliberately disabled in daemon mode (the daemon outlives its launcher by design); direct + proxy modes keep it.

## Files

- `src/mcp/transport.ts` — `JsonRpcTransport` interface; `StdioTransport` gains an optional `onClose` (so the daemon's launcher session can detach without exiting the daemon); new `SocketTransport`.
- `src/mcp/engine.ts` — shared CodeGraph + watcher + ToolHandler, idempotent init.
- `src/mcp/session.ts` — per-connection initialize / tools/list / tools/call state machine (extracted from `MCPServer`).
- `src/mcp/daemon.ts` — listener, hello line, lockfile dance, ref count + idle timer, graceful shutdown.
- `src/mcp/daemon-paths.ts` — socket / pidfile path resolution (POSIX socket length-limit fallback, Windows named-pipe hash).
- `src/mcp/proxy.ts` — stdio↔socket pipe + PPID watchdog.
- `src/mcp/version.ts` — package version for hello-line handshake.
- `src/mcp/index.ts` — `MCPServer` now dispatches direct / daemon / proxy at `start()`. Public API unchanged.

## Test plan

- [x] Full pre-existing suite still passes unmodified — 815 tests in 36 files.
- [x] New `__tests__/mcp-daemon.test.ts` (5 tests) — second client attaches as proxy; `CODEGRAPH_NO_DAEMON=1` opts out; stale-pidfile takeover; idle-timeout exit; version-mismatch fallback to direct mode.
- [ ] Manual: two `claude code` sessions in same monorepo share daemon (count inotify watches with `find /proc -name 'fdinfo' | xargs grep -l inotify` before/after).
- [ ] Manual: cross-platform smoke — Windows named pipe path (see `.parallels` recipe in CLAUDE.md).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,21 +11,28 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Added
 - **Shared MCP daemon — running multiple AI agents in the same project no
-  longer multiplies the inotify, SQLite, and indexing cost.** The first
-  `codegraph serve --mcp` per project becomes a per-project daemon listening
-  on `.codegraph/daemon.sock` (named pipe on Windows). Subsequent invocations
-  for the same project attach as thin stdio↔socket proxies — one file
-  watcher, one SQLite connection, one tree-sitter warm-up no matter how many
-  Claude Code / Cursor / Codex / opencode sessions you point at the repo.
-  Two concurrent sessions on a large monorepo used to consume ~880k of the
-  Linux 1,048,576 per-user inotify budget; they now share ~440k. The daemon
-  lingers for `CODEGRAPH_DAEMON_IDLE_TIMEOUT_MS` (default 300s) after the
-  last client disconnects so back-to-back sessions don't repay startup cost.
-  Resolves issue #411.
-- **`CODEGRAPH_NO_DAEMON=1` — opt out of the shared daemon.** Restores the
-  pre-issue-#411 behavior of one independent server process per client.
-  Useful for debugging or for environments that don't permit local
-  IPC sockets.
+  longer multiplies the file-watch, SQLite, and indexing cost.** Point more
+  than one `codegraph serve --mcp` at a project (two Claude Code windows, an
+  agent in a git worktree, `/loop` alongside an interactive session, parallel
+  sub-agents) and they now share **one** background daemon per project: a
+  single file watcher (one inotify set on Linux), one SQLite connection, and
+  one tree-sitter warm-up — instead of N independent copies. Measured on Linux:
+  three agents register **~3× fewer inotify watches** sharing one watcher
+  versus three standalone servers. Resolves issue #411. (Composable with the
+  per-watcher pruning in #276/#346 — that shrinks each watch set; this shares
+  one across agents.)
+- The daemon runs as a **detached background process** that outlives any single
+  session, so closing one editor or terminal never severs the others. Each
+  `serve --mcp` your agent host launches is a thin stdio↔socket proxy to it (a
+  Unix-domain socket, or a named pipe on Windows). When the last client
+  disconnects the daemon lingers for `CODEGRAPH_DAEMON_IDLE_TIMEOUT_MS`
+  (default `300000`) so back-to-back sessions skip the startup cost, then exits
+  and removes its lockfile — an OOM-killed or force-quit host can't leak it.
+- **`CODEGRAPH_NO_DAEMON=1`** opts out, restoring one independent server per
+  client (handy for debugging or sandboxes that disallow local IPC sockets).
+  The daemon is also version-pinned: after you upgrade codegraph, sessions
+  already attached to the old daemon keep using it while new sessions run
+  standalone until it idles out — they never mix versions over the socket.
 
 ## [0.9.5] - 2026-05-25
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,26 @@ a [GitHub Release](https://github.com/colbymchenry/codegraph/releases) tagged
 This project follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+- **Shared MCP daemon — running multiple AI agents in the same project no
+  longer multiplies the inotify, SQLite, and indexing cost.** The first
+  `codegraph serve --mcp` per project becomes a per-project daemon listening
+  on `.codegraph/daemon.sock` (named pipe on Windows). Subsequent invocations
+  for the same project attach as thin stdio↔socket proxies — one file
+  watcher, one SQLite connection, one tree-sitter warm-up no matter how many
+  Claude Code / Cursor / Codex / opencode sessions you point at the repo.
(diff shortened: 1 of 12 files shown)


colbymchenry:
Thanks so much for this, @mattmorganpdx — one of the most thoughtful contributions we've had. The `engine` / `session` / `transport` split is genuinely clean, and per-connection transports neatly sidestep the JSON-RPC id-collision problem I'd been worried about.

Merged into `main` 🎉 A few things changed during integration, all building on your design:

- **Detached daemon + always-proxy.** The in-process daemon shared the first host's process group (so closing that terminal severed the other clients) and, with the watchdog disabled in daemon mode, orphaned on host SIGKILL — which regressed #277. The daemon is now a detached background process and every `serve --mcp` is a thin proxy to it (the proxy keeps the watchdog), reaped by client-refcount + idle timeout.
- **Atomic lockfile.** Closed a concurrent-startup race (an empty-file window) that could let two daemons bind on simultaneous launches.
- **realpath root canonicalization** so a client with a symlinked cwd/rootUri still converges on the same socket.

Validated on macOS, Linux (Docker — 3× fewer inotify watches for 3 agents), and Windows (named pipes); an A/B confirmed daemon mode returns byte-identical tool output vs direct mode with no latency regression. Your authorship is preserved on the squashed commit.

Thank you again — this is a real quality-of-life win for anyone running multiple agents on one repo. 🙏

PR #433, opened 2026-05-26
fix(installer): preserve symlinks when writing agent config files

0x1306a94:
## Summary
- `atomicWriteFileSync` uses a write-tmp-then-rename pattern, which replaces the directory entry (inode) and breaks symlinks
- Added `resolveSymlink()` helper that checks `lstatSync` + `isSymbolicLink()` and resolves via `realpathSync` (with manual `readlinkSync` fallback for dangling symlinks)
- `atomicWriteFileSync` and `removeMarkedSection` now resolve symlinks before writing/unlinking, so shared `AGENTS.md` `CLAUDE.md` via symlink survives reinstall

## Motivation
Users may share a single AGENTS.md across multiple agents via symlinks:

```bash
ln -sf ~/AGENTS.md ~/.claude/CLAUDE.md
ln -sf ~/AGENTS.md ~/.codex/AGENTS.md
```

Previously, running `codegraph install --target claude,codex --yes` would replace these symlinks with regular files, breaking the shared setup. This fix ensures symlinks are followed and the real target file is written to.

## Test plan
- [x] 6 new symlink tests pass: write-through-symlink, marked section replace/remove through symlink, dangling symlink creation, non-symlink normal path
- [x] All 88 installer-targets tests pass with no regressions

### DIFF
--- a/__tests__/installer-targets.test.ts
+++ b/__tests__/installer-targets.test.ts
@@ -22,6 +22,7 @@ import { ALL_TARGETS, getTarget, resolveTargetFlag } from '../src/installer/targ
 import { uninstallTargets, refreshTargets } from '../src/installer';
 import { upsertTomlTable, removeTomlTable, buildTomlTable } from '../src/installer/targets/toml';
 import { cleanupLegacyHooks, writePromptHookEntry, removePromptHookEntry } from '../src/installer/targets/claude';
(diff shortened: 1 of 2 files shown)


0x1306a94:
@colbymchenry

0x1306a94:
@colbymchenry Can this PR be considered for merging?

PR #444, opened 2026-05-26
fix(sync): cross-check tracked files exist on disk after git fast path

### LINKED ISSUE #443: [Bug] Deleted files still visible via `codegraph files` after `codegraph sync`
## Description

After running `codegraph sync`, files that have been deleted from disk still appear in the output of `codegraph files`.

## Expected Behavior

After `codegraph sync`, files that no longer exist on disk should not appear in `codegraph files`. Sync should clean up their records from the index.

## Actual Behavior

Deleted files continue to appear in `codegraph files` output. Their records in the `files` table are never purged.

## Root Cause

The `sync()` method in `src/extraction/index.ts:1208` has two code paths:

### Git Fast Path (taken when git is available)

Flow:
1. Calls `getGitChangedFiles()` (line 1225), which runs `git status --porcelain --no-renames`
2. Only processes files that git reports as changed:
   - `deleted` → calls `this.queries.deleteFile(filePath)` (lines 1233-1239)
   - `modified` + `added` → reads file content and compares hashes (lines 1246-1268)
3. **Does NOT cross-check all DB-tracked files against the actual filesystem**

### Fallback Full Scan Path (taken when git is unavailable or fails)

1. Calls `scanDirectory()` to list files actually on disk (line 1271)
2. Iterates all DB-tracked files and checks if each still exists on disk (lines 1281-1287)
3. Missing files → calls `this.queries.deleteFile()` to clean up

**The git fast path's flaw: it only acts on git-reported changes and never verifies whether DB-tracked files still exist on disk.**

### Triggering Scenarios

**Scenario A: File deletion was committed to git**

1. File `src/old.ts` is indexed by `codegraph index`
2. File is deleted and the deletion is committed to git
3. `git status --porcelain` → clean (no changes)
4. `sync()` → git fast path → `deleted = []` → nothing happens
5. DB record for `src/old.ts` persists indefinitely

**Scenario B: File was indexed, then added to .gitignore and deleted**

1. File is indexed before it was gitignored
2. User adds the file to `.gitignore` then deletes it from disk
3. `git status --porcelain` may not report the file (git no longer tracks it)
4. `sync()` → same result, DB record remains

## Steps to Reproduce

```bash
# 1. Initialize and index project
cd my-project
codegraph init -i
codegraph index

# 2. Confirm file is indexed
codegraph files | grep old.ts
# → src/old.ts (typescript, 12 symbols)

# 3. Delete file and commit
git rm src/old.ts
git commit -m "remove old.ts"

# 4. Run sync
codegraph sync
# → Already up to date   ← deletion not detected

# 5. Deleted file still shows up
codegraph files | grep old.ts
# → src/old.ts (typescript, 12 symbols)  ← should not be here
```

## Environment

- CodeGraph version: all (present since git fast path was introduced)
- Platform: all (logic bug, not platform-specific)

## Proposed Fix

After the git fast path finishes processing git-reported changes, add a disk-existence cross-check:

```typescript
// In ExtractionOrchestrator.sync(), after processing git changes (~line 1268),
// cross-check remaining DB-tracked files against the filesystem
const allTracked = this.queries.getAllFiles();
for (const tracked of allTracked) {
  const fullPath = path.join(this.rootDir, tracked.path);
  if (!fs.existsSync(fullPath)) {
    this.queries.deleteFile(tracked.path);
    filesRemoved++;
  }
}
```

This mirrors the same cross-check that the fallback path already does (lines 1281-1287). The cost is O(n) stat() calls on tracked files, which is acceptable for typical codebase sizes.

## Code Locations

- **Bug location**: `src/extraction/index.ts` — `ExtractionOrchestrator.sync()` method, git fast path (lines 1227-1268)
- **Correct behavior (reference)**: Same file, fallback path (lines 1281-1287)
- **File deletion method**: `src/db/queries.ts` — `QueryBuilder.deleteFile()` (lines 1163-1171)
- **`codegraph files` query**: `src/db/queries.ts` — `QueryBuilder.getAllFiles()` (lines 1187-1193)

FurryWolfX:
### Title
fix(sync): cross-check tracked files exist on disk after git fast path
### Overview
Fixes #443.
### Changes
| File | Type | Description |
| --- | --- | --- |
| src/extraction/index.ts | fix | After git fast path sync, verify remaining DB-tracked files still exist on disk; delete orphaned records |
### Commits
- `3c6199e` fix(sync): cross-check tracked files exist on disk after git fast path
### Breaking Changes
None.

### DIFF
--- a/src/extraction/index.ts
+++ b/src/extraction/index.ts
@@ -1266,6 +1266,20 @@ export class ExtractionOrchestrator {
           filesModified++;
         }
       }
+
+      // Cross-check: files tracked in the DB but not reported by git status may
+      // still have been removed — e.g. the deletion was committed to git, or
+      // the file was added to .gitignore after indexing. Verify remaining DB
+      // records against the filesystem to keep the index in sync. Mirrors the
+      // same check in the fallback full-scan path below.
+      const remainingTracked = this.queries.getAllFiles();
+      for (const tracked of remainingTracked) {
+        const fullPath = path.join(this.rootDir, tracked.path);
+        if (!fs.existsSync(fullPath)) {
+          this.queries.deleteFile(tracked.path);
+          filesRemoved++;
+        }
+      }
     } else {
       // === Fallback: full scan (non-git project or git failure) ===
       const currentFiles = new Set(scanDirectory(this.rootDir));


FurryWolfX:
oh, 0.9.5 has been fixed

PR #453, opened 2026-05-26
feat(resolution): add C/C++ include path resolution

g122622:
Add full import resolution pipeline for C and C++ #include directives, connecting extracted import nodes to actual header files in the project.

- Add C/C++ extension resolution (.h, .hpp, .hxx, .cpp, .cc, .cxx)
- Add system header filtering with ~80 C and ~80 C++ stdlib headers
- Add extractCppImports() for #include import mapping extraction
- Add compile_commands.json parsing for -I/-isystem include directories
- Add heuristic include dir discovery (include/, src/, lib/, api/)
- Add resolveCppIncludePath() for include directory search
- Add C/C++ built-in symbol filtering (printf, malloc, std::*, etc.)
- Wire getCppIncludeDirs into ResolutionContext
- Add 13 new tests for C/C++ import resolution and extraction

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,28 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
-### Added
+- **C/C++ `#include` resolution — bare-basename includes now connect to the
+  actual header file, not a phantom import node (#453).** Path-prefixed
+  includes (`#include "common/args.h"`) already resolved via file-path
+  suffix matching, but bare-basename includes (`#include "uint256.h"` from a
+  caller in another directory) used to leave only a phantom edge to a
+  floating `import` node owned by the including file. The resolver now walks
+  C/C++ include search directories — pulled from `compile_commands.json`
+  (`-I`/`-isystem` flags) when present, otherwise discovered by probing
+  conventional dirs (`include/`, `src/`, `lib/`, `api/`, `inc/`) plus any
+  top-level directory containing `.h`/`.hpp` files — and resolves the
+  include to a real file node, producing a true file→file `imports` edge.
+  System headers (`<stdio.h>`, `<vector>`, `<iostream>`, ~80 C and ~80 C++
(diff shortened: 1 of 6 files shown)


colbymchenry:
Huge thanks for this @g122622 — C/C++ include resolution was a real gap and the bones of this PR were solid. The `loadCppIncludeDirs` design (compile_commands.json first, then heuristic probing of `include/`/`src/`/`lib/`/`api/`/`inc/`), the stdlib header list, and the built-ins concept were all the right shape.

I pushed a follow-up review commit (5f6830f) before merging to fix two things I caught with an A/B on bitcoin-core (1,989 files) + tinyxml2:

- **Wired the include-dir scan into the pipeline.** Probed in isolation the scan resolved paths correctly, but during indexing the `imports` reference flowed through `resolveViaImport`'s symbol-lookup branch (matched `extractCppImports`' basename-without-ext `localName` via `.startsWith`, then tried to find a symbol named like the file extension and failed). Added a C/C++ imports branch that resolves directly to the file node. Result on bitcoin: file→file imports 6,027 → 8,086 (**+34%**).
- **Guarded the built-ins filter with `hasAnyPossibleMatch`.** Unconditional filtering of `CPP_BUILT_INS`/`C_BUILT_INS` killed legitimate edges to user-shadowed names (bitcoin's `mp::move`, custom allocators, stream wrappers — `move() 1,179 → 0`, `free() 33 → 0`, `write() 149 → 7`). Now it only filters when no user-defined symbol with that name exists. The `std::` prefix stays unconditional. After: all the user-shadowed names preserved at main's edge counts; the genuine `std::move`-binds-to-`mp::move` false positives still drop (−2,154 calls, correctly).

Also added an end-to-end test that exercises `#include "X.h"` → file→file edge through the real indexing pipeline so the wiring can't silently regress again, a test documenting the cross-language `.h` heuristic (Obj-C dirs are intentionally allowed as C/C++ include dirs), dropped a duplicate `'FILE'`, and put a CHANGELOG entry under `[Unreleased]` with the measured numbers. Squash-merging — your authorship stays on the commit. 🙌

PR #479, opened 2026-05-27
test(vitest): unblock subprocess MCP tests on Node >= 25 dev machines (#478)

### LINKED ISSUE #478: MCP integration tests fail on Node ≥25 dev machines (spawned servers blocked by version check)
## Symptom

On a developer machine running Node.js 25 or later, `npm test` against `main` reports:

```
Test Files  4 failed | 44 passed (48)
     Tests  13 failed | 1019 passed | 2 skipped (1034)
```

All 13 failures are in subprocess-based MCP integration tests:

| File | Failing tests |
|---|---|
| `__tests__/mcp-daemon.test.ts` | 7 |
| `__tests__/mcp-initialize.test.ts` | 2 |
| `__tests__/mcp-ppid-watchdog.test.ts` | 1 |
| `__tests__/mcp-roots.test.ts` | 3 |

The failure mode is always the same: every assertion that waits for an MCP message from the spawned child times out with `Timed out. Messages so far: []`. The PPID-watchdog test instead fails with `expected false to be true` (the `kill(pid, 0)` probe never sees the child exit because the child died on startup, not on SIGKILL).

## Root cause

`src/bin/codegraph.ts:65–71` hard-blocks the CLI when `nodeMajor >= 25`:

```ts
if (nodeMajor >= 25) {
  process.stderr.write(buildNode25BlockBanner(nodeVersion) + '\n');
  if (!process.env.CODEGRAPH_ALLOW_UNSAFE_NODE) {
    process.exit(1);
  }
}
```

Each affected test spawns `node dist/bin/codegraph.js serve --mcp …` with the test process's own Node binary (`process.execPath`). On Node ≥25, that child writes the unsupported-version banner and exits before the test sends `initialize`, so every subsequent `waitForMessage` times out.

The block exists for good reason (Node 25's V8 turboshaft WASM JIT Zone allocator bug, issue #81), but the test scenarios — short-lived servers, no production indexing — won't hit the OOM the block protects against. The CLI itself should still enforce the block for end users.

## Confirmed fix

Setting `CODEGRAPH_ALLOW_UNSAFE_NODE=1` in the environment vitest inherits makes all 13 tests pass on Node 26.x:

```
$ CODEGRAPH_ALLOW_UNSAFE_NODE=1 npx vitest run __tests__/mcp-roots.test.ts
 Test Files  1 passed (1)
      Tests  3 passed (3)
```

```
$ CODEGRAPH_ALLOW_UNSAFE_NODE=1 npx vitest run __tests__/mcp-ppid-watchdog.test.ts
 Test Files  1 passed (1)
      Tests  1 passed (1)
```

Since each spawning test already builds its child env via `{ ...process.env, ... }` (or inherits process.env when no `env` is passed), setting the variable once in `vitest.config.ts` propagates it to every spawned subprocess without modifying any individual test file.

## Impact

- CI on Node 22/23 is unaffected (the block only fires on ≥25).
- New contributors on a recent macOS/Linux box (Node 25+ shipped via Homebrew / nvm latest) currently get a noisy `npm test` baseline and have to discover the env-var workaround to verify their changes. The note in `CLAUDE.md` about pre-existing Windows failures doesn't mention this macOS/Linux case, so the failures look unrelated to the underlying Node version.

## Proposed fix

Set `CODEGRAPH_ALLOW_UNSAFE_NODE=1` via `vitest.config.ts`'s `test.env` so it propagates into every spawned subprocess and the suite is green on any supported (or newer) Node. PR to follow.

eddieran:
## Summary

Closes #478.

On a developer machine running Node.js 25+, `npm test` against `main` reports **13 failures across 4 files** (mcp-daemon, mcp-initialize, mcp-ppid-watchdog, mcp-roots). Every failure is a subprocess test: each one spawns `dist/bin/codegraph.js serve --mcp` with `process.execPath` and waits for a JSON-RPC response. On Node ≥25 the spawned child hits the CLI's hard-block (`src/bin/codegraph.ts:65`) and exits before responding, so `waitForMessage` times out.

The CLI's block exists to keep end users off Node 25's V8 turboshaft WASM JIT Zone bug (#81). Test scenarios don't actually exercise that path — they spin up short-lived servers, exchange a few JSON-RPC frames, and tear down — so the right thing is to opt the **test runner's spawned children** out of the block, while leaving the runtime guard intact for everyone else.

## Change

One added field in `vitest.config.ts`:

```ts
env: { CODEGRAPH_ALLOW_UNSAFE_NODE: '1' },
```

Vitest puts this into `process.env` for the test process. Every affected test either (a) spawns without `env` (default = inherit `process.env`) or (b) spawns with `env: { ...process.env, ... }`, so the variable flows into the child unchanged. No test file or production source is modified.

## Validation

On Node 26.0.0 (this dev box), before:

```
Test Files  4 failed | 44 passed (48)
     Tests  13 failed | 1019 passed | 2 skipped (1034)
```

After:

```
Test Files  48 passed (48)
     Tests  1032 passed | 2 skipped (1034)
```

The runtime guard is still enforced when `codegraph` is invoked directly — `node dist/bin/codegraph.js status` on Node 26 still prints the banner and exits with code 1.

## Notes for review

- CI on Node 22/23 is unaffected — the guard doesn't fire there, so the variable is a no-op.
- The `node-version-check.test.ts` unit tests (which assert the banner content) are unrelated and continue to pass — they call the banner builders directly, not via spawn.
- `CLAUDE.md`'s "Known pre-existing Windows failures" note explains the *Windows* failure mode for `mcp-initialize.test.ts` / `mcp-roots.test.ts` (EPERM in afterEach). The macOS/Linux Node ≥25 failure mode is different (timeouts on the parent side, child exits 1 on startup), and is now addressed by this PR — happy to add a follow-up doc tweak to CLAUDE.md if useful.

## Test plan

- [x] `npm test` green on Node 26.0.0
- [x] `node dist/bin/codegraph.js status` still prints the unsupported-version banner and exits 1 on Node 26.0.0 (guard intact for end users)
- [ ] CI green on Node 22/23 (no behavior change expected)

### DIFF
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -5,6 +5,20 @@ export default defineConfig({
     globals: true,
     environment: 'node',
     include: ['__tests__/**/*.test.ts'],
+    /**
+     * Several MCP integration tests (mcp-daemon, mcp-initialize, mcp-ppid-watchdog,
+     * mcp-roots) spawn `dist/bin/codegraph.js serve --mcp` with `process.execPath`
+     * and rely on the child inheriting `process.env`. On a Node >= 25 dev machine
+     * the CLI's hard-block (src/bin/codegraph.ts) would otherwise exit the child
+     * before it ever responds, so every spawn-based test times out — see #478.
+     *
+     * Setting the override here keeps the CLI's runtime guard intact for end
+     * users (it's still enforced when `codegraph` is invoked directly) while
+     * letting the test suite run on whatever Node the contributor happens to
+     * have installed. CI on Node 22/23 is unaffected — the guard doesn't fire
+     * there, so the variable is a no-op.
+     */
+    env: { CODEGRAPH_ALLOW_UNSAFE_NODE: '1' },
     coverage: {
       provider: 'v8',
       reporter: ['text', 'json', 'html'],


colbymchenry (approved):
Nice work — diagnosis is precise (the `process.execPath` spawn pattern + the >=25 hard-block) and the fix is the right pinch point: vitest's `test.env` propagates into every spawned child, both the ones that default-inherit `process.env` and the ones that spread `{...process.env, ...}`, so a single line in the config covers all four files. Runtime guard stays intact for end users (verified `node dist/bin/codegraph.js status` still banners + exits 1 on a simulated >=25 box).

Validated locally end-to-end by temporarily lowering the bin's `nodeMajor >= 25` to `>= 22` on a Node 22 dev box, reproducing the exact `Timed out. Messages so far: []` symptom from #478, then confirming the `env` addition flips all 5 probed tests green. Reverted the probe; no regression with the gate at 25. Thanks for the clean writeup and the n=0 → n=13 numbers!

PR #480, opened 2026-05-27
feat(cli): surface codegraphVersion + lastIndexedAt in `status --json` (#329)

### LINKED ISSUE #329: Add --json flag to codegraph status for CI/scripting

eddieran:
## Summary

Closes #329.

`codegraph status --json` already shipped a rich JSON payload, but two fields the issue called out — a CLI version pin and an actual last-indexed timestamp — were missing, forcing CI consumers to either shell out to a second `codegraph --version` or peek at the SQLite file directly to assert freshness. This PR adds them.

## New JSON fields

| Field | Type | Notes |
|---|---|---|
| `codegraphVersion` | `string` | Read from `package.json`. Present in **both** the initialized and not-initialized branches so CI can verify the running CLI matches what it expects regardless of project state. |
| `lastIndexedAt` | `number \| null` | ms-since-epoch of `MAX(indexed_at)` from the `files` table (single indexed aggregate, no per-file scan). `null` when no files are tracked. |
| `lastIndexedAtIso` | `string \| null` | ISO-8601 mirror of the same instant, for human/log consumers. |

Example (this repo):

```json
{
  "codegraphVersion": "0.9.6",
  "initialized": true,
  "projectPath": "/Users/.../codegraph",
  "fileCount": 192,
  "nodeCount": 2964,
  "edgeCount": 7935,
  "lastIndexedAt": 1779841639913,
  "lastIndexedAtIso": "2026-05-27T00:27:19.913Z",
  ...
}
```

## Library-level API

A matching public method, `CodeGraph.getLastIndexedAt(): number | null`, lets library consumers do the same freshness check without spawning the CLI. It delegates to a new `QueryBuilder.getLastIndexedAt()` that runs one indexed `SELECT MAX(indexed_at) FROM files`. The existing `GraphStats` shape is deliberately untouched — adding a sibling method is cheaper and avoids the breaking-change footprint of a typed-field addition.

## Test plan

New `__tests__/status-json.test.ts` covers four cases:

- `getLastIndexedAt()` returns `null` on a fresh project (no `indexAll` yet).
- `getLastIndexedAt()` returns a timestamp inside the `Date.now()` window the test spent in `indexAll`.
- `codegraph status --json` on an uninitialized project includes `codegraphVersion` (and `initialized: false`).
- `codegraph status --json` on an indexed project includes `codegraphVersion`, `lastIndexedAt` (within a 5-minute window), and a `lastIndexedAtIso` that round-trips back to `lastIndexedAt`.

The CLI-level tests exec the built binary (`dist/bin/codegraph.js`) end-to-end through `process.execPath` so the JSON shape survives future refactors of the underlying plumbing.

Full suite:

```
Test Files  49 passed (49)
     Tests  1036 passed | 2 skipped (1038)
```

(+4 new tests over `main`; no regressions.)

## Notes for review

- The `version` field is named `codegraphVersion` rather than `version` to leave room for a future `schemaVersion` field if/when DB-format pins matter (`schema_versions` already tracks this internally). Happy to rename if you'd prefer just `version`.
- Both `lastIndexedAt` (raw ms) and `lastIndexedAtIso` are emitted because the issue suggested ISO but the rest of the JSON payload uses raw numeric values (`dbSizeBytes`, counts). Easy to drop either if you'd prefer just one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ## [Unreleased]
 
 ### Added
+- **`codegraph status --json` now reports the CLI version and a real last-indexed timestamp for CI/scripting (#329).** Three new fields are surfaced in the JSON output that previously only existed (when at all) in the human-readable form:
+  - `codegraphVersion` — the running CLI's version (matches `package.json`), included in both the initialized and not-initialized branches so CI can pin against a known version without an extra `codegraph --version` call.
+  - `lastIndexedAt` — ms-since-epoch of the most recent file `indexed_at` write, computed via `MAX(indexed_at)` over the `files` table (one indexed query, no per-file scan). `null` when the project has no tracked files.
+  - `lastIndexedAtIso` — the same instant in ISO-8601, for log/human consumers; mirrors `lastIndexedAt` exactly (or `null`).
+
+  Surfaces the same value via a new public API method, `CodeGraph.getLastIndexedAt(): number | null`, for library consumers that want index-freshness checks without shelling out to the CLI. Closes #329.
 - **Java / Kotlin imports now resolve by fully-qualified name.** Extraction
   wraps every top-level declaration of a `.kt` / `.java` file in a `namespace`
   node carrying the file's `package` (so a class `Bar` in
--- a/__tests__/status-json.test.ts
+++ b/__tests__/status-json.test.ts
@@ -1,0 +1,113 @@
+/**
+ * Tests for the fields `codegraph status --json` exposes for CI/scripting
+ * consumers (issue #329) — specifically the `codegraphVersion` field and the
+ * `lastIndexedAt` / `lastIndexedAtIso` pair populated from `MAX(indexed_at)`
+ * over the `files` table.
+ *
+ * The CLI itself is exercised end-to-end so the JSON shape and field names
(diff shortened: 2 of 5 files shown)


colbymchenry:
Thanks, @eddieran — your `getLastIndexedAt()` design and the end-to-end CLI test harness were exactly the right shape, and I reused both. You're credited as a co-author on the squash commit.

The fields landed in #655 (squashed as `7b62356`), combining your `version` + last-indexed work with the `indexPath` field requested in the issue. Two deliberate calls: we went with the issue's field names (`version`, and a single ISO `lastIndexed`) rather than `codegraphVersion` + the ms/ISO pair, to match the original request and keep the JSON contract minimal; and we dropped `agentCount` (no clear consumer). Your `schemaVersion`-headroom point is noted if we ever need it.

Closing as superseded by #655. Thank you for the thorough PR! 🙏

PR #481, opened 2026-05-27
feat(cli): add --max-file-size flag to index/sync/init (#369)

### LINKED ISSUE #369: Feature: Add --max-file-size flag to limit indexing by file size
## Problem
Large binary files or non-text assets can inflate the knowledge graph and waste tokens during indexing.

## Solution
Add a `--max-file-size` flag (default: 1MB) to skip files exceeding the threshold during `codegraph init`. Accept human-readable values: `500kb`, `2mb`, `1gb`.

## Use case
CI pipelines indexing only source code, avoiding media files that would otherwise be included.

eddieran:
## Summary

Closes #369.

CI pipelines that only want source-code symbols (and would rather skip multi-MB media/vendor blobs) currently have to patch the codebase: the 1 MiB threshold lives as a compile-time constant in `src/extraction/index.ts`. This PR makes that threshold configurable per invocation, with the same default so existing workflows are unaffected.

## New surface

**CLI** — three commands grow an identical `--max-file-size <size>` flag:

```bash
codegraph init -i --max-file-size 500kb
codegraph index --max-file-size 2mb
codegraph sync --max-file-size 1.5gb
```

`<size>` accepts:

| Input | Bytes |
|---|---|
| `1048576` | 1 048 576 (raw bytes) |
| `500kb` / `500 KB` | 512 000 |
| `2mb` / `2 MB` | 2 097 152 |
| `1.5gb` | 1 610 612 736 |
| `700KiB` / `1MiB` | IEC suffixes also supported |

Both decimal and IEC suffixes resolve to the binary base (×1024). That matches what `du -h` / `ls -lh` report and keeps the 1 MiB default the codebase has always used. Invalid values fail fast:

```
$ codegraph index --max-file-size banana
✗ Invalid --max-file-size value: "banana". Use a non-negative number with an optional unit suffix (e.g. "500kb", "2mb", "1.5gb", "1048576").
```

**Library** — `IndexOptions.maxFileSize?: number` is added to both `CodeGraph.indexAll(opts)` and `CodeGraph.sync(opts)`. Omitting it keeps the existing 1 MiB default. `DEFAULT_MAX_FILE_SIZE` (the renamed `MAX_FILE_SIZE`) is also exported from `src/extraction/index.ts` for consumers that want to clamp or report against it.

## Implementation notes

- The orchestrator's `indexAll`, `sync`, `indexFile`, and `indexFileWithContent` methods now take an optional `maxFileSize` argument with the default constant as fallback — additive, non-breaking on the existing positional-arg shape.
- `parseFileSize()` lives in `src/utils.ts` (it's small and pure, and may be useful elsewhere in the future). Returns `null` on bad input so the CLI can format a clean error and exit 1 — silently coercing to the default would have been worse than failing fast for a CI use case.
- No persistence: each invocation provides its own value, the watcher/auto-sync keeps using the default. This matched the issue's "CI pipelines indexing only source code" framing more naturally than a per-project config setting; happy to layer in a config-file knob in a follow-up if there's demand.

## Test plan

New `__tests__/max-file-size.test.ts` (8 tests) covers:

- **`parseFileSize`** — plain bytes, kb/mb/gb (with and without spaces, case-insensitive), IEC kib/mib/gib, and explicit `null` for malformed inputs (empty, negative, unknown suffix, repeated dots).
- **`CodeGraph.indexAll({ maxFileSize })`** — tightened limit drops a ~5 KiB file with `code: 'size_exceeded'` while keeping the small one; default keeps both.
- **CLI flag end-to-end** — invalid size string → exit 1 + clear stderr message; `--max-file-size 1kb --quiet` against the built binary drops the larger file from the index while keeping the smaller one.

Full suite green:

```
Test Files  50 passed (50)
     Tests  1040 passed | 2 skipped (1042)
```

(+8 new tests over `main`; no regressions.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New Features
 
+- **`codegraph index --max-file-size <size>` (also on `init` and `sync`) lets CI override the 1 MiB skip threshold (#369).** The previous compile-time `MAX_FILE_SIZE = 1 MiB` constant is now `DEFAULT_MAX_FILE_SIZE` and falls back unchanged when no override is given — so existing workflows behave identically. The flag accepts both raw byte counts (`1048576`) and human-readable sizes with binary multipliers (`500kb`, `2 MB`, `1.5GB`, `700KiB`); both decimal and IEC suffixes resolve to the binary base (×1024) to match what `du`/`ls -lh` report and the 1 MiB default the codebase has always used. Invalid values exit with a clear error (`Invalid --max-file-size value: …`) rather than silently coercing. Library consumers get the same control via `CodeGraph.indexAll({ maxFileSize })` / `sync({ maxFileSize })`. Closes #369.
+
 - `codegraph status --json` now also reports the running CLI `version`, the index directory (`indexPath`), and a `lastIndexed` timestamp (ISO-8601, or null when nothing's indexed yet), so CI and scripts can pin the CLI version and check index freshness from a single command. A matching `CodeGraph.getLastIndexedAt()` library method exposes the same freshness check without shelling out. Thanks @12122J and @eddieran. (#329)
 
 ### Fixes
--- a/__tests__/max-file-size.test.ts
+++ b/__tests__/max-file-size.test.ts
@@ -1,0 +1,162 @@
+/**
+ * Tests for the configurable max-file-size limit (#369).
+ *
+ * Three layers under test:
+ *   1. `parseFileSize` — pure unit conversion of human-readable sizes.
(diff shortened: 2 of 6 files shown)


eddieran:
Rebased this onto current `main` and resolved the two conflicts:

- `src/bin/codegraph.ts`: kept the new upstream behavior where `init` indexes by default and `-i/--index` is only a compatibility no-op, then threaded `--max-file-size` into that default indexing path.
- `CHANGELOG.md`: kept this PR scoped to the max-file-size entry under the current Unreleased section.

Validation after the rebase:

- `npm run build`
- `npx vitest run __tests__/max-file-size.test.ts` (8/8 passed)

Note: local install warned because my machine is on Node 26 while the package declares `<25`, but build/tests completed successfully.

PR #494, opened 2026-05-27
feat(mcp): multi-module Go trace-quality + small-repo retrieval tuning

colbymchenry:
## Summary

Two related streams that landed together on this branch:

### 1. Multi-module Go trace quality

Driven by an 8-question agent-eval audit (cobra, gin, prometheus, cosmos-sdk, etcd). The empirical gate ruled out `go.work` parsing as the real gap (prometheus crushes without it). Actual failure modes + their fixes:

- **Generated-file noise warped disambiguation.** `codegraph_search "Send"` on cosmos-sdk returned the gRPC stub at `tx_grpc.pb.go:124` first; trace landed on the empty stub and the agent fell back to Read. Fix: `src/extraction/generated-detection.ts` — path-pattern classifier for `.pb.go`, `.pulsar.go`, `_grpc.pb.go`, `_mock.go`, `_mocks.go`, `mock_*.go`, `.generated.[jt]sx?`, `_pb2(_grpc)?.py`, `.pb.{cc,h}`, `.g.dart`, `.freezed.dart`. Applied as a stable sort tiebreaker in `findSymbol`, `findAllSymbols`, `codegraph_search` (MCP + CLI), `codegraph_explore` file ranking, and context formatter Entry Points / Related Symbols / Code blocks.
- **Go has no static interface→impl bridge.** Structural typing means the existing `interfaceOverrideEdges` (Java/Kotlin only) doesn't apply. Fix: `goGrpcStubImplEdges` synthesizer in `callback-synthesizer.ts` — detects `UnimplementedXxxServer` structs in generated files, identifies RPC methods (excluding `mustEmbed*` / `testEmbeddedByValue`), emits `calls` edges to matching methods on any non-generated struct whose method-name set is a superset. **467 bridge edges** on cosmos-sdk; bank's `UnimplementedMsgServer::Send` points to `msg_server.go` only — not to `msgClient` siblings or mocks.
- **Trace failure used to fan out into 3-5 follow-up calls.** Fix: inline both endpoints' bodies (capped 120 lines / 3600 chars), their callers (≤6), callees (≤8), AND the other top-level functions/methods in the destination's file in one response. Replaces the `node`→`search`→`node`→Read fan-out.
- **Trace endpoint pairing picked by FTS rank.** On a multi-module repo, `EndBlocker` exists in 20+ modules. Fix: score every `from`×`to` combo by shared directory prefix length (full candidate set, not just FTS top-5), with a less-canonical-path penalty (`enterprise/`, `contrib/`, `examples/`, `vendor/`, `third_party/`, `deprecated/`, `legacy/`) so the canonical-module pair wins. FindPath probe budget capped at 20.
- **Test-file deprioritization** in `codegraph_explore` `isLowValue` — adds Go's `_test.go`, Ruby's `_spec.rb`, JS/TS `.test.ts`/`.spec.tsx`, JVM `*Test.java`/`*Spec.kt`. Without this, etcd's `watchable_store_test.go` consumed 5K chars of explore budget.

### 2. Small-repo retrieval tuning (`<500` files)

The micro-repo tier had its own failure mode: lots of small MCP calls cost more in cache-write tokens than the repo is worth. Three coordinated changes:

- **Tool surface gating.** Project under 500 indexed files exposes only the 5 core tools (`search`/`context`/`node`/`explore`/`trace`). Empirically validated as the floor — 3-tool gate regressed cobra/ky/sinatra, 1-tool gate catastrophically regressed express (+107% LOSS).
- **Sufficiency steering.** `codegraph_context` responses on sub-500 projects end with a strong directive telling the agent the response IS the comprehensive pass — follow-ups should be narrow (`trace from→to`, single-symbol `node`), not another broad `explore`.
- **Tighter budgets.** New sub-150 explore-output tier (13K total / 4 files / 3.8K each, Relationships dropped, test/spec/icon/i18n hard-excluded unless the query is about tests). `maxNodes` defaults to 8 instead of 20 on sub-150 context calls.

### 3. Other improvements that landed alongside

- **Auto-trace inline in `codegraph_context`** when the task looks like "how does X reach Y" — runs the trace internally and splices its body in. Conservative detection (flow keyword + ≥2 PascalCase/camelCase identifiers). Saves the #2 cost-driver follow-up call on multi-module flow questions.
- **Routing manifest inline** for small-repo routing queries — compact URL → handler table built from `route` nodes + their `references`/`calls` edges, plus the top handler file's source. Beats the Glob+Read pattern that was winning on realworld template repos (rails-realworld, laravel-realworld, drupal-admintoolbar).
- **Core-directory ranking boost** — projects with a dominant in-file edge-count file (sinatra's `base.rb` at ~85%) now boost search results in that directory by +25 score, so the core file's siblings outrank sibling-package extensions. Generated/test files excluded from "dominant file" candidacy.
- **`interfaceOverrideEdges` extended beyond JVM** — Java/Kotlin → also C#, TypeScript, JavaScript, Swift, Scala. Swift conformance iterates `struct` nodes too.
- **MCP catch-up gate.** Post-open `cg.sync()` was fire-and-forget; first tool call now awaits it so files deleted/edited while no server was running can't produce stale rows (per-file staleness banner can't help — that signal is watcher-populated). Subsequent calls pay nothing.
- **Shorter MCP tool descriptions.** All 10 `codegraph_*` descriptions condensed (~50% shorter); load-bearing steering stays in `server-instructions.ts`.

## Empirical results

`docs/benchmarks/call-sequence-analysis.md` and the per-arm harness in `scripts/agent-eval/` track the numbers. Headline cosmos-sdk + etcd table (n=2 per question, headless):

| Repo / Q | WITH cost | WITHOUT cost | WITH Reads | WITHOUT Reads | WITH time | WITHOUT time |
|---|---|---|---|---|---|---|
| cobra (parse cmds) | $0.27 | $0.27 | **0** | 4 | **39s** | 60s |
| prometheus (scrape→TSDB) | **$0.63** | $0.70 | **0** | 6 | **106s** | 143s |
| cosmos-sdk Q1 (MsgSend) | $0.41 | $0.26 | **1** | 2 | 67s | 64s |
| cosmos-sdk Q2 (MsgDelegate) | $0.47 | $0.46 | **0** | 5 | **50s** | 73s |
| cosmos-sdk Q3 (gov tally) | $0.34 | $0.31 | **1.5** | 3 | **54s** | 76s |
| etcd Q1 (Put→raft) | **$0.65** | $0.78 | **0** | 4 | **98s** | 129s |
| etcd Q2 (watch) | **$0.36** | $0.50 | **0** | 4+ | **58s** | 89s |

Codegraph wins on **reads and time across every question**. Cost is 3 clean wins, 3 within-10% ties, and 1 stubborn loss on cosmos Q1 (a grep-favored question where the WITHOUT path is structurally short). Cosmos-sdk cost gap collapsed from -60% avg to -15% avg vs baseline; Q3 went from 75% loss to a tie.

## Test plan

- [x] `npm test` — **1081 passed** (50 files), including new
      `__tests__/generated-detection.test.ts` (4 cases pinning the suffix
      contract), `__tests__/mcp-catchup-gate.test.ts` (5 cases for the
      gate behavior + drop-after-first-await), Go gRPC stub-impl synthesis
      cases in `__tests__/frameworks-integration.test.ts`, and the updated
      `__tests__/explore-output-budget.test.ts` covering the new
      `<150` tier
- [x] `npm run build` clean
- [x] cosmos-sdk Q1 r1 + r2 / Q2 / Q3
- [x] etcd Q1 + Q2 (real `go.work` repo, different from cosmos)
- [x] prometheus + cobra control runs (no-regression)
- [x] Bridge edge spot-check on cosmos-sdk: bank's `UnimplementedMsgServer::Send` → `msgServer::Send`, no mock/client false positives


treytracedit-lab:
Fix

PR #498, opened 2026-05-27
fix(windows): suppress console popup on child_process calls

csw-chen:
Add windowsHide: true to all spawnSync/execFileSync/execSync calls across extraction, installer, sync, and mcp modules to prevent console window popup on Windows when CodeGraph spawns git/npm subprocesses.

Affected subprocess calls:
- WASM relaunch (spawnSync)
- git status/rev-parse calls (execFileSync × 4 in extraction)
- git worktree/hooks detection (execFileSync × 3 in sync)
- npm install detection (execSync × 2 in installer)

PR #505, opened 2026-05-27
fix(windows): hide console window for git child processes

### LINKED ISSUE #485: Windows: Daemon process creates flashing black console windows
After upgrading to v0.9.5/v0.9.6 on Windows, the new shared daemon process repeatedly opens and closes blank console windows, which is very disruptive. Setting CODEGRAPH_NO_DAEMON=1 works as a temporary fix. It would be great if the daemon could run completely hidden on Windows by default.

yushengruohui:
Fixes #485.

## Summary

On Windows, the daemon flashes a black `git.exe` console window on every auto-sync. The cause is `execFileSync('git', …)` call sites in the daemon's code paths missing `windowsHide: true`. This PR adds the option to all 7 sites; the change is a one-liner per call site and a no-op on macOS/Linux.

## Root cause

The daemon is launched with `detached: true` (`mcp/index.ts`, the existing `spawn(process.execPath, …)` call), which deliberately leaves it without a parent console. On Windows, when a Console-subsystem child (`git.exe`) is started from a console-less parent, Windows allocates a fresh console window for it unless `CREATE_NO_WINDOW` is set. Node maps `windowsHide: true` to `CREATE_NO_WINDOW`. Without it, every short-lived `git` invocation flashes its own window. The daemon launcher itself already passes `windowsHide: true` — it's the git call sites that were missed.

## Call sites changed (7 total)

| File | Function / line | git command |
|---|---|---|
| `src/extraction/index.ts` | `collectGitFiles` shared `gitOpts` | `ls-files -c --recurse-submodules`, `ls-files -o --exclude-standard` |
| `src/extraction/index.ts` | `getGitVisibleFiles` | `rev-parse --show-toplevel` |
| `src/extraction/index.ts` | `getGitVisibleFiles` | `check-ignore -q` |
| `src/extraction/index.ts` | `getGitChangedFiles` | `status --porcelain --no-renames` |
| `src/sync/worktree.ts` | `gitWorktreeRoot` | `rev-parse --show-toplevel` |
| `src/sync/git-hooks.ts` | `isGitRepo` | `rev-parse --is-inside-work-tree` |
| `src/sync/git-hooks.ts` | `gitHooksDir` | `rev-parse --git-path hooks` |

The fix is the same one-line addition (`windowsHide: true`) on each options object.

## Verification

1. Apply the patch and rebuild (`npm run build`).
2. Restart the daemon: kill the PID in `.codegraph/daemon.pid`, then trigger any CodeGraph CLI/MCP call so it respawns.
3. Edit files in an indexed repo at >1 write/sec. The `git.exe` console window should no longer flash. `.codegraph/daemon.log` still shows the `[CodeGraph MCP] Auto-synced N file(s) in <ms>ms` lines — i.e. the sync still runs, just hidden.
4. macOS/Linux: `windowsHide` is ignored, behavior unchanged.

## Notes for reviewers

- `extraction/wasm-runtime-flags.ts` uses `spawnSync(process.execPath, …, { stdio: 'inherit' })` — the relaunch shim spawns Node from a Node parent that already owns a console, so no new window is allocated and no change is needed.
- Optional follow-up (not in this PR to keep the diff minimal): extract a `runGit(args, opts)` helper that injects `windowsHide: true` so this can't regress.

### DIFF
--- a/src/extraction/index.ts
+++ b/src/extraction/index.ts
@@ -191,7 +191,7 @@ export function buildDefaultIgnore(rootDir: string): Ignore {
  * (See issue #193.)
  */
 function collectGitFiles(repoDir: string, prefix: string, files: Set<string>): void {
-  const gitOpts = { cwd: repoDir, encoding: 'utf-8' as const, timeout: 30000, maxBuffer: 50 * 1024 * 1024, stdio: ['pipe', 'pipe', 'pipe'] as ['pipe', 'pipe', 'pipe'] };
+  const gitOpts = { cwd: repoDir, encoding: 'utf-8' as const, timeout: 30000, maxBuffer: 50 * 1024 * 1024, stdio: ['pipe', 'pipe', 'pipe'] as ['pipe', 'pipe', 'pipe'], windowsHide: true };
 
   // Tracked files. --recurse-submodules pulls in files from active submodules,
   // which the index would otherwise represent only as a commit pointer.
@@ -241,7 +241,7 @@ function getGitVisibleFiles(rootDir: string): Set<string> | null {
     const gitRoot = execFileSync(
       'git',
       ['rev-parse', '--show-toplevel'],
-      { cwd: rootDir, encoding: 'utf-8', timeout: 5000, stdio: ['pipe', 'pipe', 'pipe'] }
+      { cwd: rootDir, encoding: 'utf-8', timeout: 5000, stdio: ['pipe', 'pipe', 'pipe'], windowsHide: true }
     ).trim();
 
     if (path.resolve(gitRoot) !== path.resolve(rootDir)) {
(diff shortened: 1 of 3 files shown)


colbymchenry:
Thanks for the precise diagnosis and the careful per-call-site breakdown — your investigation matched the root cause exactly.

Closing as superseded by #498, which lands the same `windowsHide: true` change at the 7 git `execFileSync` sites you identified plus the 3 other call sites (WASM `spawnSync` relaunch, installer `execSync`, antigravity target) that were missing the flag for the same reason. Merged in cea78ce.

Validated on Windows 11 ARM64 (Parallels): a detached parent's 15 git spawns produce 15 visible black flash-windows without the fix and 0 with it. Fixes #485, #510, #530.

PR #521, opened 2026-05-28
fix(wasm): add windowsHide to spawnSync re-exec to suppress terminal popup on Windows

### LINKED ISSUE #510: Annoying terminal pops up
Whenever I save a file, the terminal repeatedly pops up and interrupts my editing session. This happens on every save and makes it difficult to focus on coding.

JirA44:
The `spawnSync` call in `relaunchWithWasmRuntimeFlagsIfNeeded` was missing `windowsHide: true`.

When the MCP server is launched by a host (e.g. VS Code) it is spawned with `windowsHide: true`, so it has no attached console. When the server then re-execs itself with `--liftoff-only` via `spawnSync` — without `windowsHide: true` — Windows finds no existing console to inherit and opens a new one, producing the visible terminal flash on every save/MCP invocation.

The fix is one line: add `windowsHide: true` to the `spawnSync` options, matching the flag already present on the initial daemon spawn in `src/mcp/index.ts`.

Fixes #510

### DIFF
--- a/src/extraction/wasm-runtime-flags.ts
+++ b/src/extraction/wasm-runtime-flags.ts
@@ -97,6 +97,7 @@ export function relaunchWithWasmRuntimeFlagsIfNeeded(scriptPath: string): void {
   const argv = buildRelaunchArgv(scriptPath, process.argv.slice(2));
   const result = spawnSync(process.execPath, argv, {
     stdio: 'inherit',
+    windowsHide: true,
     env: { ...process.env, [RELAUNCH_GUARD_ENV]: '1', [HOST_PPID_ENV]: String(process.ppid) },
   });
 


colbymchenry:
Thanks for spotting the missing `windowsHide: true` on the WASM `spawnSync` re-exec — your diagnosis matched the root cause exactly.

Closing as superseded by #498, which lands the same one-line change to `relaunchWithWasmRuntimeFlagsIfNeeded` plus 9 other call sites across extraction/sync/installer that were missing the flag for the same reason. Merged in cea78ce.

Validated on Windows 11 ARM64 (Parallels): a detached parent's 15 git spawns produce 15 visible black flash-windows without the fix and 0 with it. Fixes #510 (and #485, #530).

PR #553, opened 2026-05-29
fix(extraction): handle non-ASCII paths in git fast paths

### LINKED ISSUE #541: Files under directories with non-ASCII (CJK) names are silently skipped during indexing

## Summary

## Version

## Minimal reproduction

```sh
REPRO=/tmp/cg-cjk-repro

codegraph init "$REPRO"
```

### Expected

Both files indexed:

```
```

### Actual

```
## Files (1)
```

## Ruling out gitignore

```sh
```

## Impact

thismilktea:
## Summary
  - switch git-backed file discovery to `-z` / NUL-delimited parsing
  - switch git-backed change detection to `git status --porcelain -z`
  - add regression coverage for CJK directory names in both discovery and sync paths

  ## Problem
  Files under non-ASCII directory names could be silently skipped during indexing because git fast paths were parsing
  human-readable text output rather than machine-readable path output.

  When Git returned quoted/escaped non-ASCII paths, CodeGraph could treat that display-format output as a literal file
  path and skip the file before it ever reached extraction.

  ## Fix
  This PR updates the git-backed scan/change-detection paths in `src/extraction/index.ts` to use `-z` machine-readable
  output and parse NUL-delimited records instead of newline-delimited text output.

  That avoids Git quoted/escaped path output from being treated as a literal path, which restores correct handling for
  files under CJK and other non-ASCII directory names.

  ## Test plan
  - `npx vitest run __tests__/extraction.test.ts -t "issue #541"`
  - `npx vitest run __tests__/sync.test.ts -t "CJK directories via git"`
  - `npm run build`

   Close #541

### DIFF
--- a/__tests__/extraction.test.ts
+++ b/__tests__/extraction.test.ts
@@ -3393,6 +3393,37 @@ describe('Directory Exclusion', () => {
     expect(files[0]).toBe('src/components/Button.tsx');
     expect(files[0]).not.toContain('\\');
   });
+
+  it('should index tracked files under CJK directories in git repos (issue #541)', async () => {
+    const { execFileSync } = await import('child_process');
+    const git = (cwd: string, ...args: string[]) =>
+      execFileSync('git', args, { cwd, stdio: 'pipe' });
+
+    const root = path.join(tempDir, 'cjk-root');
+    fs.mkdirSync(path.join(root, 'src', 'english'), { recursive: true });
+    fs.mkdirSync(path.join(root, 'src', '中文目录'), { recursive: true });
+
+    git(root, 'init', '-q');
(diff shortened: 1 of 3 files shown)


Arvin-Hugh:
[Files under directories with non-ASCII (CJK) names are silently skipped during indexing #541](https://github.com/colbymchenry/codegraph/issues/541)

Looking forward to the update!

colbymchenry:
Thanks for this — closing as superseded. Non-ASCII/CJK git paths are now handled on `main` via `git ls-files -z` (NUL-separated, unquoted output — robust for any path bytes), and #541 is closed. Your fix targeted the same issue; it just landed a different way. Appreciate you catching it.

PR #561, opened 2026-05-29
init: exclude local .codegraph indexes from git

Wood-Q:
## Summary

Automatically add newly initialized `.codegraph/` directories to the local Git exclude file (`.git/info/exclude`) during `codegraph init`.

## Problem

Today, every time a user runs `codegraph init`, the generated `.codegraph/` directory can appear in `git status`. In practice, that means users often have to manually add `.codegraph/` to `.gitignore` after initialization.

That is annoying repeated cleanup, and it also creates a product mismatch: `.codegraph/` is local generated index state, not source code. Repositories published to GitHub should not need to carry CodeGraph-specific ignore rules just because one developer initialized an index locally. Other users can simply run `codegraph init` themselves.

## Solution

This PR writes the ignore rule to `.git/info/exclude` instead of the tracked `.gitignore`.

That gives users the desired behavior locally:

- `.codegraph/` no longer appears in `git status` after init
- project `.gitignore` remains untouched
- GitHub/released source trees do not gain CodeGraph-specific metadata
- repeated init flows do not duplicate existing exclude entries
- nested project roots in monorepos write the correct relative exclude path

If the project is not inside a Git worktree, or Git metadata cannot be accessed, initialization still succeeds and the exclude step is skipped.

## Tests

- Added coverage for adding `.codegraph/` to `.git/info/exclude`
- Added idempotency coverage for existing exclude entries
- Added nested project coverage for monorepo-style paths
- Added non-Git initialization coverage
- Verified with:
  - `npm run build`
  - `npx vitest run __tests__/foundation.test.ts`
  - `git diff --check`

### DIFF
--- a/__tests__/foundation.test.ts
+++ b/__tests__/foundation.test.ts
@@ -8,6 +8,7 @@ import { describe, it, expect, beforeEach, afterEach } from 'vitest';
 import * as fs from 'fs';
 import * as path from 'path';
 import * as os from 'os';
+import { execFileSync } from 'child_process';
 import { CodeGraph } from '../src';
 import { Node, Edge } from '../src/types';
 import { isInitialized, getCodeGraphDir, validateDirectory } from '../src/directory';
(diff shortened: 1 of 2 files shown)


Wood-Q:
@colbymchenry Can you have a check? Thank you!

PR #581, opened 2026-05-30
fix(extraction): visit anonymous function bodies

### LINKED ISSUE #528: Anonymous function_expression bodies not visited - AMD/CommonJS modules lose all inner symbols
## Bug Description

When a JavaScript file uses the AMD `define()` pattern (common in NetSuite SuiteScript, RequireJS, and other AMD-based projects), all named function declarations inside the anonymous callback are lost during extraction.

**Example:**
```javascript
define(['N/record', 'N/search'], function(record, search) {
    function getInputData() { /* ... */ }
    function map(context) { /* ... */ }
    function reduce(context) { /* ... */ }
    return { getInputData, map, reduce };
});
```

**Expected:** `getInputData`, `map`, `reduce` extracted as function nodes.
**Actual:** Only the file node is created (`node_count: 1`). All inner functions are invisible to the graph.

## Root Cause

In `src/extraction/tree-sitter.ts`, `extractFunction()` returns early when `name === '<anonymous>'` without visiting the function body:

```typescript
if (name === '<anonymous>')
    return; // Skip anonymous functions
```

Since `visitNode()` already set `skipChildren = true` before calling `extractFunction()`, the anonymous function's body (containing named `function_declaration` nodes) is never traversed.

The `isMisparsedFunction` branch (a few lines below) correctly handles a similar case — it skips the node but still visits the body. The anonymous case should do the same.

## Suggested Fix

Replace:
```typescript
if (name === '<anonymous>')
    return; // Skip anonymous functions
```

With:
```typescript
if (name === '<anonymous>') {
    const body = this.extractor.resolveBody?.(node, this.extractor.bodyField)
        ?? getChildByField(node, this.extractor.bodyField);
    if (body) {
        this.visitFunctionBody(body, '');
    }
    return;
}
```

This way the anonymous function itself doesn't become a node, but `visitFunctionBody` will extract any named nested functions (line ~1646-1651 already handles this case).

## Impact

Tested on a 107-file SuiteScript project (all AMD `define()` modules):

| Metric | Before | After |
|--------|--------|-------|
| function nodes | 39 | **670** |
| total nodes | 146 | **777** |
| edges | 177 | **2573** |

The 39 functions that were previously extracted came from files that happened to use top-level `function` declarations without `define()`.

## Environment

- codegraph v0.9.6
- Windows 11, Node.js v24.12.0
- All files are `.js` with AMD `define([], function() { ... })` pattern

msnandhis:
## Summary

Fixes #528.

- Keep traversing anonymous JavaScript/TypeScript function bodies while still suppressing anonymous function nodes.
- Add regression coverage for AMD/CommonJS anonymous callbacks so inner handlers and calls remain in the graph.
- Add an Unreleased changelog fix note.

## Root Cause

`extractFunction()` returned immediately for anonymous `function_expression` / arrow functions. That avoided noisy `<anonymous>` symbols, but it also skipped the callback body. For AMD/CommonJS loader callbacks, that meant named functions and calls inside the callback never reached the graph.

## Testing

- `npx -y -p node@22 -p npm@10 npm exec -- vitest run __tests__/extraction.test.ts -t 'anonymous AMD'`
- `npx -y -p node@22 -p npm@10 npm exec -- vitest run __tests__/extraction.test.ts`
- `npx -y -p node@22 -p npm@10 npm run build`
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixes
 
+- JavaScript and TypeScript callbacks passed anonymously to AMD or CommonJS loaders now keep their inner functions and calls in the graph, without adding noisy anonymous function symbols. (#528)
 - Indexing a project that contains only config-style files (YAML, Twig, or `.properties`) no longer misleadingly reports "No files found to index" — these files are tracked at the file level and are now counted as indexed. Thanks @luojiyin1987 (#357).
 
 ## [0.9.7] - 2026-05-28
--- a/__tests__/extraction.test.ts
+++ b/__tests__/extraction.test.ts
@@ -350,6 +350,51 @@ export const fetchData = async () => {
       isExported: true,
     });
   });
+
+  it('should traverse anonymous AMD and CommonJS callback bodies without anonymous nodes', () => {
+    const code = `
+define(['dep'], function(dep) {
+  function amdHandler() {
+    return dep.load();
+  }
+
+  const amdCallback = () => dep.ready();
+  register(amdHandler);
+});
+
+require(['legacy'], function(legacy) {
+  function commonHandler() {
+    return normalize(legacy.read());
+  }
+
+  var commonCallback = function() {
+    return legacy.ready();
+  };
+  boot(commonHandler);
+});
+`;
+    const result = extractFromSource('legacy-module.js', code);
+
+    const functionNames = result.nodes
+      .filter((n) => n.kind === 'function')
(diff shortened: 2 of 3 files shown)


msnandhis:
Updated in a380a6d to address the automated review feedback: visitFunctionBody now delegates nested function nodes to extractFunction, so true anonymous callbacks still have their bodies traversed without creating <anonymous> nodes, while variable-assigned arrow/function-expression callbacks can use the existing parent variable_declarator name inference. I also expanded the AMD/CommonJS regression test to cover const arrow and var function-expression callbacks.

PR #589, opened 2026-05-31
fix: guide models to use short keywords for codegraph_context task param

### LINKED ISSUE #571: codegraph_context task parameter description misleads models into passing natural language instead of keywords
## Problem

The `task` parameter of `codegraph_context` is described as:

> "Description of the task, bug, or feature to build context for"

This wording leads models (tested with GPT-5.5, likely affects others too) to pass verbose natural language sentences, e.g.:

"Find and explain the call chain for the ${keyword} API endpoint in this Java project. Include controller endpoint, service methods..."

Since the underlying search is FTS5-based, this dilutes the signal with noise words and causes misses on queries that would otherwise succeed with a short keyword like `"bondIncome/workflow/callback"`.

## Impact

Models that follow the current description will get worse results than models that happen to pass short keywords, making codegraph_context unreliable across different model backends.

wwhsaber:
## Problem

The `task` parameter description of `codegraph_context` currently reads:

> "Description of the task, bug, or feature to build context for"

This leads models to pass verbose natural language sentences, diluting FTS5 search results with noise words.

## Fix

Updated the description to guide models toward short keywords with a concrete explanation:

> `Short keywords or symbol paths to search for (e.g. "AuthService/login" or "payment webhook handler"). Use concise terms — the search is FTS5-based, so verbose sentences dilute results with noise words.`

Fixes #571

### DIFF
--- a/src/mcp/tools.ts
+++ b/src/mcp/tools.ts
@@ -441,7 +441,7 @@ export const tools: ToolDefinition[] = [
       properties: {
         task: {
           type: 'string',
-          description: 'Description of the task, bug, or feature to build context for',
+          description: 'Short keywords or symbol paths to search for (e.g. "AuthService/login" or "payment webhook handler"). Use concise terms — the search is FTS5-based, so verbose sentences dilute results with noise words.',
         },
         maxNodes: {
           type: 'number',


wwhsaber:
Closing this PR as it only modifies description text. Will submit more substantial contributions instead.

Issue #556, opened 2026-05-29
SAP xsjs and xsjslib files from SAP HANA 1.0 are js files but ignored by codegraph


radu103:
Can  you please add support for processing indexing ?

It's just a matter of file extensions, javascript language is the same
https://help.sap.com/docs/SAP_HANA_PLATFORM/d89d4595fae647eabc14002c0340a999/b907648a90cd49caabb30dc2d5a7de05.html?locale=en-US

radu103:
I had also an open PR 1205 : https://github.com/colbymchenry/codegraph/pull/1205

maxmilian:
Adding an observation in case it's useful — the extraction half of this landed, but the import-resolution half looks like it's still open.

**Extraction** — `.xsjs` / `.xsjslib` map to `javascript` in `src/extraction/grammars.ts:72-73`, and symbols inside those files index fine.

**Resolution** — the extension list that import paths get resolved against doesn't include them. On `main` (`c6aaa20`), `src/resolution/import-resolver.ts:29`:

```ts
javascript: ['.js', '.jsx', '.mjs', '.cjs', '/index.js', '/index.jsx'],
```

(`xsjs` appears 0 times in that file.)

The effect isn't a missing edge but a wrong one: an extensionless import resolves to nothing, the call falls back to same-name matching, and it can bind to an unrelated file. Repro on `main` (`c6aaa20`) with three files in one directory:

- `helpers.xsjslib` — `export function buildQuery(table) { return "SELECT * FROM " + table; }`
- `decoy.js` — `export function buildQuery(table) { return "DECOY " + table; }` (never imported)
- `service.xsjs` — `import { buildQuery } from "./helpers"; function run() { return buildQuery("users"); }`

| target file | `getFileDependencies('service.xsjs')` | `run()`'s call edge points at |
|---|---|---|
| `helpers.xsjslib` | `['decoy.js']` | `decoy.js` ❌ |
| same test, target renamed to `helpers.js` | `['helpers.js']` | `helpers.js` ✅ |

Only the target's extension differs between the two rows.

If that half is still in scope, #594 is a ready patch for it — the two extensions added to that one list, plus the case above as a test. Happy either way; just flagging the gap.

PR #594, opened 2026-05-31
fix(resolution): resolve ES imports targeting .xsjs/.xsjslib files (#556)

maxmilian:
## What changed

The headline of #556 — indexing SAP HANA `.xsjs` / `.xsjslib` files as JavaScript — **already shipped on `main`** via #654 (the extension map + detection + extraction tests, credited to #556). This PR has been **rebased and slimmed to the one piece #654 did not cover**: import resolution.

## Remaining gap

`EXTENSION_RESOLUTION.javascript` in `src/resolution/import-resolver.ts` did not list `.xsjs` / `.xsjslib`, so an extensionless ES import from an xsjs file —

```js
// service.xsjs
import { buildQuery } from "./helpers";   // -> helpers.xsjslib
```

resolved to nothing. The cross-file call edge was dropped, so `codegraph_callers` / `codegraph_impact` reported the imported symbol as having no callers even though the import is live.

## Fix

Add `.xsjs`, `.xsjslib` to the `javascript` entry in `EXTENSION_RESOLUTION` (one line). Now `./helpers` resolves to `helpers.xsjslib` and the call edge is created.

## Tests

- Kept one resolution test: `.xsjs` -> `.xsjslib` ES import produces a cross-file call edge (`run()` -> `buildQuery`). Passes.
- Dropped the two detection/extraction tests this PR originally added — they overlap with the coverage #654 already merged (`detectLanguage('service.xsjs')`, `extractFromSource('service.xsjs', …)`).

## Scope

Resolution only. The grammar/extension map is untouched here since #654 owns it now.


maxmilian:
Rebased onto current `main`.

To clarify why this is still worth landing even though #556 was closed by #654: #654 registered `.xsjs`/`.xsjslib` as JavaScript on the **extraction** side (`grammars.ts`), so those files are now parsed and indexed — but the **import-resolution** side was untouched. `EXTENSION_RESOLUTION.javascript` in `import-resolver.ts` still doesn't list `.xsjs`/`.xsjslib`, so an extensionless `import { x } from './helpers'` that targets `helpers.xsjslib` resolves to nothing and the cross-file call edge is dropped.

This 1-line addition closes that residual gap; the resolution test `should resolve an ES import from a .xsjs file to a .xsjslib file (#556)` covers it and coexists with the extraction-level `.xsjs` test from #654.

maxmilian:
@colbymchenry heads-up before this goes stale: #556 is closed as completed, but the **import-resolution** half is still missing on `main`. `EXTENSION_RESOLUTION.javascript` in `src/resolution/import-resolver.ts` doesn't include `.xsjs`/`.xsjslib`, so imports targeting those files still don't resolve — the extraction side shipped in #654, but resolution didn't. This PR is the one-line gap-closer. Want me to keep it open / rebase, or is the unresolved import acceptable? I'll close it if you'd rather not carry it.

maxmilian:
Rebased onto the latest `main` — conflict-free. The conflict was textual (upstream added the `arkts` entry directly above the `javascript` one in `EXTENSION_RESOLUTION`); both sides are kept.

While re-verifying after the rebase I found my own test was worthless, and fixing it turned up that this bug is worse than the PR originally described.

**The old test passed with the fix reverted.** With only `helpers.xsjslib` on disk, the import fails to resolve, the call then falls back to same-name matching, and that fallback happens to pick the right file — so the edge exists either way and the test stayed green. It asserted the symptom, not the cause.

**What actually happens without the fix:** the import doesn't resolve to *nothing* — it resolves to the *wrong file*. Dropping a `decoy.js` that exports the same symbol name and is never imported, on current `main`:

```
DEPS=["decoy.js"]   EDGE_TO_REAL=false   EDGE_TO_DECOY=true
```

`import { buildQuery } from "./helpers"` inside `service.xsjs` binds `run() -> buildQuery` to the unrelated `decoy.js`. With `.xsjs` / `.xsjslib` in the resolution list it binds to `helpers.xsjslib` and `DEPS=["helpers.xsjslib"]`. So a `.xsjs` codebase doesn't just lose edges, it gets **fabricated** ones — which is worse for `codegraph_callers` / `codegraph_impact` than a missing edge, since a wrong caller reads as real.

The test now includes that decoy, so it fails on a regression (`expected [ 'decoy.js' ] to deeply equal [ 'helpers.xsjslib' ]`) rather than passing regardless. I mutation-checked it both ways. I also corrected the CHANGELOG line, which claimed the import "resolved to nothing".

One note on scope: #556 is closed as completed by #654, which added the `.xsjs` / `.xsjslib` → JavaScript *extraction* mapping. That half is in; the *resolution* half (this one line) never landed, which is why the mis-binding above still reproduces on `main` today. Happy to close this if you'd rather fold the one-liner in elsewhere.

(Full `extraction.test.ts` run crashes a vitest worker locally, but it does so identically on an unmodified `main` checkout, so that's pre-existing and not from this branch.)

PR #595, opened 2026-05-31
feat(mybatis): index <resultMap> and link statements that use it (#592)

### LINKED ISSUE #592: Should we also include the resultMap and SQL from the MyBatis XML in the index
Title

Should we also include the resultMap and SQL from the MyBatis XML in the index？

Problem
CodeGraph 0.9.6 can index MyBatis XML statements，such as <select>, <update>. However, there are no indexes for `resultMap` and `sql`. In some scenarios, such as when you need to retrieve which methods a specific field in an XML file is associated with, this query will fail.

maxmilian:
## Summary

Fixes #592. The MyBatis XML extractor indexes `<select|insert|update|delete|sql>` but not `<resultMap>`, so "which statements use this result map" queries return nothing.

(Note: `<sql>` fragments raised in the original report **are** already indexed on current `main` — this PR adds the remaining `<resultMap>` piece.)

## Fix

**`src/extraction/mybatis-extractor.ts`**
- Add `resultMap` to the element scan, so each `<resultMap id="...">` becomes a node — the same method-shaped model already used for statements and `<sql>` fragments, qualified `<namespace>::<id>`, with a `contains` edge from the file. No new node kind.
- Emit a reference from any statement carrying `resultMap="..."` to that result-map node, mirroring the existing `<include refid="...">` → `<sql>` link.

**`src/resolution/callback-synthesizer.ts`**
- Guard the Java↔XML bridge. It suffix-matches xml method nodes by `<Class>::<id>`; without a filter, a Java method whose name collides with a `<sql>`/`<resultMap>` id would wrongly bridge to it. Skip those non-statement nodes (identified by signature) — only real statements correspond to Java mapper methods. (This also closes the same latent over-bridge for `<sql>` fragments.)

This lets `callers` / `impact` / `trace` answer "which statements use this result map".

## Tests (`__tests__/frameworks-integration.test.ts`)

- A `<resultMap id="userResult">` is indexed (qualified `…UserMapper::userResult`, signature `<resultMap> type=User`).
- A `<select resultMap="userResult">` produces an edge to that result map.
- A Java method deliberately named `userResult` does **not** bridge to the result map (guard verified).

Full suite green locally (`npm test`).

## Scope

Result-map-level linkage only. Field-level mapping (`<result column=… property=…>` / `<id>` entries as individual nodes — the "specific field → methods" association in the issue) is **not** modeled here; that needs a column/property node kind and is a larger design decision. Happy to follow up if you want that granularity. `<resultMap extends="…">` inheritance is likewise left for a follow-up.


maxmilian:
Closing — #592 was resolved as **not-planned**, so the `<resultMap>` / SQL indexing this PR adds no longer has a home in scope. Happy to revisit if the MyBatis surface comes back on the roadmap.

PR #597, opened 2026-05-31
fix(extraction): index C functions decorated with a leading export macro (#515)

### LINKED ISSUE #515: The interface cannot be indexed; the function is decorated with the macro AX_VIN_GLB_API
CodeGraph did not directly index the AX_VIN_Init symbol (likely because it is decorated with the AX_VIN_GLB_API macro)

maxmilian:
## Summary

Fixes #515. A C function written as `EXPORT_MACRO ReturnType Name(...)` — a leading export/visibility macro in front of a project-defined (typedef'd) return type — was indexed under the garbage name `(params)` instead of its real name, so it couldn't be searched or navigated to. This pattern is ubiquitous in vendor SDKs, e.g. the Axera SoC SDK from the report: `AX_VIN_GLB_API AX_S32 AX_VIN_Init(AX_VOID)`.

## Root cause

Both the macro and the typedef'd return type are unknown identifiers in type position, so tree-sitter-c peels `MACRO RET` off as a separate (broken) `declaration` and parses the real function with its **name absorbed as the `type` field** and the parameter list wrapped in a bare `parenthesized_declarator` instead of a `function_declarator`:

```
function_definition
  type: type_identifier "AX_VIN_Init"      <- the real name
  declarator: parenthesized_declarator     <- the params
    identifier "AX_VOID"
  body: compound_statement
```

`extractName` then falls through to `getNodeText(parenthesized_declarator)` → `"(AX_VOID)"`.

## Fix

`recoverMacroDecoratedFunctionName` (wired only into the C extractor's `resolveName`): when a `function_definition`'s declarator field is directly a `parenthesized_declarator`, take the name from the `type` field instead. That shape is one valid C never produces.

## Tests

`__tests__/extraction.test.ts` → new `C/C++ Extraction` block:
- recovers `AX_VIN_Init` from the macro-decorated definition (and is **not** named `(AX_VOID)`)
- plain `int compute(int a)` is unaffected
- `int (foo)(void)` (redundant parens — keeps a `function_declarator`) does **not** misfire and is **not** renamed to `int`

Full suite green locally: `npm run build` clean, `1097 passed | 2 skipped` (+3 new).

## Scope

Deliberately narrow — verified against the real grammars with parse dumps before/after:

- **Macro + primitive return** (`AX_API int F(...)`), **macro + pointer return** (`AX_API char* F(...)`), and **all C++ forms** already parse to a `function_declarator` and extract correctly on `main` — they don't reach this recovery and aren't touched (no `resolveName` change for the C++ extractor).
- **Bare prototypes in headers** (`AX_API AX_S32 AX_VIN_Init(void);`) are reparsed by tree-sitter as a top-level call expression. Recovering a symbol from that would risk synthesizing symbols from genuine call sites, so it's left as a follow-up. The definition site (this fix) is where the symbol lives, which makes it searchable.

No node-count change — `resolveName` only renames; it cannot add or drop nodes.


maxmilian:
@colbymchenry friendly ping — #515 is still open and this one's self-contained: it recovers C functions whose signature is preceded by a leading export macro (e.g. `EXPORT void foo(void)`), which the extraction pass currently skips. No overlap with the recent extraction merges (#956/#980 are about file/repo discovery, not the C symbol scanner). Happy to rebase if it's drifted — any feedback welcome!

maxmilian:
Thanks for #1102 — the universal post-parse salvage is a much better answer than a per-library list, and it subsumes most of what I was chasing here. I re-checked this PR against it before deciding whether to close.

**It doesn't quite close this one.** `recoverMangledCppName` keys on an *internal space* in an already-mangled name (`"WTFString computeThing"`). The C misparse this PR targets doesn't produce a space — tree-sitter-c peels `MACRO RET` off as a broken `declaration`, leaves the real name in the `type` field, and wraps the params in a bare `parenthesized_declarator`, so `extractName` yields **`"(AX_VOID)"`**. That hits the `if (!/\s/.test(name)) return name;` guard on the first line and returns unchanged.

Reproduced on `edb9f2f` (current `main`), via `extractFromSource`:

```c
typedef int AX_S32;
typedef void AX_VOID;
#define AX_VIN_GLB_API __attribute__((visibility("default")))

AX_VIN_GLB_API AX_S32 AX_VIN_Init(AX_VOID) { return 0; }
```

| input | extracted name |
|---|---|
| `ax.c` | **`(AX_VOID)`** ❌ |
| same source as `ax.cpp` | `AX_VIN_Init` ✅ |
| `ax.c`, primitive return (`AX_API int compute(int)`) | `compute` ✅ |
| `recoverMangledCppName("(AX_VOID)")` | `"(AX_VOID)"` (no-op) |

So the residual hole is narrow and specific: **C grammar + macro + a typedef'd (non-primitive) return type** — exactly the vendor-SDK shape in #515 (still open).

This PR's `recoverMacroDecoratedFunctionName` handles it structurally (declarator field is directly a `parenthesized_declarator` — a shape valid C never produces) rather than by string-shape, so it composes with the salvage rather than competing with it.

Happy to rebase onto the post-#1102 `c-cpp.ts`, and equally happy to refold it as a guard *inside* the `recoverMangledName` hook if you'd rather keep one entry point. Just say which and I'll push.

maxmilian:
Closing this — `b6a05d1` (#1311, merged Jul 16) covers the case this PR was written for.

I checked rather than assuming, by running this PR's own test cases against plain `upstream/main`:

| input | `main` alone | `main` + this PR |
|---|---|---|
| `AX_VIN_GLB_API AX_S32 AX_VIN_Init(AX_VOID)` — the exact case from #515 | ✅ `AX_VIN_Init` | ✅ |
| `AX_EXPORT(1) AX_S32 …` | ✅ | ✅ |
| `API AX_S32 …` (3-char macro) | ✅ | ✅ |
| macro on its own line | ✅ | ✅ |
| `Ax_Api AX_S32 …` (mixed-case macro) | ❌ `(AX_VOID)` | ✅ `AX_VIN_Init` |

`blankCLeadingAttrMacros` blanks leading ALL-CAPS macros before parsing, which is the same shape this PR was targeting. So four of the five rows already pass on `main`, and the regression test in this PR is vacuous there.

The one row that still fails is a mixed-case attribute macro — not the shape #515 reported, and not worth keeping a 73-day PR open for. If it ever comes up in a real header, it's a small follow-up on top of `blankCLeadingAttrMacros` rather than this diff.

One thing worth doing separately: **#515 looks fixed but is still open.** The table above is against current `main`, so it can be closed on that evidence whenever convenient.

PR #598, opened 2026-05-31
fix: avoid sibling methods in impact traversal

### LINKED ISSUE #536: Siblings shouldn't be included in impact by default
### Issue
Impact search does not consider edge type when traversing, meaning that any parent containers and their children are considered impacted.

### Impact
I have a C# function that does not do anything. It has no callers and no callees.
When an LLM searches for impact, it is returning 220 symbols due to other functions that happen to exist in the same C# class
``` 
codegraph_impact
{
  "symbol": "DoNothingFunction",
  "depth": 2
}
```

This happens consistently in C# classes and I assume other languages, making "impact" queries detrimental to LLMs unless your codebase has tight classes.

#### Suggestion
When traversing upstream from a 'contains' edge, the downstream traversal of retrieving all container children should be skipped.

In most languages a node will not have an impact on sibling functions unless there is an explicit link.

If there are languages where this sibling impact is needed, maybe this can be added as a flag.

#### Scenario
```
    public class TestClass {
        public static void NoReferences()
        {
            return;
        }
        public static void SisterFunction()
        {
            return;
        }
    }
```
` codegraph impact NoReferences`

#### Return
```
ImpactTest.cs
  method      NoReferences:2
  class       ImpactTest:1
  method      SisterFunction:6
  file        ImpactTest.cs:1
```

sjh9714:
## Summary

Fixes #536.

`codegraph impact` could walk from a method to its parent class through an incoming `contains` edge and then fan back out to every child of that class, so unrelated sibling methods appeared as impacted. This keeps the parent in the impact result but suppresses the sibling fan-out for containers reached through child-to-parent `contains` traversal.

## Changes

- Track when impact traversal reaches a container through an incoming `contains` edge and skip expanding that container's children in that case.
- Add a C# regression fixture covering a method with an unrelated sibling method.
- Add an Unreleased changelog entry for the impact behavior fix.

## Testing

- `npx vitest run __tests__/graph.test.ts -t "should not include sibling methods reached through parent contains edges"` (failed before the fix, passed after)
- `npx vitest run __tests__/graph.test.ts`
- `npm run build`
- `node dist/bin/codegraph.js impact NoReferences --path "$tmp" --depth 2 --json` after creating the C# fixture used for reproduction
- `npx vitest run __tests__/mcp-roots.test.ts -t "honors an explicit rootUri without asking the client for roots"` (rerun after one full-suite timeout)
- `npm test` (passed on rerun; first run had one transient `mcp-roots` timeout)

Note: I used Codex while preparing this change, reviewed the final diff, and ran the listed checks locally.


treytracedit-lab (approved):

treytracedit-lab:
Git config/CD

sjh9714:
Closing this since the same issue has already been fixed upstream in #654 and #536 is closed. After syncing, this branch no longer has a useful remaining diff against the current main. Thanks for the review context.

PR #599, opened 2026-05-31
[codex] Fix Claude setup and macOS path issues

### LINKED ISSUE #305: docs: add Contributing section to README (local dev setup, FTS5 workaround)
## Problem

There are no contributing or local development instructions in the project. The README only covers end-user installation. A first-time contributor has no guidance on:

- How to fork and set up the project locally
- How to wire the `codegraph` CLI command to their local build instead of the published package
- How to run the test suite
- The `no such module: fts5` error that blocks indexing when running from source on Node 22/23 macOS

## Proposed solution

Add a **Contributing** section to the README covering:

1. Fork + clone + add upstream remote
2. `npm install` + `npm run build`
3. `npm link` to point the global `codegraph` command at your local build (and how to restore the published version)
4. Development workflow (`npm run dev`, `npm test`, `npm run test:watch`)
5. FTS5 workaround — Node 22 and 23 on macOS ship without FTS5; switch to Node 24 via nvm
6. A quick verification snippet to confirm FTS5 is available before indexing
7. PR checklist notes (don't bump `package.json` versions, don't update `package-lock.json` unless you changed dependencies)

## Context

- FTS5 is missing from the official Node 22.x and 23.x macOS binaries. Node 24+ includes it. The self-contained bundled release (curl install / `npm i -g`) ships its own Node binary that has FTS5, so end users are unaffected — this only hits contributors running from source.
- `npm run cli` is an alternative to `npm link` for running the local build without touching the global symlink.

### LINKED ISSUE #486: change config or setting with out backup
Agent  │                       change files                                    │ backup？ │
  ├────────┼────────────────────────────────────────────────┼────────┤
  │ hermes │ ~/.hermes/config.yaml                                    │ no    │
  ├────────┼────────────────────────────────────────────────┼────────┤
  │ claude │ ~/.claude/settings.json + CLAUDE.md           │ no     │
  ├────────┼────────────────────────────────────────────────┼────────┤
  │ cursor │ .cursor/mcp.json + .cursor/rules/codegraph.mdc │ no   │
  ├────────┼──────────────────────────────────────────────   ──┼────────┤
  │ codex  │    ~/.codex/config.toml + AGENTS.md                     │ no    │

### LINKED ISSUE #512: About HarmonyOS and ArkTS
Description
Currently, CodeGraph is incredibly powerful for web and mobile development, but it skips HarmonyOS ArkTS (`.ets`) files during indexing because it recognizes them as unknown extensions. 

ArkTS is a flagship language for Huawei's HarmonyOS ecosystem. It is a superset of TypeScript, sharing almost 90% of TS's core syntax, but it introduces declarative UI structures (like `@Component`, `@Entry`, `build()`, etc.).

Current Workaround & Limitation
Forcing a custom extension mapping via `$env:CODEGRAPH_CUSTOM_EXTENSIONS=".ets=typescript"` partially works, but because standard Tree-sitter parsers for TypeScript don't completely understand the declarative UI blocks of ArkTS, the symbol resolution can sometimes be incomplete.
 Proposal
It would be amazing if CodeGraph could natively support `.ets` files by:
1. Treating `.ets` files as a first-class citizen alongside TypeScript by default.
2. Gracefully handling or adding a Tree-sitter grammar layer that accommodates ArkTS UI components.

This would be a massive game-changer for developers using AI agents (like OpenCode, Claude Code, or Cursor) to build HarmonyOS / OpenHarmony mobile apps! Thank you for this amazing tool!

Official ArkTS Language Specification:     https://developer.huawei.com/consumer/cn/doc/harmonyos-guides/arkts-overview

### LINKED ISSUE #527: Bug: validatePathWithinRoot allows reading files outside project root via symlinks
### Reproduction

```bash
#!/bin/bash
# End-to-end: symlink inside project → file outside root → leaked via MCP
TMPDIR=$(mktemp -d)
trap "rm -rf $TMPDIR" EXIT

mkdir -p "$TMPDIR/project/src" "$TMPDIR/secrets"

# Sensitive file OUTSIDE the project root
echo 'DATABASE_URL="postgres://admin:password@prod-db:5432"' > "$TMPDIR/secrets/config.ts"

# Symlink INSIDE the project pointing outside
ln -s "$TMPDIR/secrets/config.ts" "$TMPDIR/project/src/database.config.ts"

# A legit source file so indexing has something to scan
echo 'export function hello() { return "world"; }' > "$TMPDIR/project/src/index.ts"

cd "$TMPDIR/project"
codegraph init -i    # indexes the project (follows symlink, stores "src/database.config.ts")

# Later, an MCP agent calls codegraph_explore or codegraph_node with includeCode=true.
# validatePathWithinRoot resolves "src/database.config.ts" to an in-root path,
# but readFileSync follows the symlink to $TMPDIR/secrets/config.ts.
```

Run with: `bash poc.sh`
Expected: Symlinks pointing outside the project root should be rejected — their content should not be readable through MCP tools.
Actual: The symlinked file is indexed and its contents are served to MCP agents, leaking files outside the project root.

### Environment
- Version: 02935d7 (latest main as of 2025-05-28)
- Node: v25.9.0
- OS: macOS (arm64)

### Root Cause

`validatePathWithinRoot` (`src/utils.ts:57`) uses only `path.resolve` to check containment, which resolves `..` lexically but does **not** resolve symlinks. A symlink inside the project pointing to a file outside the root passes the check because the *logical* path appears within the root.

The codebase already has `isPathWithinRootReal` (`src/utils.ts:97`) that properly calls `fs.realpathSync` to catch this exact class of attack — but it is never used in the file-read paths of the MCP tools or extraction pipeline.

Two places where this matters:
1. **MCP tool file reads** (`src/mcp/tools.ts:1403, 1432, 1743`): `readFileSync` follows the symlink after `validatePathWithinRoot` approves the logical path.
2. **Extraction/indexing** (`src/extraction/index.ts`): The `walk()` function *does* resolve symlinks with `realpathSync`, but stores the **relative** path (`path.relative(rootDir, fullPath)`) using the symlink's name, not the real target. Later reads of stored paths follow the symlink.

### Suggested Fix

Replace `validatePathWithinRoot` calls in file-read paths with `isPathWithinRootReal` (or add `fs.realpathSync` to `validatePathWithinRoot`):

```typescript
// src/utils.ts — add realpath check to validatePathWithinRoot
export function validatePathWithinRoot(projectRoot: string, filePath: string): string | null {
  const resolved = path.resolve(projectRoot, filePath);
  const normalizedRoot = path.resolve(projectRoot);

  if (!resolved.startsWith(normalizedRoot + path.sep) && resolved !== normalizedRoot) {
    return null;
  }

  // Resolve symlinks to prevent escapes
  try {
    const realPath = fs.realpathSync(resolved);
    const realRoot = fs.realpathSync(normalizedRoot);
    if (!realPath.startsWith(realRoot + path.sep) && realPath !== realRoot) {
      return null;
    }
    return realPath;
  } catch {
    // File doesn't exist yet (e.g., during write) — logical check already passed
    return resolved;
  }
}
```

This preserves backward compatibility while closing the symlink escape. The repo's existing `isPathWithinRootReal` shows the maintainers are aware of this pattern — it just needs to be applied consistently.

### LINKED ISSUE #541: Files under directories with non-ASCII (CJK) names are silently skipped during indexing

### LINKED ISSUE #565: Bug: Claude installer auto-allow omits `codegraph_files` despite the README/manual Claude example including it
## Summary

`codegraph install`'s Claude target writes a `permissions.allow` list that omits `mcp__codegraph__codegraph_files`.

That directly contradicts the current manual Claude setup examples in both the README and the docs site, which explicitly include `codegraph_files` in `~/.claude/settings.json`.

### Manual reproduction

From the repo root:

1. Build the current tree:

```bash
cd /path/to/codegraph
npm run build
```

2. Run the real installer against an isolated temp home so it cannot touch your actual Claude config:

```bash
tmp="$(mktemp -d)"
HOME="$tmp/home" node dist/bin/codegraph.js install --target=claude --location=global --yes
cat "$tmp/home/.claude/settings.json"
```

3. Confirm the generated allowlist does **not** contain `mcp__codegraph__codegraph_files` : 

```json
{
  "permissions": {
    "allow": [
      "mcp__codegraph__codegraph_search",
      "mcp__codegraph__codegraph_context",
      "mcp__codegraph__codegraph_callers",
      "mcp__codegraph__codegraph_callees",
      "mcp__codegraph__codegraph_impact",
      "mcp__codegraph__codegraph_node",
      "mcp__codegraph__codegraph_status"
    ]
  }
}
```

## Expected behavior

The installer-generated Claude allowlist should include `mcp__codegraph__codegraph_files`, matching the documented manual Claude configuration.

## Actual behavior

`src/installer/targets/shared.ts` hardcodes a seven-entry Claude permission list that stops at `codegraph_status` and omits `codegraph_files`.

## Docs evidence

- `README.md:301-307` says the installer will "Set up auto-allow permissions when Claude Code is one of the targets".
- `README.md:362-377` includes `mcp__codegraph__codegraph_files` in the manual `~/.claude/settings.json` example.
- `site/src/content/docs/reference/integrations.md:43-60` includes the same `codegraph_files` entry in the manual Claude allowlist example.
- `site/src/content/docs/reference/mcp-server.md:16-27` lists `codegraph_files` as a first-class MCP tool.

This is a direct installer-vs-docs contradiction.

## Likely root cause

This looks like stale hand-maintained installer state:

- The original Claude-only installer introduced the current 7-entry allowlist on 2026-01-19 in commit `9baf59e`.
- `codegraph_files` was added to the MCP tool surface on 2026-02-09 (`d0ee6f7`).
- The README manual Claude allowlist example was updated to include `codegraph_files` on 2026-02-10 (`4c78276`).
- The multi-target installer refactor on 2026-05-17 (`a447e1d`) copied the older allowlist into `src/installer/targets/shared.ts`.

So the docs and tool surface moved forward, but the installer allowlist did not.

## Suggested fix

- Add `mcp__codegraph__codegraph_files` to `getCodeGraphPermissions()` in `src/installer/targets/shared.ts`.

## Scope note

This same stale allowlist also omits `codegraph_explore` and `codegraph_trace`.

I am not promoting those as the main issue here because `codegraph_files` is the direct README/docs contradiction. Maintainers can decide in a follow-up issue/PR whether the Claude auto-allow list should also expand to cover `codegraph_explore` and `codegraph_trace`.

## Affected files

- `src/installer/targets/shared.ts`
- `src/installer/targets/claude.ts`
- `src/mcp/tools.ts`
- `README.md`
- `site/src/content/docs/reference/integrations.md`
- `site/src/content/docs/reference/mcp-server.md`

### LINKED ISSUE #571: codegraph_context task parameter description misleads models into passing natural language instead of keywords

Ununp3ntium115:
## Summary

- Add `codegraph_files` to Claude auto-allow permissions, create non-destructive installer backups, and make `codegraph_context` guidance/schema keyword-oriented.
- Treat ArkTS `.ets` files as TypeScript for first-pass indexing.
- Preserve UTF-8 Git paths for non-ASCII directories, including macOS/CJK paths, and block project-root escapes through symlinked files or directories.
- Add contributor setup docs covering local CLI linking and the macOS Node/FTS5 source-build workaround.

## Root Cause

Claude-facing setup had drifted from the current MCP tool surface, the installer could overwrite existing agent config without a reliable restore point, and `codegraph_context` wording encouraged agents to send verbose natural-language tasks into a keyword-backed lookup.

For the macOS/path issues, Git quotes non-ASCII paths by default, so tracked files under CJK directories could reach extension detection as quoted octal escapes. Path containment also only used lexical `path.resolve`, so a symlink inside the project could resolve outside the project root after passing validation.

## Issues

Fixes #565
Fixes #486
Fixes #571
Fixes #512
Fixes #541
Fixes #527
Fixes #305

## Validation

- `npx -y -p node@22 node ./node_modules/vitest/vitest.mjs run` (52 files, 1102 passed, 2 skipped)
- `npm test -- __tests__/extraction.test.ts`
- `npm test -- __tests__/security.test.ts __tests__/installer.test.ts __tests__/installer-targets.test.ts __tests__/mcp-tool-allowlist.test.ts`
- `npm run build`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -544,6 +544,80 @@ add a negation — `!vendor/`. The defaults apply uniformly, so committing a
 dependency or build directory doesn't force it into the graph; the `.gitignore`
 negation is the explicit opt-in.
 
+## Contributing
+
+Set up a local checkout from your fork:
+
+```bash
+git clone git@github.com:<your-user>/codegraph.git
+cd codegraph
+git remote add upstream https://github.com/colbymchenry/codegraph.git
+npm install
+npm run build
+```
+
+To point the global `codegraph` command at your local build while developing:
+
+```bash
+npm link
+codegraph --version
+```
+
+To switch back to the published package:
+
+```bash
+npm unlink -g @colbymchenry/codegraph
+npm i -g @colbymchenry/codegraph@latest
+```
+
+You can also run the local build without changing the global command:
+
+```bash
+npm run cli -- status
+```
+
+Useful development commands:
+
+```bash
+npm run dev          # TypeScript watch build
+npm test             # full test suite
+npm test -- <file>   # focused test file
+npm run test:watch   # Vitest watch mode
+```
+
+### macOS Source Builds and FTS5
+
+The self-contained release and npm package use CodeGraph's bundled runtime, so
+end users do not need to manage SQLite or FTS5. Contributors running directly
+from source use their local Node runtime instead. On macOS, some official Node
+22.x and 23.x builds include `node:sqlite` without FTS5 enabled, which makes
+indexing fail with `no such module: fts5`.
+
+Use Node 24.x for local development on macOS if you see that error:
+
+```bash
+nvm install 24
+nvm use 24
+npm install
+npm run build
+```
+
+Quick FTS5 check before indexing:
+
+```bash
+node - <<'NODE'
+const { DatabaseSync } = require('node:sqlite');
+const db = new DatabaseSync(':memory:');
+db.exec('CREATE VIRTUAL TABLE cg_fts_check USING fts5(value)');
+db.close();
+console.log('FTS5 available');
+NODE
+```
+
+PR hygiene: do not bump `package.json` versions unless you are preparing a
+release, and do not update `package-lock.json` unless your change modifies
+dependencies.
+
 ## Supported Platforms
 
 Every release ships a self-contained build (bundled Node runtime — nothing to
--- a/__tests__/extraction.test.ts
+++ b/__tests__/extraction.test.ts
@@ -33,6 +33,7 @@ function cleanupTempDir(dir: string): void {
 describe('Language Detection', () => {
   it('should detect TypeScript files', () => {
     expect(detectLanguage('src/index.ts')).toBe('typescript');
+    expect(detectLanguage('entry/src/main/ets/pages/Index.ets')).toBe('typescript');
     expect(detectLanguage('components/Button.tsx')).toBe('tsx');
   });
 
@@ -3380,6 +3381,44 @@ describe('Directory Exclusion', () => {
 
     expect(files).toContain('src/index.ts');
     expect(files.every((f) => !f.includes('.git'))).toBe(true);
+  });
+
+  it('should index git-visible files under non-ASCII directories (issue #541)', async () => {
+    const { execFileSync } = await import('child_process');
+    const git = (...args: string[]) =>
+      execFileSync('git', args, { cwd: tempDir, stdio: 'pipe' });
+
+    git('init', '-q');
+    fs.mkdirSync(path.join(tempDir, 'src', 'english'), { recursive: true });
+    fs.mkdirSync(path.join(tempDir, 'src', '中文目录'), { recursive: true });
+    fs.writeFileSync(
+      path.join(tempDir, 'src', 'english', 'Foo.cs'),
+      'namespace Demo;\npublic class Foo { public void Bar() {} }\n',
+    );
+    fs.writeFileSync(
+      path.join(tempDir, 'src', '中文目录', 'Baz.cs'),
+      'namespace Demo;\npublic class Baz { public void Qux() {} }\n',
+    );
+    git('add', '-A');
+
+    const scanned = scanDirectory(tempDir).sort();
+    expect(scanned).toEqual(['src/english/Foo.cs', 'src/中文目录/Baz.cs']);
+
+    const cg = CodeGraph.initSync(tempDir);
+    try {
+      const result = await cg.indexAll();
+      expect(result.filesIndexed).toBe(2);
+      expect(cg.getFiles().map((f) => f.path).sort()).toEqual(scanned);
+
+      fs.writeFileSync(
+        path.join(tempDir, 'src', '中文目录', 'Baz.cs'),
+        'namespace Demo;\npublic class Baz { public void Qux() {} public void Zap() {} }\n',
+      );
+      const changed = cg.getChangedFiles();
+      expect([...changed.added, ...changed.modified]).toContain('src/中文目录/Baz.cs');
+    } finally {
+      cg.close();
+    }
   });
 
   it('should return forward-slash paths on all platforms', () => {
--- a/__tests__/installer-targets.test.ts
+++ b/__tests__/installer-targets.test.ts
@@ -880,6 +880,14 @@ describe('Installer targets — partial-state idempotency', () => {
     expect(cfg.mcpServers.codegraph).toBeDefined();
   });
 
+  it('claude: auto-allow includes codegraph_files (#565)', () => {
+    const claude = getTarget('claude')!;
+    claude.install('local', { autoAllow: true });
+
+    const settings = JSON.parse(fs.readFileSync(path.join(tmpCwd, '.claude', 'settings.json'), 'utf-8'));
+    expect(settings.permissions.allow).toContain('mcp__codegraph__codegraph_files');
+  });
+
   it('claude: install does NOT create a CLAUDE.md instructions file (#529)', () => {
     const claude = getTarget('claude')!;
     const result = claude.install('local', { autoAllow: false });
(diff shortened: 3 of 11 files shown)


Ununp3ntium115:
Superseded by the focused draft PRs #600 (Claude installer/MCP context fixes) and #601 (macOS path/symlink/docs fixes). Closing this combined draft so maintainers can review the smaller branches independently.

PR #601, opened 2026-05-31
[codex] Fix macOS path and symlink edge cases

### LINKED ISSUE #305: docs: add Contributing section to README (local dev setup, FTS5 workaround)

### LINKED ISSUE #527: Bug: validatePathWithinRoot allows reading files outside project root via symlinks

### LINKED ISSUE #541: Files under directories with non-ASCII (CJK) names are silently skipped during indexing

Ununp3ntium115:
## Summary

This PR fixes the macOS/path-related issue reports by hardening path handling and documenting the local development trap around FTS5:

- Disable Git path quoting for path-emitting commands so tracked and changed files under non-ASCII/CJK directories are parsed as UTF-8 paths.
- Preserve path bytes except the final record separator, so valid path names with spaces are not accidentally trimmed while parsing Git output.
- Add realpath containment checks to `validatePathWithinRoot` so symlinked files and symlinked parent directories cannot escape the project root.
- Skip symlink targets that resolve outside the root during filesystem scanning before they can be queued for indexing.
- Add README contributor setup docs, including `npm link`, focused test commands, and the macOS Node/FTS5 source-build workaround.

## Issue Mapping

- Fixes #541: Git-visible files under non-ASCII/CJK directories are indexed and detected as changed.
- Fixes #527: symlink escapes through files or parent directories are blocked before reads/indexing.
- Fixes #305: README now documents local contributor setup and the macOS FTS5 workaround.

## Root Cause

Git quotes non-ASCII path bytes by default (`core.quotePath=true`), returning quoted octal escapes for CJK path segments. CodeGraph parsed that output as normal UTF-8 file paths, so extension checks could silently reject otherwise supported source files.

Path containment also relied on lexical `path.resolve` checks. A path like `src/secret.ts` could pass because it was textually inside the project, even if `src/secret.ts` was a symlink to a file outside the root. The fix keeps the lexical check, then verifies existing targets and nearest existing parents with `realpath` so both direct symlink files and symlinked directories are rejected.

For contributors, the README only covered end-user installation. That left macOS source builds with no guidance for the `node:sqlite` / FTS5 mismatch that can appear on some local Node builds.

## Validation

- `npx -y -p node@22 node ./node_modules/vitest/vitest.mjs run __tests__/extraction.test.ts __tests__/security.test.ts`
  - 2 files passed, 317 passed, 2 skipped
- `npm run build`
- `git diff --check`

## Notes

This PR is split from the Claude installer/MCP guidance fixes so the path/security changes can be reviewed independently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -544,6 +544,80 @@ add a negation — `!vendor/`. The defaults apply uniformly, so committing a
 dependency or build directory doesn't force it into the graph; the `.gitignore`
 negation is the explicit opt-in.
 
+## Contributing
+
+Set up a local checkout from your fork:
+
+```bash
+git clone git@github.com:<your-user>/codegraph.git
+cd codegraph
+git remote add upstream https://github.com/colbymchenry/codegraph.git
+npm install
+npm run build
+```
+
+To point the global `codegraph` command at your local build while developing:
+
+```bash
+npm link
+codegraph --version
+```
+
+To switch back to the published package:
+
+```bash
+npm unlink -g @colbymchenry/codegraph
+npm i -g @colbymchenry/codegraph@latest
+```
+
+You can also run the local build without changing the global command:
+
+```bash
+npm run cli -- status
+```
+
+Useful development commands:
+
+```bash
+npm run dev          # TypeScript watch build
+npm test             # full test suite
+npm test -- <file>   # focused test file
+npm run test:watch   # Vitest watch mode
+```
+
+### macOS Source Builds and FTS5
+
+The self-contained release and npm package use CodeGraph's bundled runtime, so
+end users do not need to manage SQLite or FTS5. Contributors running directly
+from source use their local Node runtime instead. On macOS, some official Node
+22.x and 23.x builds include `node:sqlite` without FTS5 enabled, which makes
+indexing fail with `no such module: fts5`.
+
+Use Node 24.x for local development on macOS if you see that error:
+
+```bash
+nvm install 24
+nvm use 24
+npm install
+npm run build
+```
+
+Quick FTS5 check before indexing:
+
+```bash
+node - <<'NODE'
+const { DatabaseSync } = require('node:sqlite');
+const db = new DatabaseSync(':memory:');
+db.exec('CREATE VIRTUAL TABLE cg_fts_check USING fts5(value)');
+db.close();
+console.log('FTS5 available');
+NODE
+```
+
+PR hygiene: do not bump `package.json` versions unless you are preparing a
+release, and do not update `package-lock.json` unless your change modifies
+dependencies.
+
 ## Supported Platforms
 
 Every release ships a self-contained build (bundled Node runtime — nothing to
--- a/__tests__/extraction.test.ts
+++ b/__tests__/extraction.test.ts
@@ -3382,6 +3382,44 @@ describe('Directory Exclusion', () => {
     expect(files.every((f) => !f.includes('.git'))).toBe(true);
   });
 
+  it('should index git-visible files under non-ASCII directories (issue #541)', async () => {
+    const { execFileSync } = await import('child_process');
+    const git = (...args: string[]) =>
+      execFileSync('git', args, { cwd: tempDir, stdio: 'pipe' });
+
+    git('init', '-q');
+    fs.mkdirSync(path.join(tempDir, 'src', 'english'), { recursive: true });
+    fs.mkdirSync(path.join(tempDir, 'src', '中文目录'), { recursive: true });
+    fs.writeFileSync(
+      path.join(tempDir, 'src', 'english', 'Foo.cs'),
+      'namespace Demo;\npublic class Foo { public void Bar() {} }\n',
+    );
+    fs.writeFileSync(
+      path.join(tempDir, 'src', '中文目录', 'Baz.cs'),
+      'namespace Demo;\npublic class Baz { public void Qux() {} }\n',
+    );
+    git('add', '-A');
+
+    const scanned = scanDirectory(tempDir).sort();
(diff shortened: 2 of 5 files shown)


Arvin-Hugh:
[Files under directories with non-ASCII (CJK) names are silently skipped during indexing #541](https://github.com/colbymchenry/codegraph/issues/541)

Looking forward to the update!

colbymchenry:
Thanks for surfacing these macOS path/symlink edge cases — closing as superseded, since both halves have since landed on `main` independently:

- The symlink path-escape (the #527 part) shipped in #724, with the same realpath approach.
- The non-ASCII/CJK git-path handling (#541) is already in via `git ls-files -z` (NUL-separated, unquoted output), which is robust for arbitrary path bytes; #541 is closed.

So there's nothing left to merge here, but the edge cases you flagged were real and are now covered. Appreciate it!

PR #624, opened 2026-06-01
fix(installer): preserve Codex TOML array tables

chiehwangs:
## Summary

- preserve TOML `[[array-of-tables]]` sections that follow the Codex MCP entry
- add regression coverage for both reinstall and uninstall paths
- document the user-visible installer fix in the changelog

## Root cause

The Codex TOML helper located the end of `[mcp_servers.codegraph]` by searching for the next table header, but intentionally skipped `[[array-of-tables]]` headers. When an array table followed the CodeGraph MCP entry, reinstalling or uninstalling CodeGraph treated that sibling configuration as part of the CodeGraph block and removed it.

## Validation

- `git diff --check`
- `npx vitest run __tests__/installer-targets.test.ts`
- `npm run build`
- full Vitest suite with fork workers using `--liftoff-only`: 52 test files passed, 1096 tests passed, 2 skipped

## Notes

On Node v24.15.0, plain `npm test` reproduces the repository-documented V8 WASM Turboshaft `Fatal process out of memory: Zone` failure. Passing `--liftoff-only` to Vitest fork workers applies the existing runtime workaround and completes the full suite.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ### Fixes
 
 - Indexing a project that contains only config-style files (YAML, Twig, or `.properties`) no longer misleadingly reports "No files found to index" — these files are tracked at the file level and are now counted as indexed. Thanks @luojiyin1987 (#357).
+- Codex installs and uninstalls now preserve TOML array-of-table sections that appear after the CodeGraph MCP entry, so unrelated user configuration is no longer removed.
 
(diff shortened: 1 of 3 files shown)


colbymchenry:
Thanks for this, @chiehwangs — you were first to spot and fix the `[[array-of-tables]]` deletion bug back in June, and your diagnosis was spot on. We've just landed a fix for the same issue in #1370, which also carries the multiline-string edge cases and the install→reinstall→uninstall round-trip regression tests, so I'm closing this as superseded. Genuinely appreciate you flagging it — sorry it sat as long as it did. Hope to see more PRs from you.

PR #625, opened 2026-06-01
fix(mcp): clear initialize response suppression after first match

chiehwangs:
## Description

The local-handshake proxy immediately returns an initialize response while the shared daemon connects in the background. It forwards the initialize request to the daemon and suppresses the daemon's duplicate response. However, a later JSON-RPC request can reuse the initialize request ID after the handshake completes. In that case, the proxy incorrectly drops the later response.

## Root Cause

The proxy stores the initialize request ID in `clientInitId` to identify the daemon's duplicate initialize response. After filtering the matching response, it leaves `clientInitId` unchanged. The suppression logic therefore continues matching every later daemon response with the same request ID.

## Fix

Clear `clientInitId` immediately after suppressing the first matching daemon response. Add a real daemon regression test that initializes with ID 1, reuses ID 1 for a subsequent ping request, and verifies that the later response is relayed to the client.


chiehwangs:
Updated this PR against the latest `main`; it is mergeable now.

Verification:
- `npm run build`
- `__tests__/mcp-daemon.test.ts`: 8/8 passed
- Full Vitest suite with `--liftoff-only`: 59/59 test files passed, 1129 passed, 2 skipped

This is a small MCP proxy fix with a regression test for JSON-RPC request ID reuse after `initialize`.

PR #646, opened 2026-06-02
fix(cpp): resolve singleton/factory/chained calls via return types

### LINKED ISSUE #645: C++: method calls through singletons, factories, and chained getters resolve to the wrong class (or not at all)
## Summary

In C++, a method call whose receiver is **another call's result** loses the
receiver's type during extraction. The call is recorded as a bare method name,
so resolution can't tell which class it belongs to. When two classes share a
method name, the call silently resolves to **whichever class was indexed first**
— or doesn't resolve at all. This corrupts `callers`, `callees`, `impact`, and
`trace`, and (worse than a miss) does so silently, with a plausible-looking
wrong edge.

This hits some of the most common C++ idioms: singletons (`Foo::instance()`),
factories, and chained getters.

## Reproduction

Two classes sharing a method name `writeLog`. `Logger` sorts first, so it wins
any name-only tie. Every call below targets **`Metrics`**.

`logger.hpp` / `metrics.hpp`:
```cpp
class Logger  { public: static Logger&  instance(); void writeLog(const std::string&); };
class Metrics { public: static Metrics& instance(); void writeLog(const std::string&); };
```
(+ out-of-line definitions in a `.cpp`)

`app.cpp`:
```cpp
void a() { Metrics::instance().writeLog("x"); }              // chained singleton
void b() { auto& m = Metrics::instance(); m.writeLog("x"); } // stored in auto
```

Index, then query callers of each `writeLog`.

### Expected
Both `a` and `b` are callers of **`Metrics::writeLog`**.

### Actual (0.9.8 / `main`)
- `a` resolves to **`Logger::writeLog`** ❌ (mis-attributed to the first-indexed
  same-named method).
- `b` produces **no `calls` edge to any `writeLog`** ❌ (the `auto` receiver type
  is never recovered).

Only the explicitly-typed form `Metrics& m = Metrics::instance(); m.writeLog()`
resolves correctly today.

## Why it matters

CodeGraph's value is letting an agent answer flow/impact questions without
reading source. A *silently wrong* caller edge is worse than a missing one: the
agent trusts a plausible-but-incorrect relationship. And the trigger is
universal — any codebase with two same-named methods (extremely common) hits it
the moment a singleton/factory call is involved.

## Root cause

For `Foo::instance().bar()`, tree-sitter produces a `field_expression` whose
receiver is the inner `Foo::instance()` call. The call extractor only keeps a
receiver when it's a plain identifier, so a call-expression receiver is dropped
and the callee degrades to the bare name `bar`. Resolution then falls back to
name matching, which ties between same-named methods and picks the first-indexed
one. C++ also captures no return-type information today (method signatures are
empty), so there's nothing to recover the receiver's type from.

## Scope / proposed direction

Resolve the receiver by **what the inner call returns**, which requires first
capturing C++ return types during extraction. In order of value:

- [ ] Singletons / self-returning accessors: `Foo::instance().bar()`,
      `Foo::getInstance()->bar()` (any accessor name, not just `instance`).
- [ ] Factories returning a *different* type: `WidgetFactory::create().draw()`
      resolves on `Widget`, not `WidgetFactory`.
- [ ] Free-function factories: `openSession()->run()`.
- [ ] The same patterns stored in an `auto` local first, plus `new` /
      `std::make_unique` / `std::make_shared` / casts / direct construction.
- [ ] Single-level member chains: `manager.view().render()`.

**Deliberately out of scope** (need a real type environment — symbol tables +
overload resolution by argument types — not a heuristic; can be tracked
separately):
- Deep chains `a().b().c()`.
- Multi-level member access `h.mgr.view().render()`.
- Overload-correct selection, `typedef`/`using` alias resolution, templated
  return types, inherited methods.

Safety: every inferred type should be validated against the graph (the class
must actually have the method) before an edge is created, so a wrong guess falls
through silently rather than producing a wrong edge.

## Environment
- Reproduced on `0.9.8` and `main`.
- Language: C++ (tree-sitter).

stabey:
Fixes #645.

## Problem

In C++, a method call whose receiver is **another call's result** lost the
receiver's type during extraction and degraded to a bare method name. When two
classes shared a method name, the call silently resolved to whichever class was
indexed first — or didn't resolve at all. This corrupted `callers`, `callees`,
`impact`, and `trace` for the most common C++ idioms (singletons, factories,
chained getters), and did so *silently* with a plausible-looking wrong edge.

## Approach

Resolve the receiver by **what the inner call returns**. This required first
capturing C++ return types, which weren't extracted at all before.

- **Extraction** — capture C++ return types into method/function signatures
  (`extractCppSignature`, emitted as `(params) -> ReturnType`; base type from the
  `type` field, smart-pointer template arg preserved). Encode chained-call
  receivers as a resolvable reference (`Recv().method`, `obj.getThing().method`)
  instead of dropping them to a bare name.
- **Resolution** — `matchCppChainedAccessor` resolves those by the callee's
  return type, with a self-returning-accessor-*name* fallback when the return
  type isn't indexed (e.g. an external accessor). `auto` locals are inferred from
  `new` / `std::make_unique` / `std::make_shared` / casts / direct construction /
  named accessors / the initializer call's return type. Single-level member
  chains resolve the object's type, then the chained method's return type.

## Covered

- [x] Singletons / self-returning accessors: `Foo::instance().bar()`,
      `Foo::getInstance()->bar()` (any accessor name, not just `instance`).
- [x] Factories returning a *different* type: `WidgetFactory::create().draw()`
      resolves on `Widget`, not `WidgetFactory`.
- [x] Free-function factories: `openSession()->run()`.
- [x] The same patterns stored in an `auto` local first, plus `new` /
      `std::make_unique` / `std::make_shared` / casts / direct construction.
- [x] Single-level member chains: `manager.view().render()`.

## Deliberately out of scope

Need a real type environment (symbol tables + overload resolution by argument
types), not a heuristic — left uncovered (silent, not wrong):

- Deep chains `a().b().c()`.
- Multi-level member access `h.mgr.view().render()`.
- Overload-correct selection, `typedef`/`using` alias resolution, templated
  return types, inherited methods.

## Safety

Every inferred type is validated against the graph (the class must actually have
the method) before an edge is created, so a wrong guess falls through silently
rather than producing a wrong edge.

## Testing

- New end-to-end tests in `frameworks-integration.test.ts` for the singleton,
  factory-returns-other-class, oddly-named-accessor, free-function-factory, and
  member-chain cases — each with a same-named decoy class that sorts first, so a
  green test proves resolution was driven by type, not indexing order.
- New unit tests in `resolution.test.ts` for `inferCppTypeFromInitializer` and
  `parseCppReturnType` (tier boundaries, smart-pointer unwrap, primitive/void
  rejection).
- Node count stable across re-index (no node explosion).
- Full suite green (1110 passed / 2 skipped).

### DIFF
--- a/__tests__/frameworks-integration.test.ts
+++ b/__tests__/frameworks-integration.test.ts
@@ -484,6 +484,60 @@ describe('C++ end-to-end — virtual override synthesis', () => {
     }
   });
 
+  it('does not match a same-suffix class when inferring a chained method return type (Manager vs OtherManager)', async () => {
+    // Regression for the suffix-boundary bug (Codex review, PR #646): the
+    // chained-method return-type lookup matched a qualified name by `endsWith`,
+    // so resolving `view()` on `Manager` could pick `OtherManager::view`
+    // (its name ends with `Manager::view`). OtherManager sorts first and returns
+    // a different type, so without a `::` boundary check the final `render()`
+    // misroutes to that type's class.
+    tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'cg-cpp-'));
+    let cg: CodeGraph | undefined;
+    try {
+      // `other.hpp` sorts before `types.hpp`, so OtherManager indexes first and
+      // would win a naive endsWith('Manager::view') match.
+      fs.writeFileSync(
+        path.join(tmpDir, 'other.hpp'),
+        'class Decoy { public: void render(); };\n' +
+          'class OtherManager { public: Decoy view(); };\n'
+      );
+      fs.writeFileSync(
+        path.join(tmpDir, 'types.hpp'),
+        'class Panel { public: void render(); };\n' +
+          'class Manager { public: Panel view(); };\n'
+      );
+      fs.writeFileSync(
+        path.join(tmpDir, 'defs.cpp'),
+        '#include "other.hpp"\n' +
+          '#include "types.hpp"\n' +
+          'void Decoy::render() {}\n' +
+          'Decoy OtherManager::view() { return Decoy(); }\n' +
+          'void Panel::render() {}\n' +
+          'Panel Manager::view() { return Panel(); }\n'
+      );
+      fs.writeFileSync(
+        path.join(tmpDir, 'app.cpp'),
+        '#include "types.hpp"\n' +
+          'void useManager() { Manager m; m.view().render(); }\n'
+      );
+
+      cg = CodeGraph.initSync(tmpDir);
+      await cg.indexAll();
+
+      const panelRender = cg.getNodesByKind('method').find((n) => n.qualifiedName === 'Panel::render');
+      const decoyRender = cg.getNodesByKind('method').find((n) => n.qualifiedName === 'Decoy::render');
+      expect(panelRender, 'Panel::render node').toBeDefined();
+      expect(decoyRender, 'Decoy::render node').toBeDefined();
+
+      // Manager::view() returns Panel — resolve render() on Panel, never on the
+      // same-suffix OtherManager::view() (which returns Decoy).
+      expect(cg.getCallers(panelRender!.id).map((c) => c.node.qualifiedName)).toContain('useManager');
+      expect(cg.getCallers(decoyRender!.id).map((c) => c.node.qualifiedName)).not.toContain('useManager');
+    } finally {
+      cg?.close();
+    }
+  });
+
   it('bridges a base virtual method to the subclass override', async () => {
     tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'cg-cpp-'));
     fs.writeFileSync(
--- a/__tests__/frameworks-integration.test.ts
+++ b/__tests__/frameworks-integration.test.ts
@@ -538,6 +538,57 @@ describe('C++ end-to-end — virtual override synthesis', () => {
     }
   });
 
+  it('resolves a bare factory call to the free function, not a same-named method', async () => {
+    // Regression for Codex review (PR #646): a bare `make()->draw()` call encodes
+    // as `make().draw`, an unqualified callee. The return-type lookup must use
+    // the free function `make`, not an unrelated `Decoy::make` method that
+    // happens to share the name and sorts first — otherwise draw() misroutes to
+    // whatever that method returns.
+    tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'cg-cpp-'));
+    let cg: CodeGraph | undefined;
+    try {
+      fs.writeFileSync(
+        path.join(tmpDir, 'other.hpp'), // sorts first → Decoy::make indexed first
+        'class Decoy { public: void draw(); };\n' +
+          'class Other { public: Decoy make(); };\n'
+      );
+      fs.writeFileSync(
+        path.join(tmpDir, 'widget.hpp'),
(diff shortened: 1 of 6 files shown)


stabey:
Thanks @codex — both P2 findings were valid. The return-type lookups (`cppReturnTypeOf` namespace fallback and `cppReturnTypeOfMethodOnType` member-chain) matched qualified names with a plain `endsWith`, so `Foo::create` also accepted `OtherFoo::create` and `Manager::view` accepted `OtherManager::view`. If the same-suffix class indexed first and returned a different type, the chained/auto call misrouted.

Fixed in e424e31: both now match at a `::` boundary (exact or `::<suffix>`), mirroring `resolveMethodOnType`. Added a regression test with an `OtherManager` decoy that sorts first — verified red before the fix, green after.

stabey:
Rebased onto latest `main` (resolving the conflict with #647 — dropped this PR's CHANGELOG entry so it no longer touches that file) and addressed the third finding.

The bare-callee branch of `cppReturnTypeOf` (used for `makeWidget()->draw()`-style unqualified chained calls) included same-named *methods*, so an unrelated `Decoy::makeWidget` indexed first could supply the return type and misroute the call. A bare unqualified call is a free function, so the lookup is now restricted to `kind === 'function'`. Added a regression test with a same-named method that sorts first (red before, green after). Full suite green (1126 passed).

stabey:
Addressed the namespace finding (4th Codex comment).

The chained scoped-call encoding kept only the segment before the accessor, so `a::Factory::create()` and `b::Factory::create()` collapsed to the same `Factory::create`. Now the **full qualifier** is encoded (`a::Factory::create().draw`), and the return-type lookup matches most-specific first (exact qualified name → full-qualifier `::` suffix → `Class::accessor`) and **bails when matches disagree** on the return type, so an ambiguous class name across namespaces resolves to the right one (or stays silent) rather than guessing. The accessor-name fallback now keeps the namespace in the owner type too. Added a regression test with a `Factory` class shared across two namespaces. Full suite green (1127 passed).

stabey:
Addressed the 5th finding (ambiguous return-type basename across namespaces).

The return type is normalized to a bare basename, so `a::Factory::create()` returning `a::Widget` and a separate `b::Widget` collapse to `Widget`. The return-type-driven chained resolutions now go through a wrapper that bails when the basename has **more than one owner across namespaces** (overloads on the *same* owner still resolve normally) — so it stays silent rather than linking to whichever `::Widget::draw` was indexed first. Added a regression test (verified red before the guard, green after). Full suite green (1128 passed).

stabey:
Addressed both findings from the latest review — both common modern-C++ syntax, so good catches:

- **Trailing return types** (`auto Factory::create() -> Widget`): the `type` field is just `auto`; the real type lives on the declarator's trailing-return node. Extraction now reads that, so the signature is `() -> Widget` and `Factory::create().draw()` / the `auto` forms resolve instead of hitting `() -> auto`.
- **cv-qualified smart-pointer pointees** (`std::shared_ptr<const Widget>`): the unwrap regex matched `const` as the type. The smart-pointer, `make_*`, and cast patterns now skip a leading `const`/`volatile` and capture `Widget`.

Added unit tests (`parseCppReturnType`, `inferCppTypeFromInitializer`) and an end-to-end trailing-return test. Full suite green (1131 passed).

PR #668, opened 2026-06-03
feat(cli): add codegraph visualize for a standalone HTML graph

xy200303:
## What this adds

`codegraph visualize [symbol]` writes an interactive graph of the indexed project to a **single self-contained HTML file** — no server, no network, no third-party script.

```bash
codegraph visualize                                    # whole project -> .codegraph/graph.html
codegraph visualize handleRequest --depth 3 --limit 500
codegraph visualize --kind class,function,method --edge-kind calls,extends,implements
codegraph visualize --open                             # write it, then open it
```

Without a symbol it draws a project overview; with a symbol it draws that symbol's neighbourhood. The page carries its own node/edge filters and layout controls (gravity, repulsion, link distance, fit/spread/cluster/pause), so the graph can be narrowed without regenerating the file.

## Why this needed rebasing

The branch was written against a tree ~619 commits behind, and the pieces the command relied on had moved:

- `src/types.ts` had since grown the runtime `EDGE_KINDS` list the module reads, and the `union` node kind was added.
- The helpers the command used — `parseBoundedInteger`, `parseAllowedList`, `parseDirection`, `slugForFileName` — no longer exist.

So this revision keeps the graph builder and the HTML renderer as written (their type layer is unchanged; the only edit they needed was a `KIND_WEIGHT` entry for `union`) and re-adds the four small flag helpers. Two deliberate changes against current conventions:

- `--open` now goes through the existing `openBrowser` helper, so it honours `CODEGRAPH_BROWSER` exactly like `codegraph ui` instead of carrying a second platform opener.
- The graph is closed with `close()` rather than the deprecated `destroy()`, in a `finally`, so a render failure cannot leak the connection.

## `visualize` vs `codegraph ui`

These answer different needs and I'd rather be explicit about it: `ui` is the live viewer — it serves an index over `127.0.0.1`, re-reads the graph as the project changes, and saves trails. `visualize` writes one file you can attach to a review, send to someone, or open on a machine where running a server is awkward. If you'd rather this existed as a view inside the `ui` viewer instead of a second command, say so and I'll re-scope it — the graph-building half is the part worth keeping either way.

## Verification

Run end to end against a small indexed project (7 nodes / 9 edges):

| Check | Result |
| --- | --- |
| `visualize` (overview) | 7 nodes / 9 edges, 46 KB file |
| `visualize alpha` (symbol) | 6 nodes / 8 edges, focused subgraph |
| `visualize -k function` | 3 nodes / 2 edges |
| `visualize -k bogus` | exits 1, names the allowed kinds |
| `visualize --direction sideways` | exits 1 |
| `visualize -l 0` | clamped to 1 node |
| unindexed project | exits 1 with a clear message |
| generated page | no `<script src>`, no stylesheet link, no `http(s)://` reference |

- `npx vitest run __tests__/visualization.test.ts` — 3 passed (the two original, plus one asserting the page stays self-contained)
- `npx tsc --noEmit` — clean
- `npm run build` — clean
- Full suite compared against a clean baseline of the same checkout: **no file fails only with this change**. The failing files are the pre-existing Windows/environment ones already documented in `AGENTS.md` (symlink privileges, `EPERM` on temp-dir cleanup, POSIX file-mode and pid-based lock assertions).

## Not verified

I could not open a browser here, so the **rendered page itself is unverified** — the canvas drawing, the physics behaviour, and the controls are checked only insofar as the generated markup contains the expected elements and is syntactically sound. Please give it a look in a browser before merging; if the layout needs work I'm happy to iterate.

CHANGELOG entry added under `[Unreleased] → New Features`, and the CLI reference + README list the new command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -452,6 +452,7 @@ codegraph sync [path]             # Incremental update
 codegraph status [path]           # Show statistics
 codegraph query <search>          # Search symbols (--kind, --limit, --json)
 codegraph files [path]            # Show file structure (--format, --filter, --max-depth, --json)
+codegraph visualize [symbol]      # Generate interactive HTML graph (--depth, --limit, --output)
 codegraph callers <symbol>        # Find what calls a function/method (--limit, --json)
 codegraph callees <symbol>        # Find what a function/method calls (--limit, --json)
 codegraph impact <symbol>         # Analyze what code is affected by changing a symbol (--depth, --json)
--- a/__tests__/visualization.test.ts
+++ b/__tests__/visualization.test.ts
@@ -72,6 +72,11 @@ describe('visualization', () => {
     const html = renderVisualizationHtml(graph);
 
     expect(html).toContain('Graph &lt;danger&gt;');
+    expect(html).toContain('<canvas id="graph"');
+    expect(html).toContain('id="gravitySlider" type="range" min="0" max="140" value="18"');
+    expect(html).toContain('id="linkSlider" type="range" min="70" max="900" value="165"');
+    expect(html).toContain('id="spread"');
+    expect(html).toContain('function stepPhysics');
     expect(html).toContain('\\u003c/script\\u003e');
     expect(html).not.toContain('</script><img');
   });
--- a/__tests__/visualization.test.ts
+++ b/__tests__/visualization.test.ts
@@ -1,0 +1,78 @@
+import { describe, expect, it } from 'vitest';
+import {
+  buildVisualizationGraphFromSubgraph,
+  renderVisualizationHtml,
+} from '../src/visualization';
+import { Edge, Node, Subgraph } from '../src/types';
+
+function makeNode(id: string, name: string, kind: Node['kind'] = 'function'): Node {
+  return {
+    id,
+    kind,
+    name,
+    qualifiedName: `src/example.ts::${name}`,
+    filePath: 'src/example.ts',
+    language: 'typescript',
+    startLine: 1,
+    endLine: 3,
+    startColumn: 0,
+    endColumn: 1,
+    isExported: true,
+    updatedAt: 1,
+  };
+}
+
+describe('visualization', () => {
+  it('keeps root nodes and removes dangling edges when limiting a subgraph', () => {
+    const root = makeNode('root', 'root');
+    const child = makeNode('child', 'child');
+    const omitted = makeNode('omitted', 'omitted');
+    const edges: Edge[] = [
+      { source: root.id, target: child.id, kind: 'calls' },
+      { source: child.id, target: omitted.id, kind: 'calls' },
+    ];
+    const subgraph: Subgraph = {
+      nodes: new Map([
+        [root.id, root],
+        [child.id, child],
+        [omitted.id, omitted],
+      ]),
+      edges,
+      roots: [root.id],
+    };
+
+    const graph = buildVisualizationGraphFromSubgraph({
+      subgraph,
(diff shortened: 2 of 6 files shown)


xy200303:
Flagging this one for a scope decision rather than merging as-is — `main` has moved a long way (this branch is ~619 commits behind) and the viewer it adds now overlaps heavily with shipped functionality.

What `main` has today:

- `codegraph ui` (alias `web`) — "Open the CodeGraph viewer in your browser — read your indexed project as a graph", backed by the `src/ui-server/` package.
- That viewer already provides the graph-reading flows this PR is aiming at: pick a symbol and see callers on the left, source in the middle, callees on the right at the calling line; `a -> b` flow queries rendered as one card per hop with dynamic-dispatch hops drawn dashed; a **Map** tab that draws the whole project by module with dependencies pointing down; an Entry points tab (routes, load-time scripts, tests); live re-index watching; and saved "trails".
- There is no third-party graph dependency upstream either — the viewer ships as a static bundle served by `ui-server`.

This PR instead adds a standalone `visualize` command plus a self-contained ~1.8k-line Canvas force-graph in `src/visualization.ts`, with its own physics controls (gravity/repulsion/link distance) and layout actions. The physics controls and the animated Canvas force layout are the part that isn't obviously covered upstream, so the idea isn't without value — but the surface it targets (`src/bin/codegraph.ts`, `src/types.ts`, the CLI reference) has been rewritten since, and the docs/Cli entry points conflict.

Suggested path: rather than rebasing this branch, re-scope it against the current viewer — e.g. contribute the force-layout view as a tab inside the existing `ui` viewer, reusing `ui-server`'s graph APIs instead of a parallel `visualize` command. That keeps one viewer instead of two, and avoids re-deriving the graph data shape.

I'm happy to close this in favour of that, or to take a run at the re-scoped version if you'd prefer it kept open. Let me know which you'd like.

xy200303:
Rebased this onto current `main` and force-pushed, so it merges again instead of conflicting. I'm the contributor reworking this branch — thanks for the original implementation, the graph builder and renderer are unchanged.

Two notes worth your time:

**1. This supersedes my earlier comment suggesting the branch be closed.** I had read `codegraph ui` as covering the same ground, and it does cover the *reading a project as a graph* case. What changed my mind is that `visualize` writes a file rather than serving a port: you can attach the output to a review, hand it to someone else, or open it on a box where running a server is awkward, and it works with no network and no third-party script. That felt like a genuinely different affordance rather than a duplicate. If you disagree, the graph builder is the part worth keeping and I'd rather re-scope it into the `ui` viewer than leave two overlapping surfaces — just say which and I'll do it.

**2. What is and isn't verified.** I ran it end to end against a real indexed project — overview, symbol mode, `--kind` filtering, `--limit` clamping, and the error paths for an unknown kind, an unknown direction and an unindexed project all behave — and the generated page is confirmed self-contained. `tsc`, `npm run build` and the new test pass, and a full-suite run against a clean baseline of the same checkout shows no file failing only with this change.

What I could **not** do is look at it in a browser, so the drawing itself, the physics, and the layout controls are unverified. Please open the generated file before deciding; happy to iterate on how it looks.

The changes against current conventions are small and deliberate: `--open` reuses your `openBrowser` helper (so `CODEGRAPH_BROWSER` behaves like `codegraph ui`), and the connection is closed with `close()` in a `finally` instead of the deprecated `destroy()`. CHANGELOG entry is under `[Unreleased] → New Features`.

PR #670, opened 2026-06-03
fix: prefer ~/.config/opencode/ over %APPDATA% on Windows (#535)

### LINKED ISSUE #535: opencode config path on Windows uses %APPDATA% instead of ~/.config/opencode/
**Describe the bug**

On Windows, CodeGraph's opencode installer target writes config to `%APPDATA%\opencode\` instead of `~/.config/opencode\`. However, opencode uses `~/.config/opencode/` on **all platforms including Windows**, so the MCP config is never picked up.

This is a known issue affecting multiple projects — see:
- https://github.com/NoeFabris/opencode-antigravity-auth/issues/265
- https://github.com/NoeFabris/opencode-antigravity-auth/issues/251
- https://github.com/vercel-labs/skills/issues/208

**To Reproduce**

1. Install CodeGraph globally on Windows
2. Run `codegraph init -i` in a project
3. Select opencode as a target
4. Config is written to `%APPDATA%\opencode\` (wrong location)
5. OpenCode does not detect CodeGraph MCP server

**Expected behavior**

Config should be written to `~/.config/opencode/opencode.jsonc` on all platforms including Windows.

**Environment**

- OS: Windows
- CodeGraph version: latest
- OpenCode config path: `C:\Users\<user>\.config\opencode\opencode.jsonc`

**Workaround**

Manually create/edit `~/.config/opencode/opencode.jsonc`:

```json
{
  "mcp": {
    "codegraph": {
      "type": "local",
      "command": ["<path-to-codegraph-binary>", "serve", "--mcp"],
      "enabled": true
    }
  }
}
```

WodenJay:
## Problem

Fix #535 
On Windows, the codegraph installer always writes opencode's global config to `%APPDATA%/opencode/`, even when `~/.config/opencode/` already exists. opencode itself prefers the XDG-style `~/.config/` path when it exists on Windows, so the installer and the agent end up looking at different config files — the user's codegraph entry is "invisible" to opencode.

## Fix

`globalConfigDir()` now checks for `~/.config/opencode/` first on Windows. If it exists, that path is used (matching opencode's own behavior). Only when it doesn't exist does it fall back to `%APPDATA%/opencode/`.

## Changes

- `src/installer/targets/opencode.ts` — `globalConfigDir()`: add `fs.existsSync` guard on `~/.config/opencode/` before the APPDATA fallback (4 lines). Update file-header comment.
- `__tests__/installer-targets.test.ts` — two new `it.runIf('win32')` tests: (1) prefers `.config/opencode` when it exists; (2) falls back to APPDATA when it doesn't.

## Testing

```
npm test -- --testPathPattern installer-targets
# 139 passed (137 existing + 2 new)
```

Both new tests are Windows-gated and verified passing on a real Windows machine.

### DIFF
--- a/__tests__/installer-targets.test.ts
+++ b/__tests__/installer-targets.test.ts
@@ -339,6 +339,42 @@ describe('Installer targets — partial-state idempotency', () => {
     expect(fs.existsSync(path.join(process.cwd(), 'AGENTS.md'))).toBe(false);
   });
 
+  it.runIf(process.platform === 'win32')('opencode: prefers ~/.config/opencode/ over %APPDATA% when it exists (#535)', () => {
+    const opencode = getTarget('opencode')!;
+    // Simulate Windows layout where APPDATA and ~/.config are different
+    // directories. On Windows, globalConfigDir() should prefer
+    // ~/.config/opencode/ (XDG-style) when it exists, matching
+    // opencode's own behavior (issue #535).
+    const tmpRoaming = fs.mkdtempSync(path.join(os.tmpdir(), 'appdata-'));
+    process.env.APPDATA = tmpRoaming;
+    const configDir = path.join(tmpHome, '.config', 'opencode');
+    fs.mkdirSync(configDir, { recursive: true });
+    const result = opencode.install('global', { autoAllow: true });
+    expect(result.files[0].action).toBe('created');
+    // Must land under ~/.config/opencode, not %APPDATA%/opencode.
+    const normalised = result.files[0].path.replace(/\\/g, '/');
+    expect(normalised).toContain('.config/opencode');
+    expect(normalised).not.toMatch(/appdata-[^/]*\/opencode\/opencode\.jsonc$/);
+    fs.rmSync(tmpRoaming, { recursive: true, force: true });
+  });
+
+  it.runIf(process.platform === 'win32')('opencode: falls back to %APPDATA% when ~/.config/opencode/ does not exist (#535)', () => {
+    const opencode = getTarget('opencode')!;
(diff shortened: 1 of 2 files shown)


colbymchenry:
Thanks for taking this on, @WodenJay — your PR is what got #535 moving, and you're credited in the changelog for it.

I ended up landing a different approach in #836 and am closing this one as superseded. The reason is a detail of how opencode itself resolves the path: it uses the [`xdg-basedir`](https://github.com/sindresorhus/xdg-basedir) package ([sst/opencode `packages/core/src/global.ts`](https://github.com/sst/opencode/blob/dev/packages/core/src/global.ts)), which returns `XDG_CONFIG_HOME ?? ~/.config` **unconditionally on every platform** — there's no existence check and no `%APPDATA%` anywhere (that layout belonged to the discontinued Go fork).

That unconditional resolution is what breaks the prefer-if-exists fallback here: a user who installs codegraph *before* ever launching opencode has no `~/.config/opencode` yet, so this patch would still write to `%APPDATA%` — and opencode then creates `~/.config/opencode` on its first run and never sees the entry. Since opencode's resolution isn't conditional, ours can't be either.

#836 therefore writes XDG-style on all platforms (honoring `XDG_CONFIG_HOME`, which opencode also reads on Windows), and additionally sweeps a stale codegraph entry out of the legacy `%APPDATA%\opencode` location on install/uninstall so existing affected users self-heal. Validated end-to-end on a real Windows 11 machine — including real opencode 1.17.4 reporting `✓ codegraph connected` from the new path.

Appreciate the patch and the Windows test run — hope to see you in the contributors list again.

Issue #671, opened 2026-06-03
Rust support


durgesh-k-sharma:
Rust is not yet supported.

colbymchenry:
Hi @durgesh-k-sharma — Rust is actually fully supported, and has been for several releases. CodeGraph parses `.rs` files (structs, enums, traits, `impl` blocks, methods, functions, `use` imports), builds the call graph (`callers`/`callees`/`impact`), resolves trait implementations, and extracts web routes for **Axum, Actix-web, and Rocket** — plus cargo-workspace cross-crate resolution.

To use it:
1. `codegraph install` (wires up your agent)
2. `codegraph init` then `codegraph index` in your repo
3. `codegraph status` will list your `.rs` files under "Files by Language: rust"

If you ran it and Rust *didn't* index, could you share your version (`codegraph --version`), the `codegraph status` output for the repo, and whether the `.rs` files might be gitignored or under an excluded directory? And if you were expecting something specific (a framework beyond Axum/Actix/Rocket, a particular cross-crate query, or macro-generated code), tell us the exact case and we'll dig in.

Closing as already-supported — happy to reopen with a concrete repro. Thanks for raising it: it did surface that our supported-language list is harder to discover than it should be, which we're improving.

durgesh-k-sharma:
Ohh my bad that's what my agent came with out of the blue while doing a mini refactor maybe I shouldve probed it a little bit :)

Issue #529, opened 2026-05-28
CLAUDE.md instructions duplicate MCP server instructions — wastes tokens every turn


lbzepoqo:
## Problem

CodeGraph injects its usage instructions into the agent context **twice** per session:

1. **MCP `SERVER_INSTRUCTIONS`** — emitted in the MCP `initialize` response, surfaced by Claude Code as a system reminder (`## codegraph` block). This is automatic for any MCP server.
2. **`INSTRUCTIONS_TEMPLATE`** — written into `CLAUDE.md` (global or project-level) by `codegraph init`, wrapped in `<!-- CODEGRAPH_START -->` / `<!-- CODEGRAPH_END -->` markers.

Both contain ~90% identical content (tool selection table, rules of thumb, anti-patterns, staleness banner guidance). The CLAUDE.md version is loaded into the system prompt alongside the MCP server instructions, so the agent reads the same advice twice on every single turn.

### Token cost

The CLAUDE.md block is ~35 lines / ~1,500 tokens of input. Since it duplicates what the MCP server already provides, it burns ~1,500 extra input tokens **per message** for no benefit.

### Source files

- `SERVER_INSTRUCTIONS` in `mcp/server-instructions.ts` — the MCP-level injection
- `INSTRUCTIONS_TEMPLATE` in `installer/instructions-template.ts` — the CLAUDE.md injection
- `writeInstructionsEntry()` in `installer/targets/claude.ts` — writes the template into CLAUDE.md on `codegraph init`

## Suggested fix

Since the MCP server instructions are always injected by the MCP protocol, the CLAUDE.md block is redundant for Claude Code. Options:

1. **Skip the CLAUDE.md write for Claude Code** — the MCP server instructions already cover it. Other targets (Cursor `.mdc`, Kiro) may still benefit from their own instruction files since they don't surface MCP server instructions the same way.
2. **Make the CLAUDE.md write opt-in** — add a `--no-instructions` flag to `codegraph init` so users can skip it.
3. **Deduplicate the content** — have the CLAUDE.md block contain only a one-liner like "CodeGraph is configured — see MCP server instructions" instead of the full template.

Option 1 seems cleanest since Claude Code already surfaces `SERVER_INSTRUCTIONS` reliably.

colbymchenry:
Fixed in #538 (merged to `main`, ships in the next release).

[...]

PR #678, opened 2026-06-04
feat(mcp): surface supported languages in MCP server instructions

colbymchenry:
## What

The MCP `initialize` instructions — the first thing every agent sees about codegraph — listed **no languages**. So agents (and the users driving them) assumed languages like Rust weren't supported even though they're fully indexed, and skipped the codegraph tools entirely. That was the real gap behind #671 ("Rust support"), where Rust has in fact worked end-to-end for several releases (verified: structs/traits/`impl`/methods/calls, `callers`/`callees`/`impact`, trait `implements` edges, and Axum/Actix/Rocket route extraction + cargo-workspace resolution).

This adds one concise line to the opening paragraph of `server-instructions.ts`:

> It indexes 20+ languages — TypeScript/JavaScript, Python, Go, Rust, Java, C#, C/C++, PHP, Ruby, Swift, Kotlin, and more.

## Why here

`server-instructions.ts` is the single source of truth for agent-facing guidance (#529) and the first place the false "language X isn't supported" assumption forms. The README documents the full list, but it isn't reachable from *inside* the tool, which is where agents actually meet it.

## Testing

- `npm run build` — clean
- `npx vitest run __tests__/mcp-initialize.test.ts` — 3/3 pass (exercises the `initialize` response that carries these instructions)

refs #671


colbymchenry:
Superseded by the recut just merged — this branch predated the explore-first rewrite of server-instructions.ts and had content conflicts in both files. Same change, re-applied against the current instructions.

PR #688, opened 2026-06-04
feat: add GLSL support via tree-sitter-glsl (#685)

### LINKED ISSUE #685: Feature Request: Add support for GLSL
**Is your feature request related to a problem? Please describe.**
I work on a large C++ graphics codebase and `CodeGraph` has been fantastic for preventing Claude Code from thrashing through gigabytes of context. However, a significant portion of our logic lives in GLSL shaders. Currently, Claude falls back to raw `grep` and file reads when it hits the shader boundary, which burns through tokens and tool calls.

**Describe the solution you'd like**
I would love it if CodeGraph could index GLSL files alongside C/C++. Since CodeGraph relies on `tree-sitter`, the good news is that there are robust community grammars already available for GLSL.

**Technical Details for Implementation:**

**1. Target File Extensions:**
To properly support GLSL, the following standard extensions would need to be recognized:

* `.glsl` (Generic/Headers)
* `.vert` (Vertex shaders)
* `.frag` (Fragment shaders)
* `.comp` (Compute shaders)
* `.geom` (Geometry shaders)
* `.tesc` / `.tese` (Tessellation shaders)

**2. Available Tree-Sitter Grammars:**
There are currently two well-maintained grammars:

* **[Caellian/tree-sitter-glsl](https://github.com/Caellian/tree-sitter-glsl):** A newer, specification-compliant parser (GLSL 4.60). It does not inherit from C, which makes it much better at parsing real-world macros, uniforms, and `layout` qualifiers without generating bloated expression trees.
* **[tree-sitter-grammars/tree-sitter-glsl](https://github.com/tree-sitter-grammars/tree-sitter-glsl):** The traditional grammar, which is an extension of `tree-sitter-c`.

**Describe alternatives you've considered**
For now, Claude just uses its native file tools for `.glsl` files while relying on CodeGraph for the C++ side. This works, but indexing GLSL would allow for cross-referencing uniform names, buffer layouts, and shader functions directly through MCP.

Thanks for the amazing tool!

KadirKess:
## Summary

Add GLSL shader file indexing (`.glsl`, `.vert`, `.frag`, `.comp`, `.geom`, `.tesc`, `.tese`) — functions, structs, and intra-file calls are extracted so you can navigate and query shader code alongside the rest of the project.

Closes #685

## Changes

- **Vendor** `tree-sitter-glsl.wasm` (ABI-15) in `src/extraction/wasm/`
- **Add** `glslExtractor` with:
  - `resolveBody` — GLSL `function_definition` exposes `compound_statement` as a plain named child, not a `body` field
  - `visitNode` — intercepts `declaration` nodes containing nested `struct_specifier` so structs aren't swallowed by the variable extraction path
- **Register** all 7 shader extensions in the extension map, WASM grammar map, and display name map
- **Add** `'glsl'` to the `LANGUAGES` const array in `types.ts`
- **Add** extraction tests covering language detection, function extraction, struct extraction, and call extraction
- **Update** README supported languages table and summary line
- **Update** CHANGELOG

## Testing

All 5 new GLSL extraction tests pass:
- Language detection for all 7 shader extensions
- `isLanguageSupported('glsl')` and `getSupportedLanguages()` include `'glsl'`
- Function extraction (`void main()`, typed `vec3 computeNormal(...)`)
- Struct extraction (`struct Light { ... }`)
- Intra-file call extraction (`square()` → unresolved reference)

`tsc --noEmit` passes cleanly.


treytracedit-lab:
C++

PR #691, opened 2026-06-04
feat: SystemVerilog/Verilog language support (incl. UVM)

StopJoking97:
## Summary

Adds **SystemVerilog/Verilog** as an indexed language (`.sv` `.svh` `.v` `.vh`), backed by the [gmlarumbe/tree-sitter-systemverilog](https://github.com/gmlarumbe/tree-sitter-systemverilog) grammar (vendored as a prebuilt wasm, tree-sitter ABI 15). SystemVerilog is a strict superset of Verilog, so one grammar covers both — including UVM (which is "just" SystemVerilog classes + macros).

Goal: make an RTL/UVM codebase navigable with the same `callers` / `callees` / `impact` / graph tooling CodeGraph already gives software — the design hierarchy, the class/UVM-component tree, and transaction dataflow.

## What it indexes

- **Structure:** modules / programs / packages / interfaces / classes; typedefs (an enum typedef becomes an `enum` + members)
- **Design hierarchy:** module instantiation (emitted as `instantiates` **and** `calls`, so `callers`/`callees`/`impact` surface the RTL tree)
- **Inheritance:** class `extends` with package-qualified base resolution (`pkg::Base`)
- **Methods:** inline, `extern` out-of-class bodies (`Class::method`), and the `function new` constructor
- **Inheritance-aware calls:** `this.`/`super.` calls resolve up the `extends` chain — `super.build_phase()` binds to the **parent's** method, never a false self-edge
- **Composition (UVM has-a):** a typed-handle field → `references` edge, so `env → agent → driver/monitor` is queryable
- **UVM factory + TLM:** `type_id::create` (override-aware) composition, and `.connect()` TLM **dataflow** (e.g. `monitor → scoreboard`)

## Engine changes (additive, opt-in)

Most of the work is a self-contained extractor (`src/extraction/languages/systemverilog.ts`). The few shared-engine touch-points are no-ops for existing languages:

- **`LanguageExtractor.methodScopeKinds`** (new, optional): gates method-vs-function by enclosing scope. Languages that leave it unset keep their exact prior behavior.
- **name-matcher:** a small `extends`/`implements` kind-bias toward type-like targets (a base class should outrank a same-named function).
- **Two SystemVerilog-gated resolution post-passes** (`src/resolution/index.ts`): bind the inheritance-aware handle calls and the factory/TLM markers after the class graph exists. Gated on `language === 'systemverilog'` + a `__sv_*` marker prefix, so other languages never enter them.

## Tests

`__tests__/systemverilog.test.ts` — 31 cases covering extraction **and cross-file resolution** (the resolved edges, not just extraction): the inheritance chain-walk, composition, factory override, and the TLM dataflow.

## Notes

- The vendored grammar wasm is ~21 MB (the SystemVerilog grammar is large). It follows the existing `src/extraction/wasm/` vendoring pattern — happy to move it to git-lfs or a download-on-install step if you'd prefer.
- VHDL is **not** covered (different language family — would need its own grammar/extractor).

Feedback very welcome — glad to split this up, rework the engine touch-points, or adjust anything to fit the project's direction.

### DIFF
--- a/__tests__/systemverilog.test.ts
+++ b/__tests__/systemverilog.test.ts
@@ -1,0 +1,826 @@
+/**
+ * SystemVerilog / Verilog extraction tests.
+ *
+ * Locks in the design-hierarchy mapping (module/instantiation), subroutine-body
+ * call capture, enum-member emission, and package/import handling.
+ */
+import { describe, it, expect, beforeAll, afterEach } from 'vitest';
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+import { extractFromSource } from '../src/extraction';
+import { detectLanguage, initGrammars, loadAllGrammars } from '../src/extraction/grammars';
+import { matchReference } from '../src/resolution/name-matcher';
+import { CodeGraph } from '../src';
+import { DatabaseConnection } from '../src/db';
+import type { Node } from '../src/types';
+import type { ResolutionContext, UnresolvedRef } from '../src/resolution/types';
+
+beforeAll(async () => {
+  await initGrammars();
+  await loadAllGrammars();
+});
+
+describe('SystemVerilog language detection', () => {
+  it('maps the HDL extensions to systemverilog', () => {
+    expect(detectLanguage('rtl/fifo.sv')).toBe('systemverilog');
+    expect(detectLanguage('pkg/types.svh')).toBe('systemverilog');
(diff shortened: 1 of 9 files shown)


StopJoking97:
Closing this in favour of continuing the work outside the fork network — the branch has since been rebased onto v1.6.0 and extended (signal/port/parameter extraction, `ifdef handling, and drive edges). Thanks for the look.

PR #713, opened 2026-06-06
fix(daemon): keep a session alive when its daemon is restarted under it (#662)

### LINKED ISSUE #662: Bug: Daemon receives SIGTERM when new MCP session starts, orphaning existing session's proxy
### Environment
- **OS**: Linux (Ubuntu on WSL2)
- **CodeGraph**: v0.9.9 (npm package)
- **MCP Host**: OpenCode (oh-my-openagent)
- **Project**: ~500 Python files, ext4 filesystem on external USB drive

### Summary

When multiple OpenCode sessions are open against the same project, starting a **new** session causes the shared daemon to receive `SIGTERM`, shut down, and restart. The new session reconnects successfully, but the **previously-running session's proxy process dies and never reconnects**, leaving that session without CodeGraph.

### Reproduction

1. Open OpenCode session A (any terminal) → CodeGraph daemon starts, session A's proxy connects
2. Open OpenCode session B (different terminal) → daemon receives SIGTERM, restarts
3. Session B's proxy connects to restarted daemon ✅
4. Session A's proxy is dead (no reconnection logic) ❌ — CodeGraph tools unavailable in session A

### Evidence

**daemon.log** (30 daemon starts, 29 SIGTERM shutdowns, only 1 idle timeout):
```
[CodeGraph daemon] Listening on .../daemon.sock (pid 860038, v0.9.6). Idle timeout 300000ms.
... (normal operation ~50 lines) ...
[CodeGraph daemon] Shutting down (SIGTERM; clients=1).
[CodeGraph daemon] Listening on .../daemon.sock (pid 1024101, v0.9.6). Idle timeout 300000ms.
[CodeGraph daemon] Shutting down (SIGTERM; clients=2).
[CodeGraph daemon] Listening on .../daemon.sock (pid 1042182, v0.9.6). Idle timeout 300000ms.
... (repeats 29 times total) ...
```

Key observations from the log:
- **Only 1 out of 30 shutdowns was an idle timeout** (the intended lifecycle). The other 29 were all SIGTERM.
- Rapid restarts: SIGTERM → Listening in consecutive lines (no work between), indicating the daemon is killed and immediately restarted.
- Client counts at shutdown: `clients=1` (14 times), `clients=2` (12 times), `clients=3` (once), `clients=4` (once) — confirming multiple sessions were affected.

**Process tree** (when both sessions were running):
```
PID     PPID    PGID      SID       COMMAND
2792051 2791924 2791924   2773040   npm exec @colbymchenry/codegraph serve --mcp   ← session B proxy (pts/0)
2792074 2792067 2791924   2773040   MainThread (proxy)                              ← session B proxy
2792092 2792074 2792092   2792092   MainThread (daemon)                              ← detached daemon (own PGID+SID)
2793905 ...     ...       ...       npm exec @colbymchenry/codegraph serve --mcp   ← session A proxy (pts/1) [DEAD]
```

The daemon correctly has its own PGID (2792092) and SID (2792092), so it should be immune to process-group SIGTERM. Yet it still receives SIGTERM.

**After session B started**, session A's proxy processes (pts/1) disappeared entirely — `/proc/{pid}/fd/` returned 0 entries. The daemon was still running and healthy, but session A had no proxy to reach it.

**EPIPE error** observed once:
```
[CodeGraph daemon] socket error: write EPIPE
```
This occurred when the daemon tried to write to a disconnected proxy socket.

### Version history in log

```
v0.9.6 → v0.9.7 → v0.9.8 → v0.9.9
```
The problem persists across all versions.

### Analysis

The daemon architecture (from #411) is designed to support concurrent sessions:
- Detached daemon with own PGID/SID
- Thin proxy per MCP client connecting via Unix socket
- Reference counting + idle timeout

However, something sends SIGTERM to the daemon when a new session starts. Possible sources:
1. The MCP host (OpenCode) may send SIGTERM to tracked process trees
2. Race condition in lockfile arbitration during concurrent startup
3. The proxy in the old session dies when its parent (MCP host) cleans up, and somehow this propagates to the daemon

Once the daemon restarts, the new session's proxy reconnects, but the **old session's dead proxy has no reconnection mechanism** — it's gone entirely.

### Impact

- CodeGraph silently stops working in existing sessions when a new session starts
- No error message shown to the user in the affected session
- Requires restarting the affected OpenCode session to restore CodeGraph

### Suggested improvements

1. **Proxy reconnection**: The proxy should detect daemon socket loss and attempt reconnection rather than dying silently.
2. **SIGTERM source investigation**: Investigate why SIGTERM reaches a detached daemon with its own PGID/SID.
3. **Proxy health check**: The MCP host could periodically verify the proxy is alive and restart it if needed.

### Workaround

Setting `CODEGRAPH_NO_DAEMON=1` in the MCP server environment disables the shared daemon and runs each session independently (no daemon = no daemon conflicts, at the cost of per-session overhead).

colbymchenry:
## Problem (#662)

When multiple sessions run against the same project (reported on opencode / WSL2), starting a **new** session causes the shared daemon to receive `SIGTERM` and restart. The new session reconnects, but the **previously-running session's proxy dies and never recovers** — that session silently loses CodeGraph until it's restarted.

A reproduction confirmed the issue and surfaced a detail the report didn't: a `tools/call` **in flight** at the moment the daemon dies **hangs with no response** (the host waits on a reply that never comes), on top of the proxy exiting and all subsequent calls being lost.

The SIGTERM originates in the **host's** process-tree teardown, not in CodeGraph — nothing in the engine signals another process (verified). So the fix is **proxy resilience**, not chasing the signal source.

## Fix

The local-handshake proxy now treats a daemon disconnect as **recoverable, not terminal**:

- On daemon socket close/error (while the host is still attached), it falls back to its **in-process engine** for the rest of the session — the same path already used when no daemon is reachable at startup, and exactly what `CODEGRAPH_NO_DAEMON=1` does (the reporter's own workaround, now automatic).
- It tracks requests forwarded to the daemon and **re-serves any that were in flight** when the daemon died, so the host never hangs. Anything that can't be served locally is answered with an error rather than left hanging.
- The proxy **still exits when the host goes away** (stdin close / the #277 PPID watchdog) — only daemon loss is now non-fatal.

## Before → after (from the reproduction)

| | Before | After |
|---|---|---|
| Proxy after daemon SIGTERM | exits | **stays alive** |
| In-flight request | hangs, no response | **answered (in-process)** |
| Post-drop request | lost | **recovered** |

## Test plan

- **Reproduction** (proxy stays alive, in-flight request answered, post-drop request recovers).
- **Regression test** in `mcp-daemon.test.ts` — `proxy survives the daemon dying mid-session and keeps serving (#662)`: kills the daemon under a live proxy via SIGTERM, asserts the proxy is still alive and still answers a subsequent `tools/call`.
- macOS full suite green (1243 passed, 2 skipped); the daemon suite + the new test pass on a **Windows 11 VM** over named pipes.

### Also in this PR

Replaces the over-the-wire liveness-sweep test added in #712 — flaky under heavy parallel load (a raced raw-socket connect) — with a **deterministic `Daemon.reapDeadClients` unit test**. The client-hello round-trip is still exercised by every daemon test, since the real proxy now sends it.

Fixes #662.


treytracedit-lab:
Git config--C++/test

PR #729, opened 2026-06-08
fix(extraction): strip generic type arguments from supertype names so `extends Base<T>` resolves

paiyahooJay:
## Problem
Class inheritance edges were silently dropped whenever a superclass or implemented interface carried generic type arguments, e.g. `class A extends Base<T>` or `class B implements Iface<T>`.

Root cause: in the shared tree-sitter supertype extraction the extends/implements/C++-base paths used getNodeText() on the supertype node. For a generic supertype the node is `generic_type` (C++ `template_type`), so the captured reference name included the type arguments ("Base<T>"). During resolution this never matched the class node named "Base", and the reference was discarded — it was not even kept in unresolved_refs, so the loss was silent.

This affects every language that goes through the shared supertype extraction path: Java, TypeScript, and C++ templates. (C#, Kotlin, Swift, Rust and Go already narrow to the bare type identifier or handle `generic_type` explicitly, so they were unaffected.) The dropped edges land disproportionately on `*ServiceImpl extends BaseXxx<T>` interface+DI classes that matter most for call-graph reachability.

## Fix
Add a shared `stripTypeArguments` helper and use it when recording the extends/implements reference name, dropping the angle-bracket suffix (including nested generics) while preserving any qualified prefix so resolution can still disambiguate same-named types across packages.

## Test
Added regression tests asserting the extends/implements reference names are stripped for generic and nested-generic Java supertypes and C++ template base classes. Verified end-to-end that a generic Java project now produces the extends/implements edges that were previously dropped.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixes
+
+- Class inheritance edges are no longer silently dropped when the superclass or implemented interface carries generic type arguments — for example `class A extends Base<T>` or `class B implements Iface<T>` (including nested generics and qualified names). This affected Java, TypeScript, and C++ templates, and most often broke exactly the interface-to-implementation service classes that matter most for navigation, so `codegraph_callers`, `codegraph_impact`, and `codegraph_trace` now follow those inheritance links instead of dead-ending.
+
 
 ## [0.9.8] - 2026-06-01
 
--- a/__tests__/extraction.test.ts
+++ b/__tests__/extraction.test.ts
@@ -940,6 +940,47 @@ public class Splitter {
   });
 });
 
+describe('Generic supertype extraction', () => {
+  // Regression: a generic superclass/interface parses as a `generic_type`
+  // (C++ `template_type`) node whose text carries the angle-bracket suffix.
+  // Without stripping the type arguments the reference name (`Base<T>`) never
+  // matched the `Base` class node during resolution, so the extends/implements
+  // edge was silently dropped — it wasn't even kept in unresolved_refs.
+  it('strips generic type arguments from Java extends/implements supertypes', () => {
+    const code = `
+class Base<T> {}
+interface Iface<T> {}
+class A extends Base<String> {}
+class B implements Iface<String> {}
+class C extends Base<java.util.Map<String, Integer>> {}
+`;
+    const result = extractFromSource('Generics.java', code);
+
+    const extendsRefs = result.unresolvedReferences.filter((r) => r.referenceKind === 'extends');
+    const implementsRefs = result.unresolvedReferences.filter((r) => r.referenceKind === 'implements');
+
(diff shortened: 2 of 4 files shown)


Guimleh:
Independent corroboration, measured on a large private Java codebase (16,822 Java files,
18,969 indexed classes/interfaces) with codegraph 1.6.0:

- **1,409 `extends` + 759 `implements` edges are missing** — every single one a supertype
  carrying type arguments.
- **1,850 classes (9.8% of all indexed Java types)** lose at least one inheritance link.
- That is **+16.2% `extends`** and **+12.0% `implements`** on top of the 7,265 / 5,587
  currently resolved — roughly one Java inheritance edge in seven.
- No other language in this index shows a generic-supertype failure, though the codebase is
  overwhelmingly Java, so that is weak evidence for the others.

The practical failure mode is worse than the raw count suggests, because the loss concentrates
on exactly the base types one would query. Three real examples from that index:

| base type | actual subclasses | edges in graph |
|---|---|---|
| generic single-value-object base `Base<T, Self>` | 37 | **0** |
| generic value-object base `Base<Self>` | 14 | 1 — the one raw-typed subclass |
| framework UI base `Base<C, S>` | 24 | **0** |

"Who extends this?" and `codegraph_explore` impact / blast-radius come back **empty, with no
signal that anything was dropped** — a false negative rather than an error. On a DDD-style
codebase, where the value-object and aggregate hierarchies are generic by construction, that is
most of the type graph.

Two notes that I believe are all that stands between this PR and mergeable:

**1. The C++ hunk is superseded.** `stripCppTemplateArgs` landed with #1043 and `tree-sitter.ts`
already calls it at the `base_class_clause` site; the PR currently reports
`mergeable_state: dirty`. Only the two Java sites are still unstripped on `main`:

- `src/extraction/tree-sitter.ts:5471` — `extends`, `const name = getNodeText(target, this.source)`
- `src/extraction/tree-sitter.ts:5520` — `implements`, same shape

**2. The native kernel needs the same change, or the fix is inert and parity breaks.** Java was
routed to the Rust kernel by default in #1326, *after* this PR was opened (`DEFAULT_ROUTED` in
`src/extraction/kernel/index.ts`). `codegraph-kernel/src/java.rs::extract_inheritance` is a
faithful port of the buggy TS:

```rust
// codegraph-kernel/src/java.rs, ~1002 (extends) and ~1015 (implements)
let name = self.text(target).to_string();
self.push_ref_at(class_row, &name, extends_kind, target);
```

whereas the C++ kernel already does the right thing at the equivalent site:

```rust
// codegraph-kernel/src/ccpp/mod.rs:1615
let name = strip_cpp_template_args(self.text(t));
```

So on a default install the TS-only fix would change nothing for Java, and the two arms would
stop being byte-identical. Mirroring `strip_cpp_template_args`
(`codegraph-kernel/src/ccpp/mod.rs:196`) into those two `java.rs` lines looks like all that is
needed.

Minor correction to the description above: as of 1.6.0 these references *are* retained in
`unresolved_refs` with `status='failed'` — that is how the numbers here were obtained — so the
loss is recoverable from the DB, even though the edges themselves are absent.

Happy to help with a rebased two-arm patch if that is useful.

Issue #634, opened 2026-06-02
[TypeScript] String literal type arguments in generic tuple elements are not indexed as symbols


jiezhiyong:
When using `codegraph` to index a TypeScript codebase that defines service contracts dynamically using generic type tuples (e.g., passing string literals as type arguments to a generic helper type), the string literal names are not indexed as symbols. As a result, running `codegraph query <method_name>` returns empty results for these definitions, even though they represent the primary API methods utilized throughout the application.

---

### Minimal Reproducible Example (MRE)

Suppose we have a TypeScript file `src/services/api.ts` defining services as follows:

```typescript
// A generic Service interface
export interface Service<Name extends string, Req, Resp> {
  name: Name;
  request: Req;
  response: Resp;
}

// A tuple type mapping all available service endpoints
export type MyServiceList = [
  Service<
    'query_apply_record',
    { pageNo: number; pageSize: number },
    { success: boolean }
  >,
  Service<
    'apply_confirm',
    { code: string },
    { success: boolean }
  >
];

// A dynamic factory function creating the service client
declare function createService<T extends any[]>();: {
  [K in T[number] as K['name']]: (req: K['request']) => Promise<K['response']>;
};

export const api = createService<MyServiceList>();
```

In another file `src/index.ts`, we invoke the method:

```typescript
import { api } from "./services/api";
async function run() {
  // Property access is fully typed in TypeScript editor, but acts as a dynamic reference
  await api.query_apply_record({ pageNo: 1, pageSize: 10 });
}
```

---

### Steps to Reproduce

1. Initialize and index the project:
   ```bash
   codegraph init
   ```
2. Query the dynamically declared method name:
   ```bash
   codegraph query query_apply_record
   ```

### Expected Behavior

`codegraph` should index the string literal `'query_apply_record'` inside the `Service` generic type parameters, identifying it as a symbol (e.g., of kind `property`, `method`, or a synthetic node) so that developers and AI agents can trace where this contract is defined.

### Actual Behavior

`codegraph` returns no search results for `query_apply_record` within the service definition file, as it only indexes standard AST declaration nodes (like formal classes, interfaces, methods, or functions) and skips string literal arguments inside type aliases/tuples.

### Environment & Context

- **CodeGraph version:** v0.9.8
- **Pattern Significance:** This pattern is widely used in modern TypeScript micro-frameworks (e.g., custom RPCs, BFF contracts, mock servers) where types act as the source of truth for runtime dynamic proxy objects.

[...]

PR #730, opened 2026-06-08
test(extract): cover TypeScript generic tuple with string-literal type arguments

Kartalops:
Adds regression coverage for the dynamic-dispatch contract pattern from #634. When a TypeScript file declares a generic tuple whose element types carry string-literal type arguments (e.g. `Service<'name', ...>`), the literal names are not currently indexed by the extractor.

## Summary
- Adds one `.skip`-marked test in `__tests__/extraction.test.ts` inside the existing `TypeScript Extraction` describe block.
- Pins the failure mode: the literal names (`'query_apply_record'`, `'apply_confirm'`) are not in the extracted node list, and the containing tuple type alias `MyServiceList` is also not indexed.
- Test is `.skip` so the suite stays green. The next fix in `src/extraction/languages/typescript.ts` can be verified by removing the `.skip`.

## Testing
- `npx vitest run` (full suite) — 64 files passed, 1261 tests passed, 3 skipped (1 newly skipped, 2 pre-existing). No regressions.
- `npx vitest run __tests__/extraction.test.ts -t "generic tuple literal"` — 1 skipped (expected).
- Verification against `extractFromSource` with the new fixture: confirmed that neither `'query_apply_record'` nor `'apply_confirm'` appear in the node list, which is the behavior the test pins.

Refs #634.

### DIFF
--- a/__tests__/extraction.test.ts
+++ b/__tests__/extraction.test.ts
@@ -298,6 +298,53 @@ function main() {
     const calls = result.unresolvedReferences.filter((r) => r.referenceKind === 'calls');
     expect(calls.some((c) => c.referenceName === 'processData')).toBe(true);
   });
+
+  // Regression coverage for issue #634: string-literal type arguments
+  // inside generic tuple elements (the dynamic-dispatch contract pattern)
+  // are not currently indexed. The assertions below pin the failure mode
+  // so any future fix in src/extraction/languages/typescript.ts can be
+  // verified. The test is marked .skip to keep the suite green until the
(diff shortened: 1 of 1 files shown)


Kartalops:
Closing this PR as part of a routine cleanup. The PR has been open without maintainer engagement, and continuing to keep it active would consume reviewer bandwidth with no signal of acceptance.

If this PR is still relevant, please open a fresh PR with the latest `main` HEAD — and feel free to ping me with the new URL.

— submitted via OAuth-authenticated cleanup bot (hermes-pr-bot).

Issue #699, opened 2026-06-05
Feature Request: Support `.ignore` files to override `.gitignore` exclusions


keeferwu:
**Problem**

CodeGraph uses `git ls-files` under the hood to discover source files, which means it inherently respects `.gitignore`. Any directory or file matched by `.gitignore` is invisible to CodeGraph and cannot be indexed.

In many real-world projects, entire directories are gitignored even though they contain valuable source code that an AI agent needs to understand. Common examples:

- A top-level `customer/` directory that contains per-customer customization code
- `Pods/` in iOS projects (tracked locally but gitignored)
- `vendor/` in Go projects
- Embedded independent git repos inside a monorepo

Currently, the only workaround is to remove entries from `.gitignore` (which changes repo semantics) or to initialize a separate CodeGraph instance inside the ignored directory (which creates a fragmented index and requires passing `--projectPath` everywhere).

**Proposed Solution**

Support `.ignore` files (using the same pattern syntax as `.gitignore`) that act as a **local override layer** on top of `.gitignore`. This is a well-established pattern used by tools like `ripgrep`, `fd`, `ag`, and `rg`:

- `.gitignore` — patterns to ignore (unchanged, respected by CodeGraph as today)
- `.ignore` — patterns to override `.gitignore` with **negation patterns** (`!` prefix)

For example, if `.gitignore` contains:

```
customer/
```

And the user creates `.ignore` with:

```
!customer/
```

Then CodeGraph would **include** the `customer/` directory in its index even though it's gitignored.

**How it works (proposed)**

1. CodeGraph reads `.gitignore` as today to get the base exclusion set.
2. CodeGraph then reads `.ignore` files (if present) from the same locations where `.gitignore` is checked.
3. `.ignore` patterns are applied **after** `.gitignore` patterns, so negation patterns (`!xxx/`) in `.ignore` can re-include files/directories that `.gitignore` excluded.
4. `.ignore` is gitignored by default (added to tooling's own ignore list) so it doesn't affect repository semantics.

**Why this is better than alternatives**

| Approach | Downsides |
|----------|-----------|
| Remove entries from `.gitignore` | Changes repo behavior for all git commands; commits required |
| `git add -f` | Modifies the repo permanently; pollutes git tracking |
| Separate CodeGraph instances | Fragmented index; `--projectPath` needed everywhere; no cross-repo queries |
| `.ignore` file | Zero impact on git; local to each developer; well-known pattern from other tools |

**Prior art**

[...]

Issue #407, opened 2026-05-25
v0.9.4 indexes node_modules in non-git projects, causing noisy context results


0xWinner98:
## Summary

This may be related to / partly covered by #320, #360, and #405, but I wanted to add a concrete real-world reproduction from the published npm package (`@colbymchenry/codegraph@0.9.4`).

On a small JavaScript/MJS project, CodeGraph indexed `node_modules/` because the project `.gitignore` did not exclude it. The result was that `codegraph context` became dominated by third-party dependency symbols instead of the app code.

## Environment

- macOS
- Node: `v26.0.0`
- npm/npx: `11.12.1`
- CodeGraph: `0.9.4` via `npx --yes @colbymchenry/codegraph`
- Project type: private JavaScript/MJS project, not a git repository

## Commands run

```bash
npx --yes @colbymchenry/codegraph index -f /path/to/private-js-project
npx --yes @colbymchenry/codegraph status /path/to/private-js-project
npx --yes @colbymchenry/codegraph context -p /path/to/private-js-project \
  "Understand workspace/status/resume cards and job interim result flow" \
  --max-nodes 20 --max-code 4 --format markdown
```

## Observed result

Index succeeded, but most indexed files were dependencies:

```text
Indexed 3,217 files
status fileCount: 3,249
status nodeCount: 45,373
status edgeCount: 121,441
DB size: 81.26 MB
```

SQLite count from `files`:

```text
node_modules_files=3032
src_files=34
scripts_files=183
total_files=3249
```

The `context` query returned symbols from dependencies, for example `node_modules/@modelcontextprotocol/sdk/.../spec.types.d.ts`, instead of the project's app modules. This makes the MCP / AI-agent use case much less useful because the agent receives noisy dependency context.

## Expected / desired behavior

For AI-agent usage, dependency folders such as `node_modules/` should be easy to exclude without changing the project's git policy. A few possible fixes would work:

1. ship `--exclude` / `--include` support for `index` and `sync` (looks like #405 is already heading there),
2. support a CodeGraph-specific ignore file like `.codegraphignore` (similar to #360),
3. or document the current behavior very explicitly: after #285, `.gitignore` is the single source of truth, so users must add `node_modules/` there or wait for a CodeGraph-specific exclude mechanism.

## Why this matters

[...]

PR #782, opened 2026-06-11
fix(sync): apply ignore matcher to git change-detection fast path (#766)

### LINKED ISSUE #766: Incremental sync (git-status fast path) bypasses the ignore matcher — tracked files in excluded dirs leak back into the index
## Summary

`getGitVisibleFiles` filters the file list through `buildDefaultIgnore` (built-in defaults + root .gitignore) — that's the #407 design: committed dependency dirs are excluded even though they're tracked. But the incremental path doesn't apply the same matcher: `getChangedFiles`' git fast path consumes `git status --porcelain` output unfiltered, on the assumption that *"git status already omits .gitignored paths"*.

That assumption has two holes:

1. **The built-in default excludes aren't gitignore at all.** A *committed* `vendor/` or `node_modules/` dir is excluded from a full index by `DEFAULT_IGNORE_PATTERNS`, but git status happily reports changes to tracked files inside it.
2. **gitignore is a no-op for tracked files.** The documented way to exclude a committed dependency dir is a .gitignore entry (tracking-blind by design, per #407) — but `git status` still reports modifications to those tracked files, because ignoring only applies to untracked paths.

In both cases the changed file isn't in the DB (`!tracked`) → pushed to `added` → parsed and indexed. The index now disagrees with what a full `index --force` produces; the next force re-index silently drops the file again (flip-flop).

## Repro

1. Project with a **committed** `vendor/` dir (tracked files). Full `codegraph index --force` → vendor excluded ✓ (per #407).
2. Modify a tracked file inside `vendor/`.
3. `codegraph sync` → the file is hashed, found missing from the DB, and **indexed**.
4. `codegraph index --force` → it disappears again.

Same repro with a tracked dir listed in .gitignore instead of the built-in defaults.

(The FSEvents watcher path is NOT affected — `watcher.js` builds `this.ignoreMatcher` from `buildDefaultIgnore` and filters events. It's specifically the git-status fast path in `getChangedFiles`.)

## Suggested fix (verified locally)

Build the shared matcher in `getChangedFiles` and skip matched paths in the modified+added loop:

```js
const ig = buildDefaultIgnore(this.rootDir);
for (const filePath of [...gitChanges.modified, ...gitChanges.added]) {
    if (ig.ignores(filePath)) continue;
    // ... existing hash/compare logic
}
```

Deleted files can stay unfiltered — they only act when the path is already tracked in the DB, where removal is always correct (and it lets a newly-excluded dir's stale entries clean up).

We're running this patch locally on a ~1,100-file Swift + TS workspace and verified: a new file in an excluded tracked dir shows up in `git status` but does not enter the index via `codegraph sync`; full index and sync now agree.

## Relevance to #699

Any implementation of the proposed `.ignore` overlay inherits this leak unless the sync path consults the same matcher as enumeration — with an extra ignore layer that git knows nothing about, *every* exclusion (not just tracked-file edge cases) would leak back in through `git status`. Worth fixing together.

maxmilian:
## Summary

Fixes #766. The git fast path in change detection (`getChangedFiles`) consumed `git status --porcelain` output **unfiltered**, while file enumeration (`getGitVisibleFiles`) filters every path through `buildDefaultIgnore` (built-in default-ignore dirs + root `.gitignore`, per #407 — applied to tracked files too).

That asymmetry means a **tracked** file inside a committed dependency/build dir (`vendor/`, `node_modules/`, …) that the index deliberately excludes still shows up in `git status` when edited, isn't found in the DB (`!tracked`), and gets pushed into `added`. Because indexing correctly skips those dirs, the change is a **phantom pending add that no `codegraph sync` or `codegraph index` ever clears** — `status` and a full re-index disagree forever.

(Scope note: on `main` this path feeds `codegraph status`'s pending-change count; the `sync()`/`index` and FSEvents-watcher paths already filter, so this is a status/change-detection consistency fix, not index corruption. It also forecloses the leak for any future `.ignore` overlay — see #699.)

## Fix

Build the same `buildDefaultIgnore` matcher in `getChangedFiles` and skip matched paths in the modified+added loop. Deletions stay unfiltered — they only act on paths already tracked in the DB, where removal is always correct (and lets a now-excluded dir's stale entries clean up).

## Test

Added `__tests__/sync.test.ts › should not report changes to excluded tracked files (issue #766)`: commits a `vendor/lib.ts`, edits it (so `git status` flags it) alongside a real `src/index.ts` change, and asserts the excluded file does not appear in `getChangedFiles()` while the genuine source change still does. Red before the fix, green after.

Full suite green locally (`npm run build && npm test` → 65 files, 1349 passed | 2 skipped).

## Credit

Root cause and a verified patch were provided by @monochrome3694 in #766.


maxmilian:
Closing — #949 (commit `2010c2d2`) applied the ignore matcher to the git change-detection fast path and closed #766: the same fix as this PR, and a more complete one (it also keeps deletions unfiltered and judges each embedded repo by its own `ScopeIgnore` rules). Superseded. Thanks!

PR #783, opened 2026-06-11
fix(extraction): capture docstrings for export/decorator-wrapped symbols (#780)

### LINKED ISSUE #780: Preceding comments / docstrings are not captured for exported or decorated symbols
`getPrecedingDocstring` (in `src/extraction/tree-sitter-helpers.ts`) only finds a symbol's preceding comment when that comment is a direct previous sibling of the emitted symbol node. For symbols wrapped by another syntax node, such as export-wrapped declarations (`export function`, `export class`, `export const x = () => {}`) and decorated/annotated declarations, the comment is a sibling of the wrapper, not of the emitted node, so extraction returns `undefined` and the node's `docstring` is stored as `NULL`. 

## Detailed explanation

`getPrecedingDocstring` walks `previousNamedSibling` from the symbol node and collects adjacent comment nodes. That works when comment and symbol are siblings:

```ts
// loads the user        ← previousNamedSibling of `function getUser` ✓
function getUser() {}     ← docstring: "loads the user"
```

But tree-sitter nests an exported declaration under an `export_statement` (and a decorated one under a decorator/`decorated_definition`). CodeGraph emits the inner declaration as the symbol node, whose `previousNamedSibling` is empty. The comment is the previous sibling of the wrapper one level up:

```
export_statement
├── comment  "loads the user"     ← attached HERE (sibling of the wrapper)
└── function_declaration getUser  ← the node CodeGraph emits; its prev sibling is empty
```

I believe the issue is grammar-specific. It seems (from my tests) that Java / C# are not affected by it, but Python and JS are. 

## Steps to reproduce

The following Python and JS files should reproduce the missing docstrings:

```ts
// plain class control captured
class Ledger {}

// exported class dropped
export class Invoice {}

// export default dropped
export default function settle() { return true; }

// exported arrow const dropped
export const refund = (amount: number) => amount;

// non-export arrow const also dropped (wrapped by lexical_declaration even without export)
const audit = (amount: number) => amount;
```

And Python:

```python
# python decorator dropped
@app.route("/x")
def py_handler():
    return 1


# python plain control captured
def py_plain():
    return 1


# python decorated class dropped
@dataclass
class PyModel:
    pass
```

## Environment

- CodeGraph: `@colbymchenry/codegraph` 0.9.9
- Node: v24.16.0
- OS: Linux 6.17.4

maxmilian:
Fixes #780.

## Problem

`getPrecedingDocstring` (`src/extraction/tree-sitter-helpers.ts`) collects a symbol's doc comment by walking `previousNamedSibling` from the **emitted** symbol node. That works only when the comment and the symbol are siblings. For declarations tree-sitter nests inside a wrapper node, the emitted node is the inner declaration while the comment is a sibling of the wrapper one level up, so the scan finds nothing and `docstring` is stored as `NULL`:

- `export function` / `export class` / `export default …` → inner declaration under `export_statement`
- `export const x = () => …` and plain `const x = () => …` → arrow/declarator under `lexical_declaration`
- Python `@decorator`'d `def` / `class` → inner definition under `decorated_definition`

Plain (unwrapped) declarations, and inline-modifier languages like Java/C# (where `public`/`private` are inline children, not wrapper nodes), were unaffected — matching the reporter's observation.

## Fix

- Factor the sibling scan into `collectPrecedingComments`.
- In `getPrecedingDocstring`, **only when the node itself has no preceding comment**, retry from each wrapper ancestor, bounded to a whitelist of wrapper node types (`export_statement`, `decorated_definition`, `lexical_declaration`, `variable_declaration`, `variable_declarator`, `decorator`) so the walk can't reach the file root and mis-attribute an unrelated leading comment.

Because the fallback runs only when the current behavior returned nothing, unwrapped declarations and all existing tests are byte-for-byte unchanged (zero regression by construction).

## Tests

Two regression tests in `__tests__/extraction.test.ts` using the reporter's repro — TS/JS (export class, export default function, exported arrow const, plain arrow const) and Python (decorated def, decorated class), each with an unwrapped control symbol that already worked. Red before the fix, green after.

`vitest run __tests__/extraction.test.ts` → 349 passed (incl. the 2 new). `tsc --noEmit` clean. The repo's pre-existing macOS daemon/MCP subprocess test failures (`mcp-daemon`, `mcp-initialize`, `mcp-roots`, `status-json`) are unrelated and identical on a pristine `main` checkout.

Re-index (`codegraph index -f`) to backfill docstrings on existing indexes.


maxmilian:
Closing — superseded upstream by `0df9246` (PR #806, for #780), which captures & cleans docstrings across all README languages (a superset of this PR's wrapper-climb fix, adding `cleanCommentMarkers` for more languages). Thanks!

PR #15, opened 2026-02-05
Add root management and autonomous indexing tools to MCP server

rickross:
Hi Colby!

We've started using CodeGraph extensively and find it incredibly valuable. We added several MCP tools to make it easier for managing multiple projects and for supporting AI agents handling project lifecycle issues (init/index/sync). We hope these will be helpful contributions.

### 1. Root Management Tools
- **codegraph_get_root** - Get currently active root path
- **codegraph_set_root** - Switch between indexed projects
  - Now returns status immediately (files, nodes, edges, DB size) - no separate status call needed
  - Better error messages for uninitialized roots

### 2. Project Lifecycle Tools
AI assistants can now manage CodeGraph project lifecycle on behalf of users:
- **codegraph_init** - Initialize CodeGraph in current root
- **codegraph_index** - Full index of current root
- **codegraph_sync** - Incremental sync of current root
- **codegraph_uninit** - Remove CodeGraph from current root (cleanup)

These operate on the "current root" set by `set_root`, providing a clean context-based API.

### 3. Performance Optimization (80x speedup for reference resolution)
Optimized reference resolution with symbol caching:
- **Before:** 80+ seconds for sync on projects with 20K+ unresolved refs (9% CPU)
- **After:** ~1 second (95% CPU utilization)
- **Cause:** Repeated SQLite queries for symbol lookups
- **Fix:** Pre-load all symbols into memory maps indexed by name/qualified name

Tested on 681-file codebase with 26,233 unresolved references.

### 4. Tree-sitter Fixes
Merged @jasques' PR #9 fixes for tree-sitter installation issues (with full credit).

## Use Cases

**Multi-project workflows:**
```typescript
codegraph_set_root("/path/to/project-a")  // Shows: 500 files, 2000 nodes...
codegraph_search("AuthService")

codegraph_set_root("/path/to/project-b")  // Shows: 800 files, 3500 nodes...
codegraph_search("AuthService")
```

**Autonomous project lifecycle:**
```typescript
codegraph_set_root("/path/to/new-project")  // Detects not initialized
codegraph_init()                              // Initialize it
codegraph_index()                             // Index it (now 80x faster!)
codegraph_search("main")                     // Ready to use!
// ... later ...
codegraph_uninit()                            // Clean up when done
```

## Testing
Verified across 3 different projects (TypeScript, Python, mixed) with successful initialization, indexing, syncing, and querying.

Let us know if you'd like any changes or have questions about the design!

PR #827, opened 2026-06-12
spec(codegraph-rs): external-package references design

gyorgybalazsi:
## Summary
Design spec for resolving references into **external dependency packages** — the Daml SDK stdlib/prelude/script today (the 6,669 `no_candidate` calls + 20 unresolved `DA.*`/`Prelude`/`Daml.Script` imports left after PR #15), and any third-party DAR later. External symbols are ingested as **resolution-only targets**: marked `external`, excluded from embeddings and search, but valid `CALLS`/`TYPE_OF`/`REFERENCES`/`IMPLEMENTS` targets.

Adds `docs/superpowers/specs/2026-06-12-codegraph-rs-external-packages-design.md`.

## Locked decisions
- First-class **`Package`** layer + **package-qualified** resolution.
- **General** mechanism driven by compiled artifacts (works for SDK stdlib and any DAR).
- Ingestion **vendors** the proven prost-based Daml-LF decoder from the sibling `canton-toolbox` `codegen` crate (no from-scratch LF reader).
- Dependency mapping from `daml.yaml` + DAR manifests; SDK packages are implicit deps.
- External symbols resolution-only (excluded from embed/search); `IMPLEMENTS`/`REQUIRES` may target external interfaces.

## Reviewed against source (codegraph-rs + canton-toolbox)
The spec was plan-verified and corrected before commit:
- **No `schema_version` bump** — it's a strict-equality config gate (`config/workspace.rs:95`); bumping would break every existing workspace. The new `Package(qid)` constraint is additive/idempotent.
- **LF names are interned** — module/def names are indices into the package string/dotted-name tables; the walk must dereference them (reuse `canton-toolbox/.../resolve_type.rs`).
- **Pin one LF version** — the SDK ships stdlib/prim for multiple LF versions with identical module names; ingesting >1 breaks module lookup.
- **Dup-tolerant module lookup** — `find_module_by_qualified_name` currently bails on >1; package-scoped lookup must return the in-scope match instead.
- **Local packages** keyed synthetically (`pkg:local:<name>`) since they have no compiled `package_id`.

## Phasing (slices)
11a model + visibility (additive, no behavior change) → 11b vendor LF decoder + DAR/DALF ingestion → 11c package-qualified resolution → 11d corpus validation.

Implementation has not started. Sign-off requested on the design.

### DIFF
--- a/docs/superpowers/plans/2026-04-22-codegraph-rs-slice-1-walking-skeleton.md
+++ b/docs/superpowers/plans/2026-04-22-codegraph-rs-slice-1-walking-skeleton.md
@@ -543,6 +543,8 @@ cargo test --test config_parsing 2>&1 | tail -10
 
 Expected: compile errors about `parse` and `WorkspaceConfig` being missing from `codegraph_rs::config::workspace`.
 
+> **Note on intermediate broken builds:** The project will not build between steps 4 and 6 of this task — `workspace.rs` imports `crate::config::naming` which is created in step 6. This is intentional. Don't run `cargo build` until step 7.
+
 - [ ] **Step 4: Implement the config types and parser**
 
 Create `src/config/workspace.rs`:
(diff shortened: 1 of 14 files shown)


gyorgybalazsi:
Closing — opened against the wrong base by mistake (meant for my own fork gyorgybalazsi/codegraph, not upstream). Apologies for the noise.

PR #830, opened 2026-06-12
feat: add Erlang language support (.erl, .hrl)

malou996:
## Summary

- Adds **Erlang** as a supported language with full extraction and cross-module resolution, benchmarked on poolboy, cowboy, and EMQX.
- Vendored the WhatsApp `tree-sitter-erlang` WASM (ABI 14) — the `tree-sitter-wasms` build doesn't ship one.

### Extraction (5 commits, layered)

| Commit | What |
|---|---|
| `aec1357` | Core wiring: grammar WASM, extractor, extension detection, corpus entries, README, CHANGELOG |
| `6ae8303` | Remote call resolution — `module:function()` calls resolve across `.erl` files without imports |
| `3a84653` | `-behaviour(Name).` declarations emit `implements` edges; OTP stdlib behaviours gracefully resolve to null |
| `3d186f1` | `.app.src` / `.app` OTP application metadata — application name as module node, `applications` list as import nodes |
| `3268159` | `.hrl` include resolution — `-include`/`-include_lib` connect to header files via `file→file` imports edges |

### What's extracted

- **Modules** (`-module(Name).`), **records** (`-record(Name, {...}).`) with fields
- **Functions** (multi-clause `fun_decl`), with name/arity signatures (`handle_msg/2`)
- **Type declarations** (`-type`, `-opaque`)
- **Macros** (`-define(NAME, ...).` → `constant` nodes)
- **Exports** (`-export([...]).`)
- **Remote calls** (`gen_server:start_link(...)`) → cross-module `calls` edges
- **Behaviours** (`-behaviour(gen_server).`) → `implements` edges + callback synthesis
- **Includes** (`-include`, `-include_lib`) → `imports` edges to `.hrl` files
- **OTP app metadata** (`.app.src`) → module + dependency import nodes

## Test plan

- [x] 11 Erlang extraction tests pass (`npx vitest run __tests__/extraction.test.ts -t "Erlang"`)
- [x] Full suite green (105 resolution tests, 374 extraction tests)
- [x] Verified on poolboy: `poolboy.app.src` → module `poolboy` + imports `kernel`/`stdlib`
- [x] Verified end-to-end `.hrl` resolution: `my_server.erl` → `include/records.hrl` imports edge
- [x] Verified behaviour resolution: no false `implements` edges (module-only matching, name-matcher bypass)


malou996:
Bug fix: module:function syntax in queries
The resolver's hasAnyPossibleMatch pre-filter handled ., ::, and / separators but missed Erlang's single-colon :. This caused all mdb:dirty_insert-style references to be silently dropped before resolveErlangRemoteCall could run. The same issue affected the CLI callers/callees/impact commands and the MCP findAllSymbols lookup — all three now strip the module prefix so agents can query with the natural module:function format.

PR #833, opened 2026-06-12
fix(resolution): resolve cross-file static method calls to the method, not the class (#825)

### LINKED ISSUE #825: Cross-file static method calls (ClassName.staticMethod()) are dropped: calls edge mis-promoted to instantiates
## Summary

When a function calls a **static method** via `ClassName.staticMethod()` across files, CodeGraph does **not** create a `calls` edge to the method. Instead, the edge is mis-resolved into a `ClassName instantiates`-style edge pointing at the **class**, and the target method is dropped entirely.

As a result, `callers`/`impact` for the static method return nothing (false negatives), and the CLI's fallback bare-name search introduces unrelated same-name hits (false positives). Both failure modes happen at once.

Confirmed still present on `main` (commit `7db4c1d2`), not just on the released `0.9.9`.

## Minimal reproduction

```ts
// helpers.ts
export class Foo {
  static bar(x: number) { return x + 1; }
}

// caller.ts
import { Foo } from './helpers';
export function run() { return Foo.bar(41); }
```

Index this and run:

```bash
codegraph callers "Foo.bar" --json   # => empty
codegraph callers "bar" --json        # => empty (or unrelated same-name hits)
```

Expected: `run` is reported as a caller of `Foo.bar`.
Actual: no caller is found for the method.

## Root cause (confirmed by querying the SQLite edges directly)

When resolving the call `Foo.bar()`:

- The call's receiver is `Foo`, which is a **class name**.
- In `resolveOne`, there is logic that promotes a `calls` edge to `instantiates` when the `calls` target resolves to a class.
- Because the static-call receiver `Foo` matches the class first, the whole edge gets re-classified as `run instantiates Foo`, pointing at the **class `Foo`** instead of the method `Foo::bar`.
- The intended `run --calls--> Foo::bar` edge is never created; method `bar` is discarded.

A direct SQLite check confirms there are **zero `calls` edges** pointing at the static method. So `callers`/`impact` cannot find them, and the CLI's bare-name fallback then surfaces unrelated symbols with the same name.

This does **not** appear to be by design: `matchMethodCall` looks intended to support static calls, but the "calls → instantiates promotion" logic fires first and intercepts it. It reads as a logic-ordering conflict / bug rather than a deliberate limitation.

## Impact

- `callers <staticMethod>` → false negatives (real callers missing).
- `impact <staticMethod>` → blast radius severely under-reported.
- CLI bare-name fallback → false positives (unrelated same-name symbols).

This is significant because `ClassName.staticMethod()` is a very common pattern in TypeScript (static utility/helper classes).

## Environment

- Released version: `0.9.9` (npm)
- Also reproduced from source on `main` @ commit `7db4c1d2`
- Languages: TypeScript

## Suggested direction

Only promote a `calls` edge to `instantiates` for actual construction (e.g. `new ClassName(...)`), not when the class name is merely the receiver of a static method call. When the receiver resolves to a class **and** a matching static method exists on that class, resolve the edge to the method (`ClassName::method`) instead of to the class.

## Notes

A separate, lower-priority issue was also observed: `affected` only accepts bare relative paths — a `./` prefix or absolute path silently returns 0 (path not normalized). Happy to file that separately if useful.

maxmilian:
Fixes #825.

## Problem

A cross-file static method call resolves to the **class**, not the method, and is then recorded as a construction:

```ts
// helpers.ts
export class Foo { static bar(x: number) { return x + 1; } }
// caller.ts
import { Foo } from './helpers';
export function run() { return Foo.bar(41); }
```

`codegraph callers "Foo.bar"` → empty. `run` ends up with a single `instantiates → Foo` edge and **no** `calls → Foo::bar`, so callers/impact for the static method are empty (false negatives) and the CLI's bare-name fallback surfaces unrelated same-named symbols (false positives).

## Root cause

In `resolveViaImport` (`src/resolution/import-resolver.ts`), the generic named-import loop matches `ref.referenceName.startsWith(imp.localName + '.')` for the ref `Foo.bar`. For a non-namespace import `memberName` is left `null`, so `findExportedSymbol` returns the class `Foo`. `resolveViaImport` returns that at confidence `0.9`, and `resolveOne` returns immediately — so the name-matcher's `matchMethodCall` (which already resolves `Foo.bar` → `Foo::bar` via its class-name strategy) never runs. `createEdges` then promotes the `calls` edge to `instantiates` because the target is a class, dropping the method.

So the `calls → instantiates` promotion is the *relabel*; the actual interception is the import resolver picking the class as the target. Guarding the promotion alone would leave the edge pointing at the class with the method still dropped — hence the fix is in the import resolver.

## Fix

When the imported receiver is a `class`/`struct` and the `calls` ref is shaped `localName.member`, resolve the trailing member to that class's **method** (keyed on the exact owner segment of the qualifiedName, so a substring-named sibling like `FooBar::bar` can't soak the call) and return it — preserving the exact-file precision the import gives over a name-match fallback. If no matching method exists, behavior is unchanged. Genuine `new Foo()` construction is unaffected (still `instantiates`).

## Tests

Added a regression test in `__tests__/resolution.test.ts` that locks:
- `run --calls--> Foo::bar` (exactly one edge), not mis-promoted to `instantiates`
- the call is **not** soaked into the substring-named sibling `FooBar::bar`
- `getCallers(Foo::bar)` surfaces `run`
- the `new Foo()` boundary still instantiates the class

Full suite green locally (`npm run build && npm test`): 1429 passed, 2 skipped.


maxmilian:
Closing as redundant — #865 (`f7441f2`) already resolves #825 with the same approach: `resolveViaImport` now descends an imported class to its static member via `resolveStaticMember`, so a `Foo.bar()` call after `import { Foo }` links to the method instead of being mis-promoted to an `instantiates` edge on the class. #865 is a superset (it also covers affected-path normalization in the CLI), so there's nothing left for this PR to add. Thanks @colbymchenry!

PR #846, opened 2026-06-13
fix(react): index forwardRef/memo components in .tsx/.jsx files (#841)

### LINKED ISSUE #841: forwardRef components get no JSX render edges — callers/impact silently return empty
**Version:** 0.9.9 (linux x64, node:sqlite backend)
**Language:** TypeScript/TSX (Turborepo monorepo, ~490 indexed files)

## Summary

Components declared as `const X = React.forwardRef(...)` are indexed as
`constant` nodes, and **no render/call edges are created for their JSX
usages** (`<X ...>`). As a result `codegraph_callers` / `codegraph_impact`
return "No callers found" for any shadcn/ui-style component, even when
dozens of files render it.

Function-declared components are fine: for `export function ThemeProvider()`
the JSX render edges exist and callers are reported correctly. The problem
is specific to the declaration style.

## Repro

```tsx
// packages/ui/src/button.tsx
import * as React from "react";
export const Button = React.forwardRef<
  HTMLButtonElement,
  React.ButtonHTMLAttributes<HTMLButtonElement>
>((props, ref) => <button ref={ref} {...props} />);
Button.displayName = "Button";

// apps/web/page.tsx
import { Button } from "../../packages/ui/src/button";
export function Page() {
  return <Button>Click</Button>;
}
codegraph init -i
codegraph callers Button
Expected: Page (function) — apps/web/page.tsx
Actual: No callers found for "Button"
```
### Also observed:

codegraph_search with kind: "component" does not return Button at all (it only exists as a constant node).
The import { Button } ... edges ARE indexed, so the data loss is only in the render/call edge extraction, not in import tracking.
Real-world impact
In our monorepo <Button> is rendered in 81 files across 3 apps
(verified with grep), yet codegraph_callers Button and
codegraph_impact Button report nothing. This is a dangerous silent false
negative: "no callers" reads as "safe to change" right before a refactor of
a design-system component. Every shadcn/ui component uses the
const X = React.forwardRef(...) pattern, so for typical React design
systems the entire UI layer is invisible to impact analysis.

Likely related (untested): React.memo(...), styled(...) and other
HOC-wrapped const components.

### Suggested fix
Treat const X = React.forwardRef(...) / React.memo(...) initializers as
component declarations in the tree-sitter extraction queries (kind:
component), and emit the same JSX render edges that function-declared
components get.

**For comparison:** function-declared components in the same codebase resolve
correctly through barrel re-exports and workspace aliases (@repo/ui), so
the resolution layer itself looks solid — it's only the component
recognition for wrapped declarations that's missing.

maxmilian:
Closes #841.

## Problem

Components declared as `const X = React.forwardRef(...)` (or `memo`, or arrow functions) in `.tsx` / `.jsx` files get no render/call edges: `codegraph callers` / `codegraph impact` return "No callers found" even when many files render them, and they don't show up under a `component` search — they exist only as `constant` nodes. Function-declared components (`export function ThemeProvider()`) work fine.

## Root cause

The React framework resolver (`reactResolver`) already has regex-based detection for forwardRef/memo/arrow components, but its applicable-languages list was `['javascript', 'typescript']`. `.tsx` / `.jsx` files are detected as their own languages (`tsx` / `jsx`), so `getApplicableFrameworks()` filtered the resolver **out of the extraction pass** for exactly those files — its `extract()` never ran and no `component` node was synthesized.

At resolution time the `<X>` render reference reaches `resolveComponent()`, but the only node named `X` is the tree-sitter `constant` node, which `COMPONENT_KINDS = {component, function, class}` filters out → `null` → no edge. Function components work because tree-sitter emits a `function` node, which *is* in `COMPONENT_KINDS`.

## Fix

Add `tsx` / `jsx` to the resolver's `languages` so the existing component-detection path runs for those files. One-line change (plus a comment).

## Testing

- New end-to-end regression test in `__tests__/frameworks-integration.test.ts`: indexes a `forwardRef` `Button` (.tsx) rendered by a `Page`, asserts `Button` is a `component` node and that `Page` is reported as a caller. Fails before the change, passes after.
- `tsc --noEmit` clean.
- Full suite: the only newly-passing test is the one added here; the pre-existing failures are all in the MCP-daemon / subprocess-spawning integration tests (unrelated to extraction) and fail identically on a clean checkout in this environment.


colbymchenry:
Thanks so much for this, @maxmilian — your root-cause is exactly right, and the writeup (the `getApplicableFrameworks` language filter dropping the resolver from the *extract* pass for `.tsx`/`.jsx`, plus the `COMPONENT_KINDS` asymmetry that lets `function` components work but not `forwardRef`/`memo`) is what made this quick to close out.

We ended up landing a slightly deeper fix, so I'm going to close this PR in its favor — but your diagnosis carried it, so I want to explain the reasoning:

- It moves the forwardRef/memo recognition into the **tree-sitter extractor** rather than the resolver regex, so the component becomes a single `component` node with a real body range and its callees captured. Crucially, that avoids the duplicate `function` + `component` (and duplicate hook) nodes the resolver path produces on `.tsx`/`.jsx` — the very thing you flagged and added a guard test for. We took that as a signal to fix it at the source instead of guarding around it.
- On top of that it picks up `styled-components`/`emotion` (`const X = styled.button` … `), and it uses your `tsx`/`jsx` `languages` insight to properly restore **React Router / Next.js route indexing** on JSX files (same root cause), with the now-redundant component/hook regex removed so nothing double-counts.

Your `tsx`/`jsx` change is part of the landed fix, and you're credited for the root-cause in the release notes. Genuinely appreciate the careful diagnosis here — please keep them coming. 🙏

(Landing in the next release.)

Issue #776, opened 2026-06-10
Ideas: shrink the MCP surface — parameterized relations tool, MCP resources, allowlist presets


sebyx07:
## Context

I run Claude Code across 100+ projects and maintain a field-notes repo on configuring it ([claude-code-bible](https://github.com/sebyx07/claude-code-bible) — the MCP patterns below are from [ch. 7](https://github.com/sebyx07/claude-code-bible/blob/main/docs/07-memory-and-mcp.md)). CodeGraph is attached to several of them, including a 972-file / 10.3k-node / 22k-edge bun monorepo with code spread across `apps/*`, `packages/*`, `scripts/` — coverage there is excellent (I diffed the index against disk: only `dist/` and `.vitepress` cache excluded, all real source indexed).

Current standing cost per session, measured on v0.9.9: 8 tools ≈ 8.0KB of `tools/list` schemas + 7.3KB server instructions ≈ **~3.8k tokens in context every request**, for every connected agent. #775 shaves the easy ~220 tokens non-breakingly; the ideas below go further but are breaking or design decisions, so they're yours to pick from rather than drive-by PRs.

## 1. Merge `callers` / `callees` / `impact` into one parameterized tool

All three are the same operation — traverse edges from a symbol — differing only in direction:

```
codegraph_relations({ symbol, direction: "callers" | "callees" | "impact", depth?, limit? })
```

8 → 6 tools. Models dispatch on a parameter trivially (it's the same decision as picking among three tools, minus three schemas in context), and the server instructions get one bullet instead of three. The general principle: every per-verb tool pays schema rent every turn; parameters don't.

This also absorbs future tool-count pressure: #728 (`codegraph_file_symbols`) is already expressible as `codegraph_node({ file, symbolsOnly: true })`, and #769 (list projects) could be a `direction`-style mode or a resource (below) instead of a 9th tool.

Migration: keep the old three names as aliases in the `execute()` switch for a minor version (log a deprecation), and `CODEGRAPH_MCP_TOOLS` already lets anyone pin the old surface.

## 2. Implement MCP resources; move read-only diagnostics there

`proxy.ts` currently answers `resources/list` with a hardcoded `[]`. Two natural first resources:

- `codegraph://status` — index health. The tool's own description says "Skip unless debugging", yet its schema costs every turn. A resource is exactly the right shape for read-only, rarely-needed diagnostics.
- `codegraph://files/{path}` (resource template) — the indexed file tree.

[...]

PR #847, opened 2026-06-13
feat(mcp): add CODEGRAPH_MCP_TOOLS presets

jsiu93:
## Summary

- Add a shared `CODEGRAPH_MCP_TOOLS` parser so static `tools/list` and live `ToolHandler` use the same allowlist behavior.
- Add presets aligned with the current single-tool MCP default: `default` -> `explore`, `compact` -> `explore,node,search`, `core` -> `explore,node,search,callers`, and `full` -> every defined MCP tool.
- Document the presets in the README and MCP server reference while keeping comma-separated custom tool lists working.

Partially addresses #776.

## Testing

- `npx vitest run __tests__/mcp-tool-allowlist.test.ts`
- `npm exec tsc -- --noEmit`

## Notes

Rebased on current `main` after the default MCP surface moved to `codegraph_explore` only. The patch keeps that default intact; additional tools appear only when `CODEGRAPH_MCP_TOOLS` is set.

Tool descriptions stay untouched, so #775 remains a separate description-only PR.


LuuOW (commented):
Technical audit: Architecture and implementation patterns verified for consistency with MCP ecosystem standards.

PR #228, opened 2026-05-20
add support for elixir

oberernst:
What it says on the tin: a pass at elixir support in codegraph.

PR #871, opened 2026-06-14
feat(extraction): Elixir language support (.ex/.exs)

allenwoods:
## What

Adds **Elixir** (`.ex` / `.exs`) to CodeGraph's tree-sitter extraction. The `tree-sitter-elixir.wasm` grammar already ships in `tree-sitter-wasms`, so this wires it into `WASM_GRAMMAR_FILES` / `EXTENSION_MAP` and adds an extractor.

## Why a custom approach

tree-sitter-elixir is metaprogramming-first — there are almost no dedicated declaration node types. `defmodule`, `def`, `defp`, `alias`, `import`, `if`, `case` … all parse as the same shape:

```
(call target: (identifier "<macro>") (arguments …) (do_block …)?)
```

So extraction runs through the `visitNode` hook and dispatches on the macro identifier instead of node types (similar in spirit to Ruby's module handling and Pascal's custom visitor). `callTypes` is empty because the core's `extractCall` keys off a `function` field Elixir lacks — the hook records calls itself.

## Extracted

- **Modules** (`defmodule`, nested) and **protocols** (`defprotocol` → `interface`)
- **Functions** — `def` / `defp` / `defmacro` / `defmacrop` / `defguard` / `defguardp` / `defdelegate`, with public/private visibility. Multi-clause definitions fold into a single symbol (same `qualifiedName`) so the resolver isn't left with duplicate nodes.
- **Dependencies** — `alias` / `import` / `require` / `use`, including multi-alias `alias A.{B, C}` expansion into one import each
- **`defimpl`** with an `implements` edge to the protocol
- **`defstruct` / `defexception`** as struct nodes
- **Call edges** — qualified `Mod.fun` and local calls, descending through control-flow special forms (`if` / `case` / `with` / `for` / pipes) while not recording those forms themselves as calls. Module attributes (`@doc` / `@spec` / …) are skipped in this pass.

## Files

- `src/extraction/languages/elixir.ts` — new extractor
- `src/extraction/grammars.ts` — grammar + extension registration
- `src/extraction/languages/index.ts` — register in `EXTRACTORS`
- `src/types.ts` — add `elixir` to `Language`
- `__tests__/extraction.test.ts` — 16 tests
- `README.md` / `CHANGELOG.md`

## Testing

- 16 new unit tests: detection, modules (nested), functions + visibility, multi-clause folding, guards, `do:` shorthand, imports (incl. multi-alias), protocol/impl/struct/delegate, call edges
- `tsc --noEmit` clean
- Full suite: **1506 passed / 2 skipped**, no regressions
- Smoke-tested against real Phoenix/OTP source (1900+ LOC modules): correct module names, public/private visibility, and hundreds of call edges

## Notes

- The grammar binary is reused from `tree-sitter-wasms` — no vendored `.wasm` added (so it's not in the `__dirname/wasm` special-case list in `loadGrammarsForLanguages`).
- I did not add an eval-corpus benchmark entry; happy to add one if you'd like the measured-coverage table updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### New Features
+
+- **CodeGraph now indexes Elixir** (`.ex` / `.exs`) — modules (including nested), `def`/`defp`/`defmacro`/`defmacrop`/`defguard`/`defdelegate` with public/private visibility, multi-clause functions folded into one symbol, `alias`/`import`/`require`/`use` dependencies (including multi-alias `alias A.{B, C}` expansion), `defprotocol`/`defimpl` (with `implements` edges), `defstruct`/`defexception`, and call edges (qualified `Mod.fun` and local calls, including inside `if`/`case`/`with`/pipe bodies). Because tree-sitter-elixir parses every construct as a generic `call`, extraction dispatches on the macro identifier rather than node types; Phoenix/OTP codebases get the full explore / impact / callers surface.
+
 
 ## [1.0.0] - 2026-06-12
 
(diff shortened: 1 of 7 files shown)


kotsutsumi:
+1 👌

yld:
What about #228 ?

kotsutsumi:
> What about #228 ?

Oh, that's great! I see this implementation was already proposed.

jbonet:
So, is this going to get approved at some point?

Issue #441, opened 2026-05-26
please add plc scl lag


yexiang1994:
Please help me,  I noticed that the project doesn't include the SCL language. I'm a PLC developer and would greatly appreciate it if SCL support could be added.  thank you.

colbymchenry:
Folding this into a consolidated language-roadmap tracker so requests don't fragment ahead of 1.0: **#648** — your request is listed there with a link back to this issue.

codegraph is tree-sitter based, so this is one of the easier areas to contribute a PR — see the `/add-lang` workflow and the cookbook tracked in #97. 👍 on #648 to signal demand, and PRs are very welcome.

Closing here in favor of the tracker (not a no — just consolidating). Thanks for the request!

PR #875, opened 2026-06-14
feat: add CUDA (.cu/.cuh) and IEC 61131-3 SCL (.scl/.st) language support

### LINKED ISSUE #387: Title: Feature request: index .cu / .cuh CUDA source files
Hello! 
First of all, thank you for CodeGraph — it has worked great on a C++ project I've been exploring.

I noticed that C++ is fully supported (.cpp, .hpp, .cc in the Supported Languages table), but CUDA source files (.cu and .cuh) don't appear to be indexed at all. They aren't listed in the table, and CodeGraph seems to skip them during indexing.

Many GPU-accelerated and ML/inference projects keep a significant portion of their logic — kernels, device functions, and host-side launch code — in .cu files. Since these files are skipped, agents using CodeGraph get an incomplete picture of the project, missing call relationships between host code and CUDA kernels, as well as imports and helper functions defined in .cu / .cuh files.

Since CUDA syntax is largely a superset of C++, the existing C++ tree-sitter grammar should already handle most constructs. Mapping the .cu / .cuh extensions to the C++ parser might cover the majority of cases. The main CUDA-specific syntax that isn't valid C++ is the kernel launch operator (<<<...>>>), which would need separate handling — but qualifiers like __global__ / __device__ generally parse fine as identifiers.

Thank you for considering this!

zouyuanqing:
## Summary

Adds language support for CUDA (\.cu\ / \.cuh\) and IEC 61131-3 SCL (\.scl\ / \.st\).

### Type & extraction

- New \cuda\ language type, mapped to \	ree-sitter-cpp.wasm\ + \cppExtractor\
- \.cu\ / \.cuh\ extensions
- CUDA is a C++ dialect. The C++ parser handles function definitions, structs, templates, and \#include\ directives. Only \<<<...>>>\ kernel launch syntax and \__global__\ / \__device__\ storage specifiers may parse as errors — the test comment explicitly documents this limitation.

### Resolution pipeline (the part that makes this actually useful)

CUDA files now share the full C/C++ resolution path:

| Module | Change |
|--------|--------|
| \import-resolver.ts\ | \.cuh\ / \.cu\ in extension resolution; \-I\ include-dir search; standard library detection; import mappings (all gated with \|| language === 'cuda'\) |
| \
ame-matcher.ts\ | CUDA added to C family (\c: 'c', cpp: 'c', cuda: 'c'\) for cross-language resolution |
| \index.ts\ | Namespace filtering shared with C/C++ |

This means \#include\ references from \.cu\ files resolve to their headers, cross-file function calls produce edges, and \.cu\ ↔ \.cpp\ ↔ \.h\ cross-references work correctly.

## SCL (IEC 61131-3 ST) — Advances #441

- New \scl\ language type, \.scl\ / \.st\ extensions
- File-level tracking only (matching yaml/twig/properties conventions)
- \GrammarLanguage\ type explicitly excludes \scl\ to prevent type errors
- **Fix:** SCL files were silently dropped during single-file indexing — \isLanguageSupported\ now returns \	rue\

### SCL
- Language detection by extension
- Support / grammar-loaded / file-level-only flags
- End-to-end file-record tracking: 2 ST files indexed, verified both appear in \getFiles()\ with correct \language: scl\, zero symbol nodes

### CUDA
- Fixture extraction from realistic selective-scan kernel (\selective_scan.cu\)
- \__global__\ and \__host__\ functions extracted with \language: cuda\
- \#include <cuda_runtime.h>\ as import node
- Function call edges within CUDA code
- \.cuh\ header struct definition extraction

## Files changed

| File | + | - | Notes |
|------|---|---|-------|
| \src/types.ts\ | 2 | 0 | LANGUAGES |
| \src/extraction/grammars.ts\ | 7 | 2 | EXTENSION_MAP, WASM, display, type fix |
| \src/extraction/languages/index.ts\ | 1 | 0 | cuda → cppExtractor |
| \src/resolution/import-resolver.ts\ | 7 | 4 | Full C++ resolution sharing |
| \src/resolution/index.ts\ | 1 | 1 | Namespace filtering |
| \src/resolution/name-matcher.ts\ | 1 | 1 | C family membership |
| \__tests__/extraction.test.ts\ | 149 | 1 | SCL + CUDA test suites |
| \__tests__/fixtures/cuda/selective_scan.cu\ | 32 | 0 | Real kernel fixture |
| \CHANGELOG.md\ | 7 | 0 | Unreleased entries |

**Total: +216 −10**

## Not in this PR

- \	ree-sitter-cuda.wasm\ (separate grammar — can replace C++ wasm in follow-up)
- SCL extractor (foundation only — \	ree-sitter-structured-text\ grammar exists and can be vendored in a follow-up)
- PTX (\.ptx\) — no tree-sitter grammar; requires new grammar from scratch

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### New Features
+
+- CodeGraph now indexes CUDA (.cu / .cuh) and IEC 61131-3 ST (.scl / .st) files. CUDA reuses the C++ parser and shares its resolution paths — `#include` headers, function calls, struct definitions, and cross-language references between `.cu`, `.cpp`, and `.c`/`.h` files all resolve correctly. SCL files are tracked at the file-record level (no symbol extraction, matching the YAML and Twig conventions).
+
+### Fixes
+
+- SCL (.scl / .st) files were silently dropped during single-file indexing instead of being tracked as file records, because `isLanguageSupported` returned `false` for the language. They now behave consistently across both the batch and single-file indexing paths.
 
 ## [1.0.1] - 2026-06-13
 
--- a/__tests__/extraction.test.ts
+++ b/__tests__/extraction.test.ts
@@ -10,7 +10,7 @@ import * as path from 'path';
 import * as os from 'os';
 import { CodeGraph } from '../src';
 import { extractFromSource, scanDirectory, buildDefaultIgnore } from '../src/extraction';
-import { detectLanguage, isLanguageSupported, getSupportedLanguages, initGrammars, loadAllGrammars, isSourceFile } from '../src/extraction/grammars';
+import { detectLanguage, isLanguageSupported, getSupportedLanguages, isGrammarLoaded, isFileLevelOnlyLanguage, initGrammars, loadAllGrammars, isSourceFile } from '../src/extraction/grammars';
 import { normalizePath } from '../src/utils';
 
 beforeAll(async () => {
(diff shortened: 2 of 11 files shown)


icyveins7:
As someone who works primarily on CUDA, I definitely need this ASAP.

EDIT: just would like to add that I've pulled his feature branch and it's working great. It seems like this won't be reviewed any time soon so I guess I'll be using it directly from there for now..

PR #877, opened 2026-06-14
Fix watcher degradation on watch exhaustion and prolonged lock contention

### LINKED ISSUE #876: Watcher can stay half-broken after watch exhaustion or prolonged lock contention
## Summary

The live file watcher has two failure modes where it can stop being trustworthy without transitioning into a clear degraded/off state:

1. **Watch-resource exhaustion (EMFILE / ENFILE)** only logs (or is partially tolerated on the per-directory path), so the watcher may remain "alive" after live watching has effectively failed.
2. **Prolonged sync lock contention (LockUnavailableError)** retries forever at the normal debounce cadence, with no bounded retry policy or terminal degraded state.

## Why this matters

This is a reliability bug, not just a retry-policy preference.

In both cases, the system can enter a half-broken state:

- auto-sync is no longer reliable
- the index can silently drift stale
- the process keeps doing repeated useless work / logging
- users and MCP callers are not clearly told that live watching has effectively stopped working

For long-running MCP/daemon usage, this is especially bad because the process may stay up while its auto-update guarantees are no longer real.

### Watch exhaustion

In `src/sync/watcher.ts`, runtime `fs.watch` exhaustion does not cleanly disable the watcher. On the recursive path it only logs; on the per-directory path it can be silently tolerated at the directory level.

### Lock contention

`CodeGraph.watch()` converts the zero-result lock-failure sentinel into `LockUnavailableError`, and `FileWatcher.flush()` keeps `pendingFiles` and reschedules forever. That is reasonable for brief contention, but not for a long-lived writer.

### For EMFILE / ENFILE

The watcher should:
- degrade/disable itself once
- emit one actionable warning
- stop repeated error spam
- direct users to `codegraph sync` / git sync hooks

### For prolonged LockUnavailableError

The watcher should:
- keep the current quiet behavior for short contention
- retry with bounded backoff
- after a threshold, enter a clear degraded/off state instead of retrying forever at the normal debounce cadence

## Severity

**High** reliability issue for watcher-based workflows, especially MCP/daemon sessions.

This does not always show up as a tight CPU spin, but it can leave CodeGraph serving stale results while appearing operational, which is worse than a clean fail in many real workflows.

thismilktea:
Closes #876 
## Summary

This PR hardens the live file watcher around two reliability failure modes:

1. **Watch-resource exhaustion (EMFILE / ENFILE)** now disables live watching cleanly instead of leaving the watcher half-broken.
2. **Prolonged sync lock contention (LockUnavailableError)** no longer retries forever at the normal debounce cadence; it now uses bounded backoff and eventually degrades auto-sync explicitly.

The goal is to fail closed and clearly once live watching is no longer trustworthy, while preserving the current behavior for normal edits and short-lived contention.

### Watch exhaustion

- Detects watch-resource exhaustion more explicitly
- Degrades/stops the watcher instead of just logging forever
- Emits a single actionable warning
- Adds an `onDegraded` callback so callers can observe permanent watcher degradation

### Lock contention

- Keeps the current quiet behavior for brief lock contention
- Adds bounded retry backoff for repeated `LockUnavailableError`
- Stops infinite normal-debounce retries under long-lived contention
- Degrades auto-sync after the retry threshold is crossed

### Internal cleanup

- Separates normal debounce scheduling from retry scheduling
- Tightens exhaustion detection so message matching is only used as a fallback when no `err.code` is available

## Why

Before this change, the watcher could remain "alive" after it had effectively stopped being trustworthy:

- EMFILE / ENFILE could leave live watching unusable without a clean degraded/off transition
- Prolonged lock contention could keep the watcher retrying forever with no terminal state
- Callers could continue assuming auto-sync was still working even while the index drifted stale

This is especially problematic for long-running MCP/daemon sessions.

## Tests

Added / extended watcher tests for:

- Startup watch exhaustion
- Runtime recursive watcher exhaustion
- Prolonged `LockUnavailableError` degradation
- Degraded-state callback notification

**Verified with:**

```bash
npx vitest run __tests__/watcher.test.ts __tests__/watch-policy.test.ts

### DIFF
--- a/__tests__/watcher.test.ts
+++ b/__tests__/watcher.test.ts
@@ -93,7 +93,8 @@ describe('FileWatcher', () => {
 
     it('should not start when fs.watch setup exhausts watch/file resources', () => {
       const syncFn = vi.fn().mockResolvedValue({ filesChanged: 0, durationMs: 0 });
-      const watcher = new FileWatcher(testDir, syncFn, { debounceMs: 100 });
+      const onDegraded = vi.fn();
+      const watcher = new FileWatcher(testDir, syncFn, { debounceMs: 100, onDegraded });
       const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
       __setFsWatchForTests(() => {
         const err = new Error('too many open files') as NodeJS.ErrnoException;
@@ -103,6 +104,8 @@ describe('FileWatcher', () => {
 
       expect(watcher.start()).toBe(false);
       expect(watcher.isActive()).toBe(false);
+      expect(onDegraded).toHaveBeenCalledTimes(1);
+      expect(onDegraded).toHaveBeenCalledWith(expect.stringContaining('auto-sync disabled'));
       expect(warnSpy).toHaveBeenCalledWith(
         expect.stringContaining('File watcher disabled'),
         expect.objectContaining({
@@ -113,6 +116,7 @@ describe('FileWatcher', () => {
 
     it('should degrade once when the recursive watcher emits EMFILE at runtime', async () => {
       const syncFn = vi.fn().mockResolvedValue({ filesChanged: 0, durationMs: 0 });
+      const onDegraded = vi.fn();
       const handlers = new Map<string, Array<(arg?: unknown) => void>>();
       const fakeWatcher = {
         on: vi.fn((event: string, handler: (arg?: unknown) => void) => {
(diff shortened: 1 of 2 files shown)


colbymchenry:
Thanks for this, @thismilktea — solid diagnosis and the degrade/backoff approach was the right one. I've merged it into `main` (with a couple of corrections) as #891, and credited you in the changelog.

Two things I adjusted on top of your branch:

1. **Linux `start()` consistency.** On the per-directory watch path (Linux), a startup exhaustion degraded but `start()` still returned `true`, so the MCP server would report the watcher *"active"* on a watcher that had just disabled itself — and the new `should not start when fs.watch setup exhausts` test would fail there. Both watch strategies now return `false` consistently (verified natively in Docker).
2. **`onDegraded` wiring.** The callback wasn't consumed anywhere, so MCP/daemon callers still weren't told. It's now wired into the MCP server (`File watcher degraded — …`).

Also validated the recursive path on Windows (Parallels) in addition to macOS/Linux. Closing in favor of #891 — thanks again for driving this. 🙏

Issue #660, opened 2026-06-03
[PHP] include/require(_once) not mapped to dependency edges — only namespace `use`; impact/callers miss the file-include graph


atahan150:
## What happened

On a procedural / script-style PHP codebase (the kind that organizes code with `include`/`require` rather than Composer/PSR-4 namespaces — plenty of legacy apps, plain-PHP sites, and older CMS themes still look like this), codegraph builds an excellent **call** graph but produces **zero file→file dependency edges** for `include` / `require` / `include_once` / `require_once`.

Concretely: a page file `foo.php` whose first line is `include("../loader.php");` yields `callers(loader.php) = 0` and no edge of any kind linking `foo.php → loader.php`. The same holds for `require_once`. Only namespace `use` statements become import/dependency edges.

The call graph itself is great here — `callers`/`callees`/`impact` correctly attribute top-level (file-scope) function calls to the file node, which is exactly right for procedural code. This report is **only** about the missing include/require dependency layer.

## Why it matters

For namespaced codebases this is a non-issue: `use` covers the dependency graph. But for include-based PHP, the include statement *is* the dependency mechanism — it's how a file gains access to the functions/constants it calls. Without those edges:

- `impact`/`callers` can't answer "if I change `helpers.php`, which files that `include` it (directly or transitively) are affected?" beyond the direct function-call edges.
- The file-level architecture graph is missing its primary structural relationship for this style of code.

## Root cause (source)

`src/extraction/languages/php.ts`:

```ts
importTypes: ['namespace_use_declaration'],
```

`importTypes` lists only `namespace_use_declaration`. The PHP tree-sitter grammar represents includes as `include_expression`, `require_expression` (and the `_once` variants), which appear in none of the captured node-type lists (`importTypes`, `callTypes`, `variableTypes`, …), and `visitNode` only special-cases `const_declaration` and trait `use_declaration`. So include/require nodes are simply never visited as dependency edges.

## Suggested fix

Capture `include_expression` / `require_expression` (+ `_once` variants) and emit a file→file dependency/import edge, resolving the argument string to a path (relative to the including file / configured roots). Even resolving only static string literals (the common case) would cover the overwhelming majority of real include sites; dynamic includes (`include $var`) can be skipped.

## Repro

1. Two files in the same dir:
   - `lib.php`: `<?php function greet() { return "hi"; }`
   - `page.php`: `<?php require_once("lib.php"); echo greet();`
2. Index the directory.
[...]

PR #901, opened 2026-06-15
test(resolution): close SQLite handles before rmSync so Windows teardown doesn't EPERM

JOhnsonKC201:
## What

Four end-to-end tests in `__tests__/resolution.test.ts` fail on Windows in their `finally` cleanup with `EPERM` — the **C/C++ include-dir scan** test and the three **PHP include resolution** tests (#660):

```
Error: EPERM, Permission denied: ...\Temp\codegraph-cpp-e2e-xxxx
Error: EPERM, Permission denied: ...\Temp\codegraph-php-e2e-xxxx
Error: EPERM, Permission denied: ...\Temp\codegraph-php-subdir-xxxx
Error: EPERM, Permission denied: ...\Temp\codegraph-php-misresolve-xxxx
```

The assertions themselves pass; only the teardown throws.

## Why

Each test opens a second read-only `DatabaseConnection` to inspect edges directly and never closes it. Each also removes its per-test `tempProject` in a `finally` that runs **before** `afterEach`'s `cg.destroy()` — so the `CodeGraph` instance's own SQLite handle is still open at removal time too.

On POSIX an open file can be unlinked, so `fs.rmSync(tempProject)` silently succeeds and the suite is green on the macOS/Linux dev machines. On Windows an open SQLite handle locks `codegraph.db`, so `fs.rmSync` throws `EPERM` and the test fails.

These four are **not** in the documented "known pre-existing Windows failures" list in `CLAUDE.md` (which covers the `mcp-initialize` / `mcp-roots` suites and the symlink test) — they're a distinct, self-contained leak.

## Fix

Close the local reader **and** the `CodeGraph` instance before `rmSync`. `DatabaseConnection.close()` is idempotent, so `afterEach`'s later `cg.destroy()` stays safe. Test-only change; no product code touched.

## Verification (Windows 11, wasm SQLite backend)

`npx vitest run __tests__/resolution.test.ts` (single-fork):

| | Before | After |
|---|---|---|
| resolution.test.ts | **4 failed** (EPERM) \| 80 passed | **0 failed \| 84 passed** |

No other test changed. The unrelated pre-existing `Worker exited unexpectedly` teardown crash is present identically on `main` and on this branch.

### DIFF
--- a/__tests__/resolution.test.ts
+++ b/__tests__/resolution.test.ts
@@ -1964,6 +1964,7 @@ func main() {
     // feature can't silently regress to a no-op in the indexing flow.
     it('connects #include to the real header file via include-dir scan (end-to-end)', async () => {
       const tempProject = fs.mkdtempSync(path.join(os.tmpdir(), 'codegraph-cpp-e2e-'));
+      let db: DatabaseConnection | undefined;
       try {
         fs.mkdirSync(path.join(tempProject, 'include'), { recursive: true });
         fs.mkdirSync(path.join(tempProject, 'src'), { recursive: true });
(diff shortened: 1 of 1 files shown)


JOhnsonKC201:
Friendly ping — this closes the SQLite handles before `rmSync` so the Windows test teardown stops hitting EPERM. Happy to rebase or adjust if anything is needed.

Issue #764, opened 2026-06-09
CodeGraph mixes up TypeScript classes with same name


Igorgro:
When using CodeGraph on large projects, for example Nestjs with multiple applications (monorepo), there can be a situation where each application has a class with same name. For example we have something lik UserService, UserModule, UserRepository in each app. The problem is that there is no way to separate them somehow (like packages in Java and namespaces in C#), so CodeGraph places them into one global scope resulting in coding agent being confused by name collision. For example Claude Code is smart enough and always gives a warning that MCP server is possibly mixing up two or more unrelated classes. It's good that Claude can understand such things but it were better if CodeGraph could someone solve this problem on it's own side

colbymchenry:
Thanks for the precise report — reproduced on a NestJS-style monorepo fixture and on a real one (amplication, 54k symbols), and fixed in #813. It turned out to be two separate problems:

**1. The conflation you (and Claude) saw was mostly in the tools.** The graph itself was keeping your apps' classes apart — but `codegraph_callers`/`codegraph_callees` merged every same-named match into one flat list, and `codegraph_impact` merged all their blast radii into one. Now, when several *unrelated* definitions share a name, these tools report **one section per definition**, each labeled with its file:

```
## Callers of findAll — 2 distinct definitions (narrow with `file`)

### UserService::findAll (method) — apps/admin/src/users/user.service.ts:6
- list (method) - apps/admin/src/users/user.controller.ts:6

### UserService::findAll (method) — apps/billing/src/users/user.service.ts:6
- list (method) - apps/billing/src/users/user.controller.ts:6
```

…and they accept a `file` argument (path or suffix) to focus exactly the definition you mean. Impact reports a separate blast radius per definition instead of an overstated union. Same-file overloads still merge — only unrelated definitions are separated.

**2. There was also a real wrong-edge source, which your report led us to.** PascalCase *type* references from plain `.ts` files were being run through React *component* resolution, which on multi-app monorepos linked a file's own type to an arbitrary same-named class in another package — on amplication this accounted for **1,036 wrong cross-package edges (96% now gone)**. That's fixed too; component resolution now only applies to JSX-capable files and never guesses among multiple candidates.

Ships in the next release — re-index afterwards (`codegraph index -f`) to drop any wrong cross-package edges from existing indexes.

Issue #750, opened 2026-06-09
Tracking: chained factory/singleton call resolution across statically-typed languages


colbymchenry:
A call whose **receiver is itself a call** — a factory / singleton / builder that returns an object — should produce a `calls` edge to the chained method:

```java
Foo.getInstance().bar();   // bar() should resolve to Foo::bar
```

This was fixed for **C++ (#645)** and **PHP (#608)**. A probe of the other statically-typed README languages found the **same gap in all of them** — and in **7 of the 9** the chained method currently resolves to the *wrong* class (a same-named method on an unrelated type), which is a correctness bug, not just missing coverage.

The fix is the 3-part #645/#608 mechanism, per language: capture the factory's declared **return type**, preserve the chained receiver at extraction, then resolve **and validate** the chained method on the inferred type — so a wrong inference produces **no edge**, never a wrong one.

| Language | Probe result (`Foo.factory().method()`) | Status |
|---|---|---|
| Java | edge missing | ✅ #751 |
| Kotlin | wrong edge → same-named decoy | ✅ #752 |
| C# | edge missing | ✅ #753 |
| Swift | wrong edge → same-named decoy | ✅ #755 |
| Rust | wrong edge → same-named decoy | ✅ #757 |
| Go | wrong edge → same-named decoy (same-package `New().Method()`; cross-package `pkg.New()` was #469) | ✅ #760 |
| Scala | wrong edge → same-named decoy | ✅ #761 |
| Dart | wrong edge → same-named decoy (factory-constructor call also unresolved) | ✅ #762 |
| TypeScript | wrong edge → same-named decoy | ⏭️ evaluated, skipped (see note) |

Dynamically-typed languages can't use this approach (no declared return types to read): **Ruby** and **plain JS** are out of scope; **Python** is partial (optional `-> T` hints, see #578).

Each language is validated the same way before merge: synthetic decoy + absent-method safety tests, then a real-repo A/B on a public OSS repo (node count unchanged, 0 edges lost, +N chained edges recovered, precision spot-checked).

---

### TypeScript — evaluated and consciously skipped

The deterministic chain mechanism was fully implemented and probe-validated for TS (5 synthetic tests passed, including decoy-collision precision and a no-regression instance-chain guard), but **real-repo A/B showed a net recall regression, so it was not shipped**:

| Repo | A/B (unique `calls` edges) |
|---|---|
| typeorm (495 files) | **+0 / −6** |
[...]

PR #920, opened 2026-06-18
feat(resolution): resolve this.<field>.method() on the field's declared type (TS/JS)

tqrcisio:
## Problem

A method call on an injected or declared field, like `this.userService.findAll()`, degrades to the bare name `findAll` during extraction: the receiver `this.userService` is a `member_expression`, not an identifier, so the receiver is dropped. When several classes define a same-named method, which is the norm in NestJS where every service has `findAll`/`create` and every repository has `findOne`, the bare name binds to an arbitrary one. Two failure modes follow, both visible in the graph:

- the real target is left with **zero callers** (a false "dead code" signal); and
- a **wrong edge** points at a same-named method on an unrelated class.

This is the misbinding #750 describes, in the shape that dominates real-world TypeScript. Minimal repro (two services sharing `findAll`):

```ts
// user.controller.ts
constructor(private readonly userService: UserService) {}
list() { return this.userService.findAll(); }   // bound to ProductService::findAll
```

## Fix

Mirrors the existing Java `this.userbo.toLogin2()` handling, in two parts:

1. **Extraction** (`tree-sitter.ts`): when a member call's receiver is `this.<field>`, re-encode the call as `<field>.method` instead of dropping to the bare method name. Deeper chains (`this.a.b.method()`) keep the bare name.

2. **Resolution** (`name-matcher.ts`): a TS/JS branch in `matchMethodCall` recovers the field's declared type from the enclosing class and resolves the method on it. Two declaration sites are covered, both ubiquitous in NestJS:
   - **constructor parameter property**, `constructor(private readonly userService: UserService)`, read from the constructor method's signature;
   - **class-body field**, `private readonly repo: Repo;`, read from the `property` node's signature.

   Resolution goes through `resolveMethodOnType`, which validates the method exists on the inferred type (and its supertypes), so a wrong inference yields **no edge rather than a wrong one**. When a per-app type repeats across a monorepo (`apps/billing` and `apps/admin` each with a `UserService`), the candidate is disambiguated by directory proximity to the call site, the same file-path proximity the bare-name path relied on before re-encoding.

When the field's type cannot be recovered (untyped field, external or generic type), nothing is forced: the ref falls through to the existing strategies, so behavior there is unchanged.

## Validation

Graph-level A/B, baseline build (`main`) vs this build, indexing five open-source NestJS repos. **Node count is identical in every repo** (the fix adds no nodes):

| repo | nodes (base = new) | calls (base) | calls (new) | retargeted | dropped |
|---|---|---|---|---|---|
| [nestjs/nest](https://github.com/nestjs/nest) | 15570 = 15570 | 12079 | 11816 | 414 | 249 |
| [ghostfolio/ghostfolio](https://github.com/ghostfolio/ghostfolio) | 10443 = 10443 | 3762 | 3605 | 289 | 188 |
| [jmcdo29/testing-nestjs](https://github.com/jmcdo29/testing-nestjs) | 1263 = 1263 | 292 | 285 | 18 | 7 |
| [immich-app/immich](https://github.com/immich-app/immich) (`server/`) | 12091 = 12091 | 16277 | 16244 | 648 | 28 |
| [twentyhq/twenty](https://github.com/twentyhq/twenty) (`packages/twenty-server/`) | 83182 = 83182 | 51982 | 51858 | 2756 | 139 |

About 4,100 call edges across the five repos are **retargeted** from a wrong/self binding to the correct injected dependency. Verified against source:

- immich `UserAdminService::getSessions`: `this.sessionRepository.getByUserId(id)` was bound to `ApiKeyRepository::getByUserId`, now `SessionRepository::getByUserId` (field `sessionRepository: SessionRepository`).
- twenty `BillingGaugeService`: `this.twentyConfigService.get(...)` was bound to `ApplicationConnectionsController::get` (a controller), now `TwentyConfigService::get`.
- nest `RolesGuard::canActivate`: `this.reflector.get(...)` was bound to `ConfigService::get`, now `Reflector::get`.
- ghostfolio `PortfolioController::getHoldings`: was a self-loop, now `PortfolioService::getHoldings`.

The **dropped** edges are bare-name mis-binds, not lost-correct edges. Audited jmcdo29/testing-nestjs: 7/7 drops were calls on an external library type, Mongoose `Model<CatDoc>` and TypeORM `Repository<Cat>`, whose `create`/`update`/`findOne`/`remove` had been bound to same-named **controller** methods. ghostfolio's drops are dominated by Angular frontend components whose `this.x.get()` was bound to a backend service or a test mock (`HttpClientMock::get`); nest's are coincidental binds on common names (`this.subject.next()` to `ExceptionsHandler::next`). The net edge-count decrease is wrong-edge removal plus correct retargeting; the raw counts understate the precision gain, because both a corrected edge and a removed wrong edge lower the total.

## Tests

`__tests__/ts-di-method-resolution.test.ts` (new): constructor-injected and class-body-field resolution, same-name disambiguation, and the type-validation path. The existing `same-name-disambiguation.test.ts` (#764) keeps passing: its NestJS monorepo fixture is exactly this scenario. The only failures on my machine are the pre-existing timing-sensitive mcp-daemon / ppid-watchdog / sync tests, which pass in isolation on a clean checkout as well.

### DIFF
--- a/__tests__/ts-di-method-resolution.test.ts
+++ b/__tests__/ts-di-method-resolution.test.ts
@@ -1,0 +1,133 @@
+/**
+ * TS/JS `this.<field>.method()` resolution via the field's declared type.
+ *
+ * A method call on an injected/declared field, like `this.userService.findAll()`,
+ * used to degrade to the bare name `findAll`, which (a) misbound to a
+ * same-named method on an unrelated class when several existed, and (b) left
+ * the real target with zero callers. The receiver `this.<field>` is now
+ * re-encoded as `<field>.method` and resolved on the field's declared type,
+ * recovered from the constructor parameter property or a class-body field.
+ *
+ * This is the static-analysis case that dominates NestJS (controllers calling
+ * injected services, services calling injected repositories) and any typed TS
+ * OOP code. resolveMethodOnType validates the method exists on the inferred
+ * type, so the typed path only ever REPLACES a wrong/ambiguous bare-name
+ * binding with the correct one. It never forces an unvalidated edge.
+ */
+
+import { describe, it, expect, beforeAll, afterAll } from 'vitest';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as os from 'os';
+import { CodeGraph } from '../src';
+import { initGrammars, loadAllGrammars } from '../src/extraction/grammars';
+
+let tmpDir: string;
+let cg: CodeGraph;
+
+/** `Class::method` qualified names every `calls` edge out of the method `fromQn` points at. */
+const callTargetsFrom = (fromQn: string): string[] => {
+  const from = cg.getNodesByName(fromQn.split('::').pop()!).find((n) => n.qualifiedName === fromQn);
+  if (!from) return [];
+  return cg
+    .getOutgoingEdges(from.id)
+    .filter((e) => e.kind === 'calls')
+    .map((e) => cg.getNode(e.target)?.qualifiedName)
+    .filter((qn): qn is string => Boolean(qn));
+};
+
+const callerCount = (methodQn: string): number => {
+  const node = cg
+    .getNodesByName(methodQn.split('::').pop()!)
+    .find((n) => n.qualifiedName === methodQn);
+  if (!node) return -1;
+  return cg.getIncomingEdges(node.id).filter((e) => e.kind === 'calls').length;
+};
+
+beforeAll(async () => {
+  await initGrammars();
+  await loadAllGrammars();
+
+  tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'cg-tsdi-'));
+  const mk = (rel: string, content: string): void => {
+    const p = path.join(tmpDir, rel);
+    fs.mkdirSync(path.dirname(p), { recursive: true });
+    fs.writeFileSync(p, content);
+  };
+
+  // Two services share findAll() and create(); the bare name alone can't tell
+  // them apart. wipe() exists only on ProductService.
+  mk(
+    'src/user.service.ts',
+    ['export class UserService {', '  findAll() { return ["u"]; }', '  create() { return "u"; }', '}'].join('\n')
+  );
+  mk(
+    'src/product.service.ts',
+    [
+      'export class ProductService {',
+      '  findAll() { return ["p"]; }',
+      '  create() { return "p"; }',
+      '  wipe() { return "gone"; }',
+      '}',
+    ].join('\n')
+  );
+  // Constructor parameter property injection (the NestJS norm).
+  mk(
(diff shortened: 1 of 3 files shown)


tqrcisio:
@colbymchenry when you have a moment, would appreciate your review on this one.

It ports the Java/Kotlin field-injection resolution to TS/JS: `this.<field>.method()` now resolves on the field's declared type (constructor parameter property or class-body field) instead of degrading to a bare method name. That fixes the NestJS controller -> service -> repository misbinding, which is the TypeScript form of #750 and a common source of false "zero callers".

Validated graph-level (`main` vs this branch) on five NestJS repos: nestjs/nest, ghostfolio, jmcdo29/testing-nestjs, immich (server), twenty (twenty-server). Node count identical in all five, around 4,100 call edges retargeted wrong->right, and the dropped edges are bare-name mis-binds to external library types (Mongoose `Model`, TypeORM `Repository`, RxJS) verified by source audit. Full numbers and method are in the description. Happy to adjust anything.

maxmilian:
This is still live on `main`, and it now has a second issue and an independent root-cause analysis behind it.

**Same defect, reported again since this PR opened.** #1496 (2026-08-01) reports `this.<field>` calls binding to a same-named method on an unrelated class, reproduced independently by two people on 1.5.0 linux-x64. It does not reference this PR — it was filed against the symptom rather than against #750 — which is probably why the two threads never met.

**The RCA in #1496 lands exactly where this PR's extraction hunk does.** @inth3shadows forced the calls to stay unresolved and read `unresolved_refs` directly: `this.<field>.method()` arrives as a bare `zzzNoSuchMethod` with the receiver gone, while an ordinary local receiver arrives as `m.zzzAlsoNoSuch` with the receiver intact. So the receiver is dropped during extraction, not mis-ordered during resolution — which is the branch this PR adds to `tree-sitter.ts`. It also explains the follow-on effect: with the receiver gone, the bare name never matches the dotMatch shape the receiver-inference branch in `name-matcher.ts` keys on, so that whole path is skipped.

**Not superseded.** I checked `main` at `44e1812`: no TypeScript-side fix has landed. The nearby recent commits are the receiver-inference scan memo (`974e6c8`), literal-receiver builtins (`c472cfb`) and the Go field-chain fix (`41c2029`) — none of them touch this path. The branch is `CONFLICTING` against current `main`, but that is drift, not obsolescence.

**It is also the same shape as work already in flight.** #1599 resolves Rust `self.field.method()` on the field's declared type; this is the TypeScript face of that, and mirrors the existing Java `this.userbo.toLogin2()` unwrap the same way.

I have not run this branch (it needs a rebase first), so this is a read of the diff and of current `main`, not a test result. Flagging it because the evidence in #1496 arrived after this PR went quiet, and it points the same way this patch already goes.

PR #927, opened 2026-06-19
fix(sync): preserve cross-file caller edges across callee re-index (#899)

### LINKED ISSUE #899: Python: module-attribute callers still missed in large package tree after #715 / 1.0.1
## Summary

Issue #578 / PR #715 appears fixed for the clean minimal Python module-attribute cases in `codegraph@1.0.1`, but the original RAGFlow production case still has missing `calls` edges after a full reindex.

The narrowed finding:

- `from pkg import mod; mod.target(...)` now works in a fresh scratch project.
- `import pkg.mod as mod; mod.target(...)` now works in a fresh scratch project.
- `from api.db.services import dialog_service; dialog_service._private_shim(...)` also works in a fresh scratch project.
- But in the real RAGFlow tree, `from api.db.services import dialog_service; dialog_service._repair_grounded_citation_alignment(...)` still reports no callers, even though both caller and callee nodes are indexed and the call expressions are visible in `codegraph node`.

So the broad #578 bug is fixed, but there is still a narrower miss in the larger package/import context.

## Environment

- `codegraph --version` -> `1.0.1`
- Linux
- Project: RAGFlow fork
- Full reindex performed after upgrading:

```bash
codegraph index .
```

The index reported:

```text
Indexed 2,982 files
0 nodes, 4,425 edges in 14.9s
```

## Control checks that now pass

I created a fresh scratch project under `/tmp/codegraph-attr-test`:

```python
# pkg/mod.py
def target(value: int) -> int:
    return value + 1

# use_from_pkg_import_module.py
from pkg import mod

def caller_from_pkg_import_module() -> int:
    return mod.target(1)

# use_import_pkg_module_as_module.py
import pkg.mod as mod

def caller_import_pkg_module_as_module() -> int:
    return mod.target(2)

# use_from_pkg_mod_import_target.py
from pkg.mod import target

def caller_direct_import() -> int:
    return target(3)
```

After `codegraph init /tmp/codegraph-attr-test`:

```bash
codegraph callers target -p /tmp/codegraph-attr-test
```

correctly reports all three callers:

```text
caller_from_pkg_import_module
caller_direct_import
caller_import_pkg_module_as_module
```

I also tested the exact package-level import shape in a fresh scratch project under `/tmp/codegraph-dialog-shape`:

```python
# api/db/services/dialog_service.py
def shim(value: int) -> int:
    return value + 10

def _private_shim(value: int) -> int:
    return value + 20

# test/unit_test/api/db/services/test_prompt.py
from api.db.services import dialog_service

def test_calls_module_attribute() -> None:
    assert dialog_service.shim(1) == 11

def test_calls_private_module_attribute() -> None:
    assert dialog_service._private_shim(1) == 21
```

After `codegraph init /tmp/codegraph-dialog-shape`:

```bash
codegraph callers shim -p /tmp/codegraph-dialog-shape
codegraph callers _private_shim -p /tmp/codegraph-dialog-shape
```

both correctly report the test callers.

## Real RAGFlow case that still fails

The real code has:

```python
# test/unit_test/api/db/services/test_dialog_service_prompt_sanitization.py:21
from api.db.services import dialog_service
```

and calls such as:

```python
# test/unit_test/api/db/services/test_dialog_service_prompt_sanitization.py:1770
answer, cited = dialog_service._repair_grounded_citation_alignment(
    "Feng Shi Pian and Feng Shi Xiao Tong Pian are listed as formulas [ID:2].",
    ledger,
)

# test/unit_test/api/db/services/test_dialog_service_prompt_sanitization.py:1774
answer = dialog_service._preserve_grounded_reference_labels(answer, ledger)
```

The target functions are indexed:

```bash
codegraph node _repair_grounded_citation_alignment -p /home/jo/rf/ragflow_p1
codegraph node _preserve_grounded_reference_labels -p /home/jo/rf/ragflow_p1
```

and the caller function is indexed with the call expressions visible:

```bash
codegraph node test_repair_grounded_citation_alignment_replaces_wrong_valid_id_before_label_injection -p /home/jo/rf/ragflow_p1
```

shows:

```python
answer, cited = dialog_service._repair_grounded_citation_alignment(...)
answer = dialog_service._preserve_grounded_reference_labels(answer, ledger)
```

But callers are still missing:

```bash
codegraph callers _repair_grounded_citation_alignment -p /home/jo/rf/ragflow_p1
codegraph callers _preserve_grounded_reference_labels -p /home/jo/rf/ragflow_p1
```

Observed:

```text
No callers found for "_repair_grounded_citation_alignment"
No callers found for "_preserve_grounded_reference_labels"
```

Same symptom for other module-attribute calls from the same test file, for example:

```python
dialog_service._compute_prior_weight(...)
dialog_service._expand_query_with_symbols(...)
```

`codegraph callers _compute_prior_weight` reports internal/import-related callers, but not the test functions that call it via `dialog_service._compute_prior_weight(...)`.

## Expected

After #715, the real RAGFlow production case from #578 should also get `calls` edges:

```text
test_repair_grounded_citation_alignment_replaces_wrong_valid_id_before_label_injection
  -> _repair_grounded_citation_alignment

test_repair_grounded_citation_alignment_replaces_wrong_valid_id_before_label_injection
  -> _preserve_grounded_reference_labels
```

and similarly for the other `dialog_service.<function>(...)` test calls.

## Why this matters

The practical failure mode is still the same as #578: these functions look unused through `codegraph_callers`, even though regression tests call them by module-attribute access. That makes dead-code cleanup and impact analysis unsafe unless users pair CodeGraph with textual grep.

The fact that fresh scratch projects pass suggests the remaining bug is not the original broad resolver gap, but something about the larger package/import context in the real RAGFlow tree.

## Ask

Can you take a second look at Python module-attribute resolution for larger package trees where:

```python
from api.db.services import dialog_service
dialog_service.some_function(...)
```

The minimal fixture now passes in 1.0.1, but the original production case still misses the edge after a full reindex.

JosefAschauer:
## Problem

`codegraph sync` (and the file-watcher auto-sync, which calls `sync()`) silently drops cross-file `calls`/`references` edges from unchanged caller files when a callee file is re-indexed. After editing any file that has module-attribute callers from elsewhere (`from pkg import mod; mod.fn(...)`), `codegraph callers fn` reports 0 callers for functions that have real call sites. A docstring-only edit on the callee is sufficient to trigger it.

Reported in #899. The watcher was a red herring in the original report — the bug is in `sync()`/`indexFile()`, which the watcher and CLI `sync` both call. `codegraph index` was already fixed by #894 (it now clears-then-rebuilds); `sync` remains the fast incremental path and still has the bug.

## Root cause

`storeExtractionResult` deletes a re-indexed file's nodes via `deleteFile`, which cascades through `edges.FK ... ON DELETE CASCADE` (schema.sql:54-55) to delete every edge whose `source` OR `target` is one of those nodes. Edges whose `source` is in the re-indexed file are re-emitted by the extractor, but edges whose `source` is in a *different* (unchanged) file are not — they're silently dropped.

## Fix

Before the delete, snapshot incoming cross-file edges paired with the target node's `(name, kind)` (`getCrossFileIncomingEdgesWithTarget` in `queries.ts`). After re-inserting the file's nodes + same-file edges, re-insert the snapshot — re-resolving each edge's `target` to the re-indexed node's **new** id by `(filePath, kind, name)`.

Node ids are `sha256(filePath:kind:name:line)`, so any line shift in the callee file (e.g. a docstring-only edit above the symbol) changes every target id. A naive re-insert by old id would drop them all; matching by `(kind, name)` is stable across line shifts. If the symbol was renamed/removed, no match is found and the edge stays dropped (correct). `insertEdges` still filters to endpoints that exist, so edges whose caller (source) was deleted are also dropped.

## Verification

- Two regression tests in `__tests__/sync.test.ts` model the RAGFlow production case from #899: a `pkg/mod.py` with two callees, both called from `test/test_callers.py` via `mod.<fn>(...)`.
  - **Test 1**: a docstring-only edit that shifts the second callee's line preserves both incoming edges (the line-shift case that breaks naive id-based restore).
  - **Test 2**: renaming a callee correctly drops its old incoming edge (no phantom preservation against a non-existent symbol).
- **Confirmed both tests fail on pre-fix `main`** and pass with the fix.
- Full vitest suite: **1412 passed, 2 skipped** (80 test files). No regressions.
- Manual repro against the real RAGFlow tree from #899: `_repair_grounded_citation_alignment` (4 callers) and `_preserve_grounded_reference_labels` (5 callers) both survive a `codegraph sync` after a docstring-only edit to `dialog_service.py`. Pre-fix: both drop to 0.

Closes #899.

### DIFF
--- a/__tests__/sync.test.ts
+++ b/__tests__/sync.test.ts
@@ -303,4 +303,138 @@ describe('Sync Module', () => {
       expect(result.changedFilePaths).toBeUndefined();
     });
   });
+
+  describe('Cross-file module-attribute caller edges survive callee re-index (#899)', () => {
+    let testDir: string;
+    let cg: CodeGraph;
+
+    beforeEach(async () => {
+      testDir = fs.mkdtempSync(path.join(os.tmpdir(), 'codegraph-899-'));
+
+      // pkg/mod.py — a module with two functions, both called from a separate
+      // test file via `mod.<fn>(...)` (module-attribute access). This is the
+      // exact shape from the RAGFlow production case in issue #899.
+      fs.mkdirSync(path.join(testDir, 'pkg'), { recursive: true });
+      fs.mkdirSync(path.join(testDir, 'test'), { recursive: true });
+      fs.writeFileSync(
+        path.join(testDir, 'pkg', '__init__.py'),
+        ``
+      );
+      fs.writeFileSync(
+        path.join(testDir, 'pkg', 'mod.py'),
+        [
+          `def callee_one(value):`,
+          `    """First callee — docstring above the second callee so edits here shift its line."""`,
+          `    return value + 1`,
+          ``,
+          ``,
+          `def callee_two(value):`,
+          `    """Second callee, called from the test file via mod.callee_two(...)."""`,
+          `    return value + 2`,
+          ``,
+        ].join('\n')
+      );
+      fs.writeFileSync(
+        path.join(testDir, 'test', 'test_callers.py'),
+        [
+          `from pkg import mod`,
+          ``,
+          ``,
+          `def test_calls_callee_one():`,
+          `    assert mod.callee_one(1) == 2`,
+          ``,
+          ``,
+          `def test_calls_callee_two():`,
+          `    assert mod.callee_two(1) == 3`,
+          ``,
+        ].join('\n')
+      );
+
+      cg = CodeGraph.initSync(testDir, {
+        config: { include: ['**/*.py'], exclude: [] },
+      });
+      await cg.indexAll();
+    });
+
+    afterEach(() => {
+      if (cg) cg.destroy();
+      if (fs.existsSync(testDir)) fs.rmSync(testDir, { recursive: true, force: true });
+    });
+
+    function callerCount(fnName: string): number {
+      const results = cg.searchNodes(fnName);
+      const def = results.map(r => r.node).find(n => n.kind === 'function' && n.name === fnName);
+      if (!def) return -1;
+      return cg.getCallers(def.id).length;
+    }
+
+    it('preserves incoming cross-file calls edges when the callee file is re-indexed', async () => {
+      // Baseline: both callees have one cross-file caller each.
+      expect(callerCount('callee_one')).toBe(1);
+      expect(callerCount('callee_two')).toBe(1);
+
+      // Docstring-only edit to callee_one — adds 1 line, shifting callee_two's
+      // line number. A naive ID-based edge restore would drop callee_two's
+      // incoming edge (its node id changed); the (kind, name) re-resolve
+      // preserves it. A docstring-only edit also confirms zero-AST-change
+      // re-indexes don't sever edges.
+      fs.writeFileSync(
+        path.join(testDir, 'pkg', 'mod.py'),
+        [
+          `def callee_one(value):`,
+          `    """First callee — docstring above the second callee so edits here shift its line."""`,
+          `    """Probe: extra docstring line to shift callee_two's start line by 1."""`,
+          `    return value + 1`,
+          ``,
+          ``,
+          `def callee_two(value):`,
+          `    """Second callee, called from the test file via mod.callee_two(...)."""`,
+          `    return value + 2`,
+          ``,
+        ].join('\n')
+      );
+
+      const result = await cg.sync();
+      expect(result.filesModified).toBe(1);
+
+      // Both incoming cross-file calls edges must survive the callee re-index.
+      expect(callerCount('callee_one')).toBe(1);
+      expect(callerCount('callee_two')).toBe(1);
+    });
+
+    it('drops incoming edges for a callee that was renamed during re-index', async () => {
+      // Baseline.
+      expect(callerCount('callee_one')).toBe(1);
+
+      // Rename callee_one -> callee_one_renamed. The old edge's target
+      // (kind=function, name=callee_one) no longer matches any re-indexed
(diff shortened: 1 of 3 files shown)


colbymchenry (approved):
Verified and merging — genuinely excellent diagnostic work here, @JosefAschauer. The scratch-vs-real narrowing, then pinning it to `sync()` / `indexFile()` rather than resolution, and finally the FK-cascade root cause (cross-file edges whose *source* sits in an unchanged file get cascade-deleted on the callee's `deleteFile` and never re-emitted) — that's a textbook writeup.

Independent verification on my end:
- Reproduced on a minimal `from pkg import mod; mod.fn()` fixture: a docstring-only edit to the callee + `codegraph sync` drops both callers to 0.
- With this PR applied: both callers survive. ✓
- Full test suite green.

Nice side benefit: because `sync()` doesn't re-run the edge synthesizers, this also rescues *synthesized* cross-file edges (dispatch / jsx-render) that sync was silently dropping on callee re-index — not just the module-attribute `calls` case.

Two non-blocking follow-up notes (not merge blockers):
- The `(kind, name)` re-resolve picks the last match when a single file declares two same-name/same-kind symbols (e.g. a conditional `def`) — fine for the common case.
- There's no UNIQUE index on `edges(source, target, kind)`, so the re-insert relies on sync not re-synthesizing (true today) to stay duplicate-free — worth remembering if synthesis ever moves into the incremental path.

Thanks again for the careful work. Merging now. 🙏

PR #942, opened 2026-06-21
chore: nix package

dzmitry-lahoda:


dzmitry-lahoda:
@colbymchenry what do you think?

JakobLichterfeld:
duplicate to #331

Issue #970, opened 2026-06-24
Allow opting out of embedded git repo discovery in gitignored directories


moshuying:
My project has a gitignored resource/ directory (4.2 GB, ~26k files) containing 27 cloned reference repositories (teammate projects kept
  locally for knowledge-base analysis). Despite being listed in .gitignore, codegraph index is extremely slow because findIgnoredEmbeddedRepos actively
  walks into this ignored directory, discovers the 27 nested .git dirs, and recursively indexes all of them via collectGitFiles.

  These repos are not part of my project — they're reference material stored alongside it. There's currently no way to tell codegraph "don't discover
  embedded repos under this path." A configuration option (e.g. a pattern list in codegraph.json to skip embedded-repo discovery in certain directories)
  would let users with this layout avoid indexing irrelevant code.

moshuying:
#972

colbymchenry:
Apologies for this — you're right, and it was a regression on our side. In 1.0.0 we added discovery of embedded git repos nested inside gitignored directories (to support "super-repo of clones" layouts), but it walked into **every** gitignored directory — including a `resource/` full of reference repos like yours — which is exactly the slow, surprising behavior you hit. `.gitignore` should mean "don't index this," full stop.

Fixed in #980 (merged): **CodeGraph respects your `.gitignore` again by default** — a gitignored directory is no longer walked, discovered, or indexed, so your `resource/` dir is left alone with zero config and no more 26k-file traversal.

If you ever *do* want nested repos under a gitignored dir indexed, that's now an explicit opt-in via `codegraph.json`:

```json
{ "includeIgnored": ["some/dir/"] }
```

This ships in **v1.1.1**, which is being released now — please upgrade as soon as it's available. Sorry again for the disruption, and thank you for the clear, well-diagnosed report. (More on your PR #972 over there.)

moshuying:
tks~

PR #972, opened 2026-06-24
 feat(config): add ignore field to codegraph.json

moshuying:
Problem

  When a project has a large gitignored directory containing nested git repos (e.g. cloned reference repos for knowledge-base analysis), codegraph index
  is extremely slow. This happens because findIgnoredEmbeddedRepos actively walks into every gitignored directory looking for .git dirs, then recursively
  indexes all nested repos it finds — even if the user explicitly gitignored them.

  Example: a resource/ directory (4.2 GB, ~26k files, 27 nested git repos) that is already in .gitignore still gets fully traversed and indexed. There
  was no way to tell codegraph "don't look inside this path."

  Solution

  Add an ignore field to codegraph.json — an array of gitignore-style patterns:

  {
    "ignore": ["resource/", "data/"]
  }

  These patterns are injected into:
  1. buildDefaultIgnore — excludes matching files from indexing (also propagates to the file watcher via buildScopeIgnore)
  2. discoverEmbeddedRepoRoots / findIgnoredEmbeddedRepos — skips matching directories during embedded-repo discovery, preventing the expensive recursive
  walk
  3. collectGitFiles / collectGitStatus — ensures consistent exclusion during incremental sync

  Why this approach

  - Single injection point for file filtering — buildDefaultIgnore is the root of both the indexer and watcher scope, so one change covers both
  - Separate injection for embedded-repo discovery — the discovery path uses defaultsOnlyIgnore() (intentionally no .gitignore) so it needed explicit
  threading of user patterns
  - Follows existing patterns — same codegraph.json file, same mtime-cached loader design as the existing extensions field
  - Zero-config preserved — no codegraph.json = no change in behavior

  Test plan

  - 13 new tests in __tests__/ignore-patterns.test.ts:
    - Unit: loadIgnorePatterns parsing, validation, caching
    - Integration: scanDirectory excludes matching files
    - Integration: discoverEmbeddedRepoRoots skips user-ignored directories
    
  link #970

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### New Features
+
+- `codegraph.json` now supports an `ignore` field — an array of gitignore-style patterns that exclude matching paths from indexing and embedded-repo discovery. Useful for large gitignored directories containing cloned reference repos or data files that should never be indexed (e.g. `{"ignore": ["resource/"]}`).
+
 
 ## [1.1.0] - 2026-06-23
 
--- a/__tests__/ignore-patterns.test.ts
+++ b/__tests__/ignore-patterns.test.ts
@@ -1,0 +1,204 @@
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as os from 'os';
+import { loadIgnorePatterns, clearProjectConfigCache } from '../src/project-config';
+import { scanDirectory, buildDefaultIgnore, discoverEmbeddedRepoRoots } from '../src/extraction';
+import { execFileSync } from 'child_process';
+
+function createTempDir(): string {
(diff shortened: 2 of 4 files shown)


colbymchenry:
Thank you for this, @moshuying — and for the report in #970 that surfaced the problem. We went with a root-cause fix in #980 (merged): **CodeGraph now respects `.gitignore` by default** and no longer walks into gitignored directories to discover embedded repos — so your `resource/` case needs no configuration at all.

Because of that, I'm going to close this PR rather than merge it. With the default respecting `.gitignore`, an `ignore` list in `codegraph.json` would be a second exclude list duplicating what `.gitignore` already says — so it's redundant for the case it was solving. For the inverse need — people who *want* their gitignored nested repos indexed (the "super-repo of clones" layout) — #980 adds the opposite knob, `includeIgnored`.

If there's a real need to exclude a path that *isn't* gitignored from indexing (without touching `.gitignore`), that's a separate, narrower feature worth its own issue — happy to discuss it there.

Genuinely appreciate you taking the time to write the fix and the tests. The regression is fixed in **v1.1.1**, releasing now.

PR #989, opened 2026-06-25
feat: Add support for custom MediatR interfaces

PressXtoChris:
## ✨ What's this?
Adds the ability for C# projects using MediatR to add in custom interfaces alongside `IRequestHandler` and `INotificationHandler` via the codegraph.json configuration file

## 🔍 Why do we want this?
When Codegraph analyses a project that uses MediatR request/notification handlers, it uses a regex match to identify the relevant handlers. If a project is using custom interfaces that derive from `IRequestHandler` (e.g. `ICommandHandler` or `IQueryHandler`), codegraph won't be able to identify the link. 

## 🏗 How is it done?
Expand project-config.ts to add in a new configuration section to codegraph.json: csharp.mediatrHandlerInterfaces, which takes an array of strings. Example:
```json
{
  "csharp": {
    "mediatrHandlerInterfaces": ["ICommandHandler", "IQueryHandler"]
  }
}
```

Update callback-synthesizer.ts's MediatR logic so that instead of the single hardcoded regular expression, it pulls additional valid handler names from the config file. The original values of `IRequestHandler` and `INotificationHandler` are always included automatically.

## 💥 Breaking changes
None


psandeep02:
It looks like you linked to a specific discussion thread within that Pull Request, but the section isn't currently visible in my immediate viewport.

If there is a particular snippet, question, or problem from that part of the PR you'd like to dive into—whether it's about expanding the `codegraph.json` configuration, handling custom MediatR interfaces in C#, or refining regex patterns in the `callback-synthesizer.ts`—just paste it here or let me know what you're looking to solve!

PR #998, opened 2026-06-26
fix(mcp): fall back to tmpdir socket on filesystems without AF_UNIX support

### LINKED ISSUE #997: Daemon fails with ENOTSUP on ExFAT / external volumes (no tmpdir fallback for unsupported socket)
## Summary

When a project lives on an **ExFAT-formatted external volume** (e.g. `/Volumes/zwl/xjp/my-project`), the daemon fails to start because ExFAT does not support Unix domain sockets. The `listen()` call throws `ENOTSUP`.

The existing code in `daemon-paths.ts` already has a `tmpdir` fallback for **long** socket paths (`> 100` chars), but there is no fallback for filesystems that **don't support sockets at all**.

## Error

```
✗ Failed to start server: listen ENOTSUP: operation not supported on socket /Volumes/zwl/xjp/my-project/.codegraph/daemon.sock
```

(`daemon.log` keeps accumulating this error on every daemon spawn attempt.)

## Environment

- **CodeGraph version**: 1.1.1
- **OS**: macOS 26.0.1 (Darwin 25.0.0, Apple Silicon)
- **Filesystem**: ExFAT (external USB drive)
- **Socket path length**: 49 chars (well under the 100-char limit, so the tmpdir fallback is not triggered)

## Root Cause

In `daemon-paths.js`, `getDaemonSocketPath()` only falls back to `os.tmpdir()` when `inProject.length > POSIX_SOCKET_PATH_LIMIT`:

```js
function getDaemonSocketPath(projectRoot) {
    const inProject = path.join(getCodeGraphDir(projectRoot), 'daemon.sock');
    if (inProject.length <= POSIX_SOCKET_PATH_LIMIT)
        return inProject;        // ← ExFAT hits this branch, then listen() fails
    return path.join(os.tmpdir(), `codegraph-${projectHash(projectRoot)}.sock`);
}
```

The daemon's `start()` method (`daemon.js`) catches the `ENOTSUP` from `linkSync` (pid lockfile) and falls back to `renameSync`, but there is **no equivalent fallback** for the socket `listen()` call — it just cleans up and re-throws, killing the daemon.

## Suggested Fix

When `server.listen()` fails with `ENOTSUP` (or `ENOTSOCK` / `EACCES`), retry with a `tmpdir`-based socket path — the same path `getDaemonSocketPath` already computes for the long-path case:

```js
// In Daemon.start(), after the first listen fails:
if (err.code === 'ENOTSUP' || err.code === 'ENOTSOCK') {
    this.socketPath = path.join(os.tmpdir(), `codegraph-${projectHash(projectRoot)}.sock`);
    // retry listen on tmpdir path
}
```

Or proactively detect non-socket-capable filesystems in `getDaemonSocketPath()` itself.

## Impact

- **Daemon** (background auto-sync): broken — no automatic incremental indexing on file changes.
- **MCP tool** (`codegraph_explore`): works fine (in-process, no daemon needed).
- **`prompt-hook`**: works fine (in-process).
- **Manual `codegraph index/sync`**: works fine.

So the functional impact is limited to losing automatic background indexing — users must run `codegraph sync .` manually after significant changes. But the repeated error in `daemon.log` is noisy.

## Workaround

Run `codegraph index .` or `codegraph sync .` manually to refresh the index.

zengwenliang416:
## Summary

Fixes #997.

`getDaemonSocketPath()` now probes whether the target directory supports Unix domain sockets before returning an in-project path. On filesystems that don't support AF_UNIX (ExFAT, NTFS-3G, some FUSE mounts), it transparently falls back to the `os.tmpdir()` hash-based socket — the same path already used for the long-path case.

## Problem

When a project lives on an ExFAT-formatted external volume (common on macOS for USB drives), the daemon fails to start:

```
✗ Failed to start server: listen ENOTSUP: operation not supported on socket /Volumes/drive/project/.codegraph/daemon.sock
```

The existing tmpdir fallback only triggered for paths exceeding 100 characters. Short paths on unsupported filesystems had no fallback.

## Approach

- Added `canSocketInDir(dir)` to `daemon-paths.ts` that spawns a short-lived child process (`execFileSync`) to attempt binding a Unix socket in the target directory
- Result is cached per device number (`fs.statSync().dev`) so the probe runs at most once per mounted filesystem per process lifetime
- Integrated into `getDaemonSocketPath()` alongside the existing length check
- Both daemon and proxy call `getDaemonSocketPath()`, so they automatically agree on the socket path — no changes needed in `daemon.ts` or `index.ts`

## Changes

- **`src/mcp/daemon-paths.ts`**: Added `canSocketInDir()`, `clearSocketSupportCache()` (test helper), updated `getDaemonSocketPath()` guard
- **`__tests__/daemon-socket-probe.test.ts`**: New test file covering probe behavior and cache semantics

## Test plan

- [x] `npm run build` — compiles cleanly
- [x] New tests pass: `canSocketInDir` returns true on tmpdir, caches results, handles nonexistent dirs
- [x] Existing `daemon-bind-failure.test.ts` still passes (no regression)
- [x] Existing `proxy-connect.test.ts` and `daemon-registry.test.ts` still pass
- [x] Manual verification: `getDaemonSocketPath('/Volumes/exfat-drive/project')` returns tmpdir path; local APFS projects still get in-project path

### DIFF
--- a/__tests__/daemon-socket-probe.test.ts
+++ b/__tests__/daemon-socket-probe.test.ts
@@ -1,0 +1,68 @@
+/**
+ * Socket-support probe — issue #997.
+ *
+ * ExFAT, NTFS-3G, and some FUSE-backed volumes don't support AF_UNIX sockets.
+ * `getDaemonSocketPath` must detect this and fall back to `os.tmpdir()` instead
+ * of returning an in-project path that will fail on `listen()`.
+ *
+ * These tests validate `canSocketInDir` on the local tmpdir (which always
+ * supports sockets on any standard OS) and verify the cache is effective.
+ */
+
+import { afterEach, describe, expect, it } from 'vitest';
+import * as os from 'os';
+import * as fs from 'fs';
+import * as path from 'path';
+import {
+  canSocketInDir,
+  clearSocketSupportCache,
+  getDaemonSocketPath,
+} from '../src/mcp/daemon-paths';
+
+afterEach(() => {
+  clearSocketSupportCache();
+});
+
+describe('canSocketInDir (#997)', () => {
+  it.runIf(process.platform !== 'win32')('returns true for a directory on a socket-capable filesystem', () => {
+    const dir = fs.mkdtempSync(path.join(os.tmpdir(), 'cg-sock-probe-'));
+    try {
+      expect(canSocketInDir(dir)).toBe(true);
+    } finally {
+      fs.rmSync(dir, { recursive: true, force: true });
+    }
+  });
+
+  it.runIf(process.platform !== 'win32')('caches the result per device — second call is free', () => {
+    const dir = fs.mkdtempSync(path.join(os.tmpdir(), 'cg-sock-probe-'));
+    try {
+      const first = canSocketInDir(dir);
+      const second = canSocketInDir(dir);
+      expect(first).toBe(second);
+    } finally {
+      fs.rmSync(dir, { recursive: true, force: true });
+    }
+  });
+
+  it('returns true when the directory does not exist (optimistic fallback)', () => {
+    expect(canSocketInDir('/nonexistent-dir-cg-probe')).toBe(true);
+  });
+});
+
+describe('getDaemonSocketPath (#997)', () => {
+  it.runIf(process.platform !== 'win32')('returns in-project path on a socket-capable filesystem', () => {
+    const root = fs.mkdtempSync(path.join(os.tmpdir(), 'cg-path-'));
+    try {
+      const sockPath = getDaemonSocketPath(root);
+      expect(sockPath).toContain('.codegraph');
+      expect(sockPath).toContain('daemon.sock');
+    } finally {
+      fs.rmSync(root, { recursive: true, force: true });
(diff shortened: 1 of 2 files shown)


inth3shadows:
This also covers a WSL case, not just macOS ExFAT — might be worth noting, since it widens the PR's reach.

On WSL2, project trees under `/mnt/c` (the Windows drive, DrvFs) don't support AF_UNIX either:

```
node -e 'require("net").createServer().on("error",e=>console.log(e.code)).listen("/mnt/c/Users/me/x.sock")'
-> ENOTSUP
```

Same `listen ENOTSUP` this PR handles, so the tmpdir fallback would let the daemon start for WSL users working on `/mnt` drives too (a common setup). Validated on WSL2 / Ubuntu.

zengwenliang416:
Thanks, good point. #1022 has now landed on main and covers the broader filesystem-capability case, including WSL2 DrvFs, plus the lockfile hard-link failure that appears before socket binding on ExFAT/FAT. This PR is superseded, so I’ll close it.

PR #1030, opened 2026-06-28
feat(extraction): make MAX_FILE_SIZE configurable via CODEGRAPH_MAX_FILE_SIZE (#1016)

### LINKED ISSUE #1016: Make MAX_FILE_SIZE configurable via environment variable
First off — thank you for building CodeGraph! It's a fantastic tool that makes code comprehension so much more accessible. I've been using it on a few projects and ran into a scenario where the 1 MB file size threshold felt a bit too restrictive, so I wanted to share a suggestion for a small, low-risk improvement.

## The Scenario

I work with projects that include some legitimate source files between 1–5 MB. These aren't minified bundles or vendored blobs; they contain real, meaningful function definitions that would be valuable in the graph.

Currently, these files are silently skipped during indexing (`size_exceeded` warning, no nodes/edges), which means the function graph for the project is incomplete. I completely understand the original design rationale — the 1 MB threshold protects against WASM heap blowup from generated/minified files, and that's a wise default.

## The Suggestion

What if the 1 MB default stayed exactly where it is (it's a great safe default!), but users could **override it via an environment variable** when they know their project has legitimate large source files?

```typescript
const MAX_FILE_SIZE = (() => {
  const envVal = process.env.CODEGRAPH_MAX_FILE_SIZE;
  if (envVal !== undefined) {
    const parsed = parseInt(envVal, 10);
    if (parsed > 0 && !isNaN(parsed)) {
      return parsed;
    }
  }
  return 1024 * 1024;  // 1 MB — unchanged default
})();
```

This keeps the existing behavior 100% intact for everyone who doesn't set the variable, while giving users with large-but-legitimate source files an escape hatch.

## Complementary Measure: Memory Safety for Large Files

If a user overrides the threshold, files above 1 MB will now be parsed — but these files consume significantly more WASM linear memory (WebAssembly spec limitation: can grow but never shrink). Without additional safeguards, parsing many large files in sequence could accumulate heap memory until OOM.

A complementary change would make this safe: **for files exceeding 1 MB, aggressively reclaim WASM memory after each parse**.

```typescript
const LARGE_FILE_THRESHOLD = 1024 * 1024;  // 1 MB — same as original MAX_FILE_SIZE
```

### Bulk indexing path (worker thread)

Recycle the worker **before** parsing each large file, ensuring a fresh WASM heap:

```typescript
const isLargeFile = stats.size > LARGE_FILE_THRESHOLD;
if (isLargeFile) {
  recycleWorker();  // Destroy worker → fresh WASM heap
}
result = await requestParse(filePath, content);
```

### Single-file extraction path (in-process)

Reset the parser **after** extracting each large file:

```typescript
const result = extractFromSource(relativePath, content, language, frameworkNames);
if (stats.size > LARGE_FILE_THRESHOLD) {
  resetParser(language);  // Delete cached parser instance → reclaim WASM heap
}
```

Both `recycleWorker()` and `resetParser()` already exist in the codebase — no new functions needed, just calling them at the right time.

### Required import

```typescript
import { ..., resetParser } from './grammars';
```

## Memory Safety Summary

| Concern                             | Mitigation                                                          |
| ----------------------------------- | ------------------------------------------------------------------- |
| WASM heap growth from large files   | `recycleWorker()` / `resetParser()` after each >1MB parse       |
| Cascading OOM after WASM corruption | Worker crash → respawn (existing mechanism)                        |
| Parse timeout on large files        | Already scales:`PARSE_TIMEOUT_MS + content.length / 100KB × 10s` |
| Worker recycling cost               | Only for >1MB files; normal files unaffected                        |
| User overrides too aggressively     | Tunable via env var; 1 MB default remains the safe baseline         |

## Changes Summary

| File                                           | Change                                                                                                |
| ---------------------------------------------- | ----------------------------------------------------------------------------------------------------- |
| `src/extraction/index.ts`                    | Make `MAX_FILE_SIZE` configurable via `CODEGRAPH_MAX_FILE_SIZE` env var (default unchanged: 1 MB) |
| `src/extraction/index.ts`                    | Add `LARGE_FILE_THRESHOLD = 1MB` for memory-reclamation gating                                      |
| `src/extraction/index.ts`                    | Import `resetParser` from `./grammars`                                                            |
| `src/extraction/index.ts` (bulk path)        | Call `recycleWorker()` before parsing files > `LARGE_FILE_THRESHOLD`                              |
| `src/extraction/index.ts` (single-file path) | Call `resetParser(language)` after extracting files > `LARGE_FILE_THRESHOLD`                      |

maxmilian:
Closes #1016.

## What

The 1 MB file-size skip threshold in `src/extraction/index.ts` was a hardcoded `const`, so repos with legitimately large hand-written sources (1–5 MB) had no way to get them indexed.

This adds a `resolveMaxFileSize()` helper that honours a `CODEGRAPH_MAX_FILE_SIZE` environment variable (in bytes), mirroring the existing `resolve*()` env idioms in `src/mcp/daemon.ts`:

```ts
export function resolveMaxFileSize(): number {
  const raw = process.env.CODEGRAPH_MAX_FILE_SIZE;
  if (raw === undefined || raw === '') return DEFAULT_MAX_FILE_SIZE;
  const parsed = Number(raw);
  if (!Number.isFinite(parsed) || parsed <= 0) return DEFAULT_MAX_FILE_SIZE;
  return Math.floor(parsed);
}
```

Unset / non-numeric / non-positive values fall back to the 1 MB default, so existing behaviour is unchanged unless the var is explicitly set. Both existing size-check sites (single-file and bulk paths) are unchanged — they read the same module `MAX_FILE_SIZE`.

## Why this is the whole change

The issue also suggested complementary WASM memory-safety (recycle the worker after large parses). That guard **already exists** — `recycleWorker()` is already invoked after large-file parses — so making the threshold configurable is safe on its own and needs no lifecycle changes.

## Tests

- New `__tests__/resolve-max-file-size.test.ts` — default fallback, valid override, fractional flooring, and invalid/zero/negative/non-numeric fallback (9 cases, all pass).
- `__tests__/extraction.test.ts` — 377 pass, no regression.
- `tsc --noEmit` — clean.

README updated with the override next to the existing 1 MB documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -599,6 +599,9 @@ What it skips out of the box:
 - **Anything in your `.gitignore`** — honored in git repos via git, and in
   non-git projects by reading `.gitignore` directly (root and nested).
 - **Files larger than 1 MB** — generated bundles, minified JS, vendored blobs.
+  Override the threshold (in bytes) with `CODEGRAPH_MAX_FILE_SIZE` for repos with
+  legitimately large hand-written sources — e.g. `CODEGRAPH_MAX_FILE_SIZE=5242880`
+  for 5 MB. Invalid or non-positive values fall back to the 1 MB default.
 
 To keep something else out, add it to `.gitignore`. To pull a default-excluded
 directory back **in** (say you really do want a vendored dependency indexed),
--- a/__tests__/resolve-max-file-size.test.ts
+++ b/__tests__/resolve-max-file-size.test.ts
@@ -1,0 +1,34 @@
+import { afterEach, describe, expect, it } from 'vitest';
+import { resolveMaxFileSize } from '../src/extraction';
+
+const DEFAULT = 1024 * 1024;
+const ENV = 'CODEGRAPH_MAX_FILE_SIZE';
+
+describe('resolveMaxFileSize', () => {
+  afterEach(() => {
+    delete process.env[ENV];
+  });
+
+  it('falls back to the 1 MB default when the env var is unset', () => {
+    delete process.env[ENV];
+    expect(resolveMaxFileSize()).toBe(DEFAULT);
+  });
+
+  it('honours a valid positive override', () => {
+    process.env[ENV] = String(5 * 1024 * 1024);
+    expect(resolveMaxFileSize()).toBe(5 * 1024 * 1024);
+  });
+
+  it('floors fractional byte counts', () => {
+    process.env[ENV] = '2097152.9';
+    expect(resolveMaxFileSize()).toBe(2 * 1024 * 1024);
+  });
+
+  it.each(['', 'not-a-number', '0', '-1', 'NaN', 'Infinity'])(
+    'falls back to the default for invalid value %j',
+    (raw) => {
+      process.env[ENV] = raw;
+      expect(resolveMaxFileSize()).toBe(DEFAULT);
+    },
+  );
+});
--- a/src/extraction/index.ts
+++ b/src/extraction/index.ts
@@ -114,7 +114,23 @@ export function hashContent(content: string): string {
  * vendored blobs blow the WASM heap and the worker-recycle budget for no useful
  * symbols. 1 MB covers essentially all hand-written source.
  */
-const MAX_FILE_SIZE = 1024 * 1024;
+const DEFAULT_MAX_FILE_SIZE = 1024 * 1024;
+
+/**
+ * Resolve the max-file-size threshold (bytes), allowing an override via the
+ * `CODEGRAPH_MAX_FILE_SIZE` environment variable for repos with legitimately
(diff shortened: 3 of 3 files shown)


maxmilian:
Closing to keep my open-PR footprint here honest — not because I think the change is wrong.

I have seven PRs open in this repo and only one merged (#663). This one has had no review activity since it was opened, and I'd rather leave a short queue of things with evidence behind them than a long one you have to page through. The two I'm keeping active have verified repros on current `main` (#594's import-resolution gap, and #597's `(AX_VOID)` case that survives the #1102 salvage).

Branch stays on my fork and this is trivially reopenable if it's ever useful.

PR #1036, opened 2026-06-28
fix(cpp): emit instantiates edge for stack-allocation constructor syntax

### LINKED ISSUE #1035: C++ stack-allocation syntax not recorded as 'instantiates' edge
## Summary

codegraph does not record an `instantiates` edge when a C++ function
constructs an object via stack-allocation syntax (`Type obj(args)`).
Heap allocation (`Type* obj = new Type(args)`) correctly produces the edge.

## Reproduction

```cpp
// stack allocation — no instantiates edge recorded
int runService(int a, int b) {
    Calculator calc(0);
    return calc.add(a, b);
}

// heap allocation — instantiates edge IS recorded
int runServiceHeap(int a, int b) {
    Calculator* calc = new Calculator(0);
    int result = calc->add(a, b);
    delete calc;
    return result;
}
```

After `codegraph init`:

```sql
SELECT s.name, e.kind, t.name FROM edges e
JOIN nodes s ON e.source=s.id JOIN nodes t ON e.target=t.id
WHERE e.kind='instantiates';

-- result: only runServiceHeap|instantiates|Calculator
-- runService has no instantiates edge despite constructing Calculator
```

## Expected

Both `Calculator calc(0)` and `new Calculator(0)` should produce a
`runService --[instantiates]--> Calculator` edge, since both call the constructor.

Dshuishui:
## Summary

C++ stack-allocation syntax (`ClassName var(args)`) does not produce an
`instantiates` edge. Heap allocation (`new ClassName(args)`) works correctly
because `new_expression` is in `INSTANTIATION_KINDS`, but the stack form is
parsed by tree-sitter as a `declaration` node, which is never matched.

This fix adds a branch inside `visitFunctionBody`'s body walker that detects
the pattern: a `declaration` node whose `type` field resolves to a class name
(via the existing `normalizeCppReturnType` primitive filter) and whose
`init_declarator` child contains an `argument_list`. When matched, an
`instantiates` unresolved reference is pushed — the same shape that
`extractInstantiation` produces for `new_expression`.

## What's covered

- `Calculator calc(0)` → `caller --[instantiates]--> Calculator` ✓
- `int x(5)` → skipped (`int` is in `CPP_NON_CLASS_RETURN`) ✓
- `Calculator calc = other` → skipped (no `argument_list` in declarator) ✓

## Known limitations

- **Default constructor** (`Calculator calc;`) — no `init_declarator` child,
  so no edge is emitted. Can be addressed in a follow-up.
- **Brace initialization** (`Calculator calc{0}`) — uses `initializer_list`
  rather than `argument_list`, not covered here.

Closes #1035

### DIFF
--- a/src/extraction/tree-sitter.ts
+++ b/src/extraction/tree-sitter.ts
@@ -28,6 +28,7 @@ import { AstroExtractor } from './astro-extractor';
 import { DfmExtractor } from './dfm-extractor';
 import { VueExtractor } from './vue-extractor';
 import { MyBatisExtractor } from './mybatis-extractor';
+import { normalizeCppReturnType } from './languages/c-cpp';
 import {
   getAllFrameworkResolvers,
   getApplicableFrameworks,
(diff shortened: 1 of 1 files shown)


Dshuishui:
Closing as this is superseded by #1049, which covers a broader set of cases. Thanks for the quick fix!

Issue #237, opened 2026-05-21
C# classes using primary constructor syntax are sometimes not indexed


d3ara1n:
## Problem

When a C# class uses the C# 12 primary constructor syntax (`public class Foo(Bar bar)`), the class and all its methods may be completely absent from the CodeGraph index. This breaks search, node lookup, call chain analysis, and context queries for those symbols.

## Reproduction

Tested with CodeGraph on a .NET 10 project (298 C# files, 3073 nodes indexed). The `classTypes: ['class_declaration']` in `csharp.ts` should match primary-constructor classes, but results are inconsistent:

| Class | Syntax | Indexed? |
|---|---|---|
| `DataService` | `public class DataService(IMemoryCache cache, ...)` | ✅ Yes |
| `InstanceService` | `public class InstanceService(InstanceManager m, ProfileManager p, ...)` | ❌ No |
| `UpdateService` | `public partial class UpdateService(...) : ILifetimeService` | ❌ No (only nested `IOptions` found) |

`InstanceService` has 2 public methods (`DeployAndLaunchAsync`, `Deploy`) — neither appears as a node. The file itself is indexed (imports are extracted), but the class declaration and its methods are skipped entirely.

## Downstream Impact

- `codegraph_search("InstanceService", kind="class")` → empty
- `codegraph_node("InstanceService")` → falls through to an `import` node instead of the class
- `codegraph_callees("PlayAsync")` → misses the call to `DeployAndLaunchAsync` (because the target symbol doesn't exist)
- `codegraph_context("Instance launch workflow")` → returns irrelevant URL-launcher helpers instead of the game launch logic

## Hypothesis

The inconsistency suggests the issue may not be solely about primary constructors (since `DataService` is indexed). Possible factors:

- `partial` modifier combined with primary constructor
- Inheritance clause (`: ILifetimeService`) after the parameter list
- Parameter types from external projects / git submodules causing tree-sitter parse errors
- Number of parameters or line count in the parameter list

It would be worth testing tree-sitter-c_sharp's AST output for these patterns to see if `class_declaration` nodes are produced correctly.

colbymchenry:
[...]

PR #1070, opened 2026-06-30
fix(cpp): recover export-macro-annotated classes instead of dropping them (#1061)

### LINKED ISSUE #1061: C++: classes annotated with a *_API export macro (UE-style) are dropped — no node, no extends edge (#946 follow-up)
### Summary

For C++ classes written in the Unreal-Engine style

```cpp
class MYMODULE_API UMyComponent : public UActorComponent { /* ... */ };
```

the leading export macro (MYMODULE_API, which expands to __declspec(dllexport/import) or nothing) makes tree-sitter misparse the declaration as a function_definition. The current handling (isMacroMisparsedTypeDecl, #946) drops the node entirely. Consequences:

The class node is missing from the index.
Its extends / base-class edge is never created.
Therefore "find subclasses / type hierarchy / impact through inheritance" return nothing for these classes.
This affects effectively every gameplay C++ class in an Unreal Engine project, since XXX_API between class and the type name is the standard UE pattern.

Repro
```cpp
// Base.h
class ENGINE_API UActorComponent { };

// Derived.h
class MYGAME_API UMyComponent : public UActorComponent { };
```

Index, then query the type hierarchy / subclasses of UActorComponent, or extends of UMyComponent. The edge is absent. (Removing MYGAME_API makes it resolve correctly.)

Root cause
tree-sitter (no macro knowledge) reads class MYGAME_API as a bodyless elaborated type specifier and the rest as a function. isMacroMisparsedTypeDecl correctly identifies this but, per its comment, treats the body as unrecoverable and drops the whole node — so the recoverable parts (class name + base clause) are lost too.

Proposed fix
Use the existing preParse extractor hook (already used by C# via blankCsharpPreprocessorDirectives, #237) to blank the export macro only where it triggers the misparse — immediately after class/struct — with equal-length spaces so every byte offset (and thus every line/column) is preserved:

```TypeScript
/** Blank a `class/struct EXPORT_MACRO Name` annotation (UE `*_API`, etc.) that
 *  tree-sitter misparses (#946). Equal-length spaces keep all offsets exact. */
export function blankCppExportMacros(source: string): string {
  if (!source.includes('_API')) return source;
  return source.replace(
    /\b(class|struct)(\s+)([A-Z][A-Z0-9_]*_API)\b/g,
    (_m, kw, ws, macro) => kw + ws + ' '.repeat(macro.length),
  );
}

// cppExtractor:
preParse: blankCppExportMacros,
```

After blanking, the declaration parses as a normal class_specifier with a base_class_clause, so the existing extends extraction emits the edge and the class node (and its members) are indexed normally — no offset shift, and the isMacroMisparsedTypeDecl drop path is no longer hit for these cases.

This is intentionally scoped to the class/struct header position so that ordinary references to *_API constants/macros elsewhere (e.g. int x = SOME_API;) are untouched.

colbymchenry:
Closes #1061. Follow-up to #946.

## Problem

A C++ class annotated with an export/visibility macro between `class`/`struct` and the type name makes tree-sitter (no macro expansion) read `class MACRO` as an elaborated type and the whole declaration as a `function_definition` named after the class. This is the standard Unreal-Engine pattern (`class MYGAME_API UMyComponent : public UActorComponent`), and the equivalent `*_EXPORT` / `*_ABI` macros in Qt, Boost, LLVM, etc. The **leading macro alone** triggers it — a base clause isn't required.

#946 caught the resulting phantom function and **dropped** it, but that also discarded the recoverable class name, members, and base-class edge. So the class never entered the graph and its `extends` edge was never created — making *find subclasses*, *type hierarchy*, and *impact-through-inheritance* return nothing for effectively every gameplay class in a UE project.

## Fix

Add `blankCppExportMacros` wired as `cppExtractor.preParse`. It blanks the macro with **equal-length spaces** before parsing (offset-preserving, exactly like C#'s `blankCsharpPreprocessorDirectives` from #237), so the declaration parses as a normal `class_specifier` and the existing extraction emits the node, members, and `extends` edge.

Generalized past UE's `*_API` to **any all-caps export macro** (covers Qt/Boost `*_EXPORT`, LLVM `*_ABI`, bare `API`, and #946's own `MAPCORE_EXPORT`), with two false-positive guards:

- **`[:{]` definition-guard** — only fires when a base clause or body follows, so elaborated-type variable declarations (`struct FOO var;`, `class FOO obj = …`, which end `;`/`=`/`[`) are left untouched. Scoped to `cppExtractor` only, so C's heavier use of `struct TAG var;` never reaches it.
- **must-be-followed-by-the-name** — an all-caps class *name* with a base (`class FOO : public Base`) is left alone; only a token *before* the name is blanked.

The #946 drop (`isMacroMisparsedTypeDecl`) stays as the fallback for any residual misparse the blanking doesn't catch — still no phantom function.

## Validation

- **Tests:** strengthened the existing `#946` block to assert *recovery* (not just de-phantoming) and added a `#1061` block (UE repro, offset-preservation, and the all-caps-name / var-decl / value-ref false-positive controls). Full suite green: **1848 passed, 4 skipped**.
- **Real repo — google/leveldb** (`LEVELDB_EXPORT`, 134 files), baseline vs. fix:

  | metric | baseline | fix | Δ |
  |---|---|---|---|
  | `class`+`struct` nodes | 266 | 293 | **+27** |
  | `extends` edges | 292 | 359 | **+67** |
  | `function` nodes (phantoms) | 588 | 513 | **−75** |

  Every export-macro real definition flipped `function`→`class` (`Env`, `Status`, `DB`, `Slice`, `Cache`, …) and `EnvWrapper extends Env` went absent→present. (Baseline's "surviving" class nodes were thin forward-declaration stubs in other headers; the real definitions were all phantom functions.)

Thanks @luoyxy for the detailed report and proposed fix.


ryancu7:
This is great. But for us, our (admittedly somewhat unusual and Windows-centric) coding style doesn't have these macros in all caps. 

E.g. we have 
`class Dll_libraryname libC_ClassName`

Issue #925, opened 2026-06-19
MCP server keeps serving a deleted DB inode after .codegraph is removed+recreated at the same path (git worktree recreate); sync can't refresh it, only a server restart does


polaris-arch:
## Summary

A long-lived `codegraph serve --mcp` server opens `.codegraph/codegraph.db` and holds the file descriptor for the session's lifetime. If the project's `.codegraph` dir is removed and recreated **at the same path** while that server is running — most easily via `git worktree remove <path>` then `git worktree add <same path>` + `codegraph init` — the server keeps reading the **old, now-unlinked inode**, while the CLI (`sync`/`query`/`status`) and any newly-spawned server read the **new inode**. The MCP server then serves a stale snapshot indefinitely. `codegraph sync` does **not** fix it; only restarting the server does.

## Environment

- codegraph 0.9.9, Linux x64, MCP stdio (`codegraph serve --mcp`), one server per Claude Code session.

## Repro

1. In a git worktree, use codegraph via MCP so the server opens `.codegraph/codegraph.db` (fd -> inode A).
2. Rename a symbol on disk (e.g. `fooOld` -> `fooNew`); commit.
3. `git worktree remove <path>` (or `rm -rf`) — unlinks `.codegraph/*`; the server still holds inode A.
4. `git worktree add <same path> <branch>` + `codegraph init <path>` — creates `.codegraph/codegraph.db` as a **new inode B**.
5. `codegraph sync` (reports e.g. `Added: 1 — 0 nodes`).
6. MCP `codegraph_search fooNew` -> **not found**; `codegraph_search fooOld` -> still reported as a live function at its old line. CLI `codegraph query fooNew` -> **correct** (found). The two diverge.

## Root cause (deleted-but-open inode)

The MCP server process is reading an unlinked inode:

```
$ ls -l /proc/<server_pid>/fd | grep codegraph.db
... 24 -> /home/.../<worktree>/.codegraph/codegraph.db (deleted)
... 25 -> .../codegraph.db-wal (deleted)
... 26 -> .../codegraph.db-shm (deleted)
```

`(deleted)` = the inode the server reads was `unlink()`ed but kept alive by the open fd (standard Unix semantics). `codegraph init`/`sync` write a **different** inode at the same path. MCP queries (fd-based) and CLI / new sessions (path-based) therefore see different databases. Killing the server (so a fresh server opens by path) restores correctness.

## Impact

[...]

Issue #850, opened 2026-06-13
codegraph cause 100% CPU


songhlc:
# `codegraph serve --mcp` worker pegs one CPU core indefinitely (unrecoverable)

## Summary

When the `codegraph serve --mcp` long-running worker (the `--liftoff-only` child) hits an uncaught exception, it does not crash. Instead it enters a tight loop in V8 stack-trace formatting and pegs one CPU core at ~100% **forever**. The process must be killed manually; there is no self-recovery.

## Environment

| Item | Value |
|---|---|
| codegraph version | `0.9.5` (npm: `@colbymchenry/codegraph@0.9.5`, native: `@colbymchenry/codegraph-darwin-arm64@0.9.5`) |
| Install path | `~/.nvm/versions/node/v20.19.2/lib/node_modules/@colbymchenry/codegraph/...` |
| Node.js | `v20.19.2` |
| OS | macOS 26.3.1 (Darwin 25.3.1), arm64 |
| Invocation | `codegraph serve --mcp` (started as a launchd-managed daemon, parent PID 1) |
| Connected client | Claude Code via the codegraph MCP server (a separate, *healthy* worker is also spawned per session and does not reproduce) |

A second, short-lived worker spawned by the Claude Code session itself (`v0.9.4` at `~/.codegraph/versions/v0.9.4/`) did **not** exhibit the issue during the same window.

## Symptoms

- PID 62394 (the `codegraph serve --mcp` child) was at **99.1% CPU on a single core**, RSS 242 MB, for **>5 hours** with no movement.
- `ps` showed the process unresponsive to signals other than SIGKILL.
- Parent daemon (PID 62392) was healthy and at 0% CPU; the issue was isolated to the child.
- No log output was captured (worker was started with `--liftoff-only`; stdout was not redirected anywhere visible).

## Root-cause analysis (from `sample 62394 3`)

Full sample saved to `/tmp/codegraph-62394.txt` on reporter's machine (503 KB). Hot path on the main thread (1909 samples over 3 s = 100%):

```
TriggerUncaughtException                                  1474
  InspectorConsoleCall                                     1430
    Accessors::ErrorStackGetter                             592
      ErrorUtils::GetFormattedStack                         585
        ErrorUtils::FormatStackTrace                        530
          PrepareStackTraceCallback  (JS-side Error.prepareStackTrace)   527
            ArrayPrototypeJoin                              471
              Builtin_CallSitePrototypeToString             447
[...]

PR #1073, opened 2026-06-30
fix(index): recover a poisoned/oversized index by recreating the DB instead of row-DELETE (#1067)

### LINKED ISSUE #1067: v1.1.4: existing oversized index from ignored gitlinks cannot be rebuilt; index watchdog still kills command
## Summary

After upgrading to CodeGraph `v1.1.4`, the scanner fix from #1065 appears to work for a fresh index, but an existing oversized `.codegraph` state produced by earlier versions still cannot recover via `codegraph index`.

In the affected repo, `codegraph index --verbose` prints only:

```text
┌  Indexing project
```

Then, after about 60 seconds:

```text
[CodeGraph] Main thread unresponsive for ~60s — killing the wedged process so a fresh one can start (#850). Disable with CODEGRAPH_NO_WATCHDOG=1.
```

Exit code is `137`.

This is specifically problematic in the Doctorine repo because previous CodeGraph versions indexed a gitignored benchmark corpus of embedded repos. Other/fresh checkouts now succeed on `v1.1.4`.

## Environment

- CodeGraph: `1.1.4`
- OS: macOS / Darwin arm64
- Repo shape: private TypeScript monorepo with a local benchmark/reference corpus of public cloned repositories under `benchmark/repos/`
- `benchmark/repos/` entries are tracked gitlinks (`160000`) and also physically exist as nested git repos on disk
- No private source content is included in this report

## Relevant ignore/gitlink shape

Root `.gitignore` contains:

```gitignore
benchmark/corpus/
benchmark/repos/
benchmark/**/node_modules/
```

Gitlink count:

```text
git ls-files -s benchmark/repos | awk '$1 == 160000 {c++} END {print c+0}'
# 19
```

Git agrees the path is ignored when checking as an ignored working-tree path:

```text
git check-ignore --no-index -v benchmark/repos benchmark/repos/fern benchmark/repos/RapiDoc benchmark/repos/swagger-ui
.gitignore:82:benchmark/repos/ benchmark/repos
.gitignore:82:benchmark/repos/ benchmark/repos/fern
.gitignore:82:benchmark/repos/ benchmark/repos/RapiDoc
.gitignore:82:benchmark/repos/ benchmark/repos/swagger-ui
```

## What is fixed in v1.1.4

Calling the installed `v1.1.4` scanner APIs directly from the affected working tree now looks correct:

```json
{
  "scanDirectory": {
    "count": 2256,
    "benchmarkRepos": 0,
    "benchmark": 5,
    "sampleBenchmarkRepos": []
  },
  "discoverEmbeddedRepoRoots": {
    "count": 0,
    "benchmarkRepos": 0,
    "sample": []
  },
  "buildScopeIgnore": {
    "benchmarkReposDir": true,
    "benchmarkReposChildDir": true,
    "benchmarkReposFile": true
  }
}
```

A fresh `v1.1.4` init on a temp copy of the same working tree, with the ignored benchmark repos physically present but no old `.codegraph`, also succeeds:

```text
◆  Indexed 2,256 files
●  25,338 nodes, 87,893 edges in 7.6s
└  Done
```

Running `codegraph index --verbose` again in that fresh temp copy also succeeds:

```text
◆  Indexed 2,256 files
●  25,338 nodes, 87,893 edges in 9.0s
└  Done
```

So the v1.1.4 scanner fix is working for fresh state.

## What is still broken

The original repo still has the old oversized graph from the pre-fix scan:

```text
codegraph status
Files:   140,701
Nodes:   1,595,121
Edges:   1,776,928
DB Size: 4929.61 MB
Backend: node:sqlite — built-in (full WAL)
Journal: wal
```

SQLite file sizes in `.codegraph/`:

```text
codegraph.db      4.8G
codegraph.db-wal  4.0G
codegraph.db-shm  7.9M
```

The indexed file distribution confirms the old bad graph is dominated by the ignored corpus:

```text
benchmark       138,450 files, 483.9 MiB source bytes
packages          1,053 files,   4.3 MiB
apps                824 files,   3.9 MiB
services            211 files,   1.1 MiB
```

More specifically:

```text
files where path like 'benchmark/repos/%' = 138,445
```

Top embedded repo contribution:

```text
benchmark/repos/fern        125,741 files
benchmark/repos/scalar        5,321 files
benchmark/repos/redocly-cli    1,665 files
benchmark/repos/docusaurus     1,277 files
benchmark/repos/zudoku           830 files
benchmark/repos/swagger-ui       822 files
```

The old graph metadata says it was built with an older version:

```text
indexed_with_version|1.1.1
indexed_with_extraction_version|24
```

On this state, `codegraph index --verbose` on `v1.1.4` never reaches the visible `Phase: scanning` output. It just prints `Indexing project` and then the watchdog kills it after about 60 seconds.

I also stopped stale `codegraph.js serve --mcp` daemons from older versions and retried. The result was unchanged:

```text
┌  Indexing project
[CodeGraph] Main thread unresponsive for ~60s — killing the wedged process so a fresh one can start (#850). Disable with CODEGRAPH_NO_WATCHDOG=1.
```

## Expected behavior

Because `codegraph index` is documented as rebuilding the full index from scratch, it should recover from an old bad index state after upgrading to `v1.1.4`.

Reasonable expected outcomes would be either:

1. `codegraph index` safely discards/replaces the old oversized DB and produces the fresh `2,256` file graph, or
2. it detects that the existing DB/WAL is too large or stale and prints a clear instruction such as removing/reinitializing `.codegraph`, rather than wedging until the watchdog kills it.

## Actual behavior

Fresh `v1.1.4` state works. Existing oversized state from the pre-fix gitlink scan does not recover with `codegraph index`; it still hangs/crashes in this repo.

## Why this is a separate follow-up from #1065

#1065's scanner/root-cause fix appears successful in `v1.1.4` for fresh state. This issue is about the recovery path for users who already have a large poisoned `.codegraph` DB/WAL from `v1.1.1`/`v1.1.3` era indexing. The advertised command says it rebuilds from scratch, but in practice the old DB state can still wedge the command before scanning starts.

colbymchenry:
## Summary

Follow-up to #1065/#1066 (same reporter). Those fixed *new* indexes from sweeping in an ignored gitlink corpus, but a project that had **already** built the multi-GB graph before upgrading still couldn't recover. `codegraph index` printed only `┌ Indexing project`, then the #850 liveness watchdog SIGKILLed it (exit `137`) ~60s later — before scanning even started.

### Root cause (not the scanner)

`index` cleared the old graph with a synchronous `DELETE FROM nodes/edges/files`. The `nodes` table carries an FTS5 `AFTER DELETE` trigger, so deleting ~1.6M rows fires ~1.6M FTS delete-markers — **O(rows)**, and it *grows* the WAL further before it can finish. A deterministic probe puts the DELETE-clear at **20.4s on 1.5M synthetic nodes** (WAL 1.16→2.14GB); at the report's denser ~2.6KB/node WAL that crosses the 60s main-thread watchdog. `open()` was never the wedge (the intro prints after it).

### Fix

`codegraph index` is documented as "same result as a fresh init" — so make it one. Instead of opening the old DB and DELETE-ing every row, **discard the database files and re-initialize**:

- **`src/db/index.ts`** — `removeDatabaseFiles(dbPath)` unlinks `codegraph.db` + its `-wal`/`-shm` sidecars (O(1) regardless of size; sidecars best-effort).
- **`src/index.ts`** — `CodeGraph.recreate(projectRoot)` discards the files and returns a fresh, empty instance. Never opens or migrates the poisoned DB. POSIX unlinks an open file fine (a live daemon heals via `reopenIfReplaced`, #925); a Windows file lock becomes an actionable *"stop the daemon / remove `.codegraph`"* error.
- **`src/bin/codegraph.ts`** — `index` now calls `recreate()` instead of `open()` + `clear()`; both `clear()` calls dropped. The public `clear()` API is unchanged.

This also reclaims the disk the bloated `db`/`-wal` were holding.

## Validation

- **Deterministic probe:** DELETE-clear O(rows) (20.4s/1.5M) vs recreate O(1) (~1.3s).
- **End-to-end through the built binary:** poisoned a real `.codegraph` to **800K nodes / 419MB**, then ran `codegraph index` → recovered in **0.3s**, no wedge, correct 6-node graph, disk reclaimed.
- **Tests:** new `recreate`/`removeDatabaseFiles` units (`foundation.test.ts`) + 2 CLI regression tests (`index-command.test.ts`: recovers stale content, and proves the file is recreated via inode swap on POSIX). Existing #874 index tests stay green. Full suite **1858 pass** (only the known-flaky #662 daemon test, green on re-run).
- CHANGELOG `[Unreleased] → Fixes` entry added.

Closes #1067.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixes
+
+- `codegraph index` can now rebuild an existing oversized index from an older version, instead of hanging until the watchdog kills it. The previous fix (#1065) stopped *new* indexes from sweeping in a gitignored corpus of nested repos, but a project that had already built the multi-gigabyte graph before upgrading couldn't recover: `codegraph index` is meant to rebuild from scratch, yet it cleared the old graph by deleting every row one at a time, and on a graph of well over a million symbols that took longer than the 60-second responsiveness watchdog allows — so the command was killed before indexing even started, leaving the bad index in place. A full re-index now discards the old database outright and starts fresh, which is near-instant regardless of the old size and also frees the disk the bloated database was holding. Thanks @AriaShishegaran for the detailed follow-up report. (#1067)
 
 ## [1.1.5] - 2026-06-30
 
--- a/__tests__/foundation.test.ts
+++ b/__tests__/foundation.test.ts
@@ -11,7 +11,7 @@ import * as os from 'os';
 import { CodeGraph } from '../src';
 import { Node, Edge } from '../src/types';
 import { isInitialized, getCodeGraphDir, validateDirectory, codeGraphDirName, isCodeGraphDataDir } from '../src/directory';
-import { DatabaseConnection, getDatabasePath } from '../src/db';
+import { DatabaseConnection, getDatabasePath, removeDatabaseFiles } from '../src/db';
 
 // Create a temporary directory for each test
 function createTempDir(): string {
@@ -23,6 +23,13 @@ function cleanupTempDir(dir: string): void {
   if (fs.existsSync(dir)) {
     fs.rmSync(dir, { recursive: true, force: true });
   }
+}
+
+/** Normalize a PRAGMA read across return shapes (array | object | scalar). */
+function pragmaValue(raw: unknown, key: string): unknown {
+  const row = Array.isArray(raw) ? raw[0] : raw;
+  if (row !== null && typeof row === 'object') return (row as Record<string, unknown>)[key];
+  return row;
 }
 
 describe('CodeGraph Foundation', () => {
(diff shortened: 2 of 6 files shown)


colbymchenry:
### Cross-platform validation (macOS · Linux · Windows)

Validated the fix on all three platforms — `removeDatabaseFiles`/`recreate` touch file-deletion semantics that differ across OSes, so this isn't macOS-only.

| Platform | How | `foundation` + `index-command` | E2E recovery (real binary) |
|---|---|---|---|
| **macOS** (arm64, Node 22) | local | 52/52 ✓ | 800K-node / 419 MB poisoned DB → recovered in **0.3s**, no wedge ✓ |
| **Linux** (`node:22-bookworm`, `docker run --init`) | Docker | 52/52 ✓ | 400K-node / 208 MB poisoned DB → recovered in **0.3s**, no wedge ✓ |
| **Windows 11** (arm64, Node 24) | Parallels VM over SSH | 52/52 ✓ | (covered by the CLI recovery test) |

The full suite is green on macOS (1858 pass; only the known-flaky `#662` daemon test, green on re-run).

**One fix during validation:** the first cut proved "recreated, not cleared in place" via an inode-equality check, which false-failed on Linux — ext4/overlayfs **recycle the inode number** after unlink+recreate. Replaced it with a `PRAGMA user_version` header sentinel: it survives an in-place `DELETE`-clear but not a from-scratch recreate, is filesystem-independent, and now runs on Windows too (the inode check was previously skipped there). Folded into the same commit.

PR #1082, opened 2026-06-30
fix(mcp): dedupe one repo's two path spellings onto one DB connection (#1057)

### LINKED ISSUE #1057: v1.1.3 Codex MCP on WSL /mnt can open same repo via upper/lowercase paths, corrupting .codegraph despite CODEGRAPH_NO_DAEMON
## Summary

On `@colbymchenry/codegraph@1.1.3`, Codex MCP still becomes unstable on WSL2 `/mnt/d` even with the recommended workaround from #974:

```toml
[mcp_servers.codegraph]
command = "env"
args = ["CODEGRAPH_NO_DAEMON=1", "codegraph", "serve", "--mcp", "--no-watch"]
```

The new distinguishing signal is that a single Codex parent process spawned two CodeGraph MCP server processes for the same physical repository with different path casing:

```text
Codex parent:
3608352 .../bin/codex
parent cwd=/mnt/d/02_SoftwareCodes/RCSS.Studio/Tekla/TeklaDrawingHelper_TeklaInternal

CodeGraph MCP children of the same parent:
3608757 cwd=/mnt/d/02_SoftwareCodes/RCSS.Studio/Tekla/TeklaDrawingHelper_TeklaInternal
3609744 cwd=/mnt/d/02_softwarecodes/rcss.studio/tekla/tekladrawinghelper_teklainternal
```

Both paths point to the same `.codegraph/codegraph.db` inode on DrvFs:

```text
upper=70:6192449487701361 /mnt/d/02_SoftwareCodes/RCSS.Studio/Tekla/TeklaDrawingHelper_TeklaInternal/.codegraph/codegraph.db
lower=70:6192449487701361 /mnt/d/02_softwarecodes/rcss.studio/tekla/tekladrawinghelper_teklainternal/.codegraph/codegraph.db
```

After this, CLI `codegraph status` reported:

```text
Failed to get status: database disk image is malformed
```

Killing the repo-matching CodeGraph MCP processes made CLI `status`/`sync` recover, but the Codex MCP tool call then failed with:

```text
Error: tool call error: tool call failed for `codegraph/codegraph_explore`
Caused by:
    Transport closed
```

## Environment

- CodeGraph: `1.1.3`
- Node: `v22.19.0`
- Client: OpenAI Codex CLI
- OS: WSL2 Ubuntu on Windows, project on `/mnt/d` DrvFs
- Project path used by shell / parent process:
  `/mnt/d/02_SoftwareCodes/RCSS.Studio/Tekla/TeklaDrawingHelper_TeklaInternal`
- MCP config:

```toml
[mcp_servers.codegraph]
command = "env"
args = ["CODEGRAPH_NO_DAEMON=1", "codegraph", "serve", "--mcp", "--no-watch"]
```

## Observed process state

`pwd` and `pwd -P` from the active shell both showed the uppercase canonical-looking path:

```text
/mnt/d/02_SoftwareCodes/RCSS.Studio/Tekla/TeklaDrawingHelper_TeklaInternal
/mnt/d/02_SoftwareCodes/RCSS.Studio/Tekla/TeklaDrawingHelper_TeklaInternal
```

The Codex parent process also had the uppercase cwd:

```text
3608352 3608345 Mon Jun 29 15:39:36 2026 .../bin/codex
parent cwd=/mnt/d/02_SoftwareCodes/RCSS.Studio/Tekla/TeklaDrawingHelper_TeklaInternal
```

But the same parent spawned two CodeGraph MCP server processes:

```text
PID 3608757
ppid=3608352
cwd=/mnt/d/02_SoftwareCodes/RCSS.Studio/Tekla/TeklaDrawingHelper_TeklaInternal
env=CODEGRAPH_NO_DAEMON=1
cmdline=node .../bin/codegraph serve --mcp --no-watch

PID 3609744
ppid=3608352
cwd=/mnt/d/02_softwarecodes/rcss.studio/tekla/tekladrawinghelper_teklainternal
env=CODEGRAPH_NO_DAEMON=1
cmdline=node .../bin/codegraph serve --mcp --no-watch
```

The Codex MCP tool call also carried the lower-case project path:

```text
codegraph.codegraph_explore({
  "projectPath": "/mnt/d/02_softwarecodes/rcss.studio/tekla/tekladrawinghelper_teklainternal",
  "query": "RunCurrentVersionSwitch CurrentVersionSwitchService ExecuteBatch ValidateBatchPreconditions RecordBatchNotExecuted DrawingTransferManagerForm CanExecuteBatchVersionSwitch DrawingTransferPlanRecord IsExecutableRecord",
  "maxFiles": 5
})
```

## Daemon-related observation despite `CODEGRAPH_NO_DAEMON=1`

Even with `CODEGRAPH_NO_DAEMON=1` in the MCP server environment, `.codegraph/daemon.pid` contained a relocated daemon:

```json
{
  "pid": 3607983,
  "version": "1.1.3",
  "socketPath": "/tmp/codegraph-befc5052950cc73f.sock",
  "startedAt": 1782718772179
}
```

Process:

```text
PID 3607983
cwd=/mnt/d/02_SoftwareCodes/RCSS.Studio/Tekla/TeklaDrawingHelper_TeklaInternal
env=CODEGRAPH_DAEMON_INTERNAL=1
cmdline=.../codegraph.js serve --mcp --path /mnt/d/02_SoftwareCodes/RCSS.Studio/Tekla/TeklaDrawingHelper_TeklaInternal
```

`.codegraph/daemon.log` included:

```text
[CodeGraph daemon] Socket /mnt/d/02_SoftwareCodes/RCSS.Studio/Tekla/TeklaDrawingHelper_TeklaInternal/.codegraph/daemon.sock unusable (ENOTSUP); relocating to /tmp/codegraph-befc5052950cc73f.sock.
[CodeGraph daemon] Listening on /tmp/codegraph-befc5052950cc73f.sock (pid 3607983, v1.1.3). Idle timeout 300000ms.
[CodeGraph MCP] File watcher disabled — project is on a WSL2 /mnt/ drive, where recursive fs.watch is too slow to be reliable. The graph will not auto-update; run `codegraph sync` (or install the git sync hooks via `codegraph init`) to refresh.
```

This may be expected under some path, but it was surprising because the user-facing MCP process had `CODEGRAPH_NO_DAEMON=1`.

## Symptoms

1. `codegraph sync` was run.
2. `codegraph status` then failed:

```text
Failed to get status: database disk image is malformed
```

3. Multiple CodeGraph MCP processes existed, including two for the same physical repo via upper/lower casing.
4. After killing repo-matching processes, CLI `status` recovered and `sync` completed.
5. The subsequent Codex MCP call failed:

```text
Error: tool call error: tool call failed for `codegraph/codegraph_explore`
Caused by:
    Transport closed
```

## Why this seems distinct from #974

#974 was on v1.1.0 and the recommended workaround was `CODEGRAPH_NO_DAEMON=1` to skip the shared daemon on WSL2 `/mnt`.

In this case:

- CodeGraph is `1.1.3`.
- The MCP server process has `CODEGRAPH_NO_DAEMON=1`.
- The MCP server also has `--no-watch`.
- The same Codex parent process spawns two CodeGraph MCP children for the same physical repo with different path casing.
- The same SQLite database inode is shared by those two path spellings.
- A relocated internal daemon still appears in `.codegraph/daemon.pid`.
- The index can become malformed, then CLI recovers after killing repo-matching MCP processes, while MCP still returns `Transport closed`.

## Expected behavior

On WSL2 `/mnt` / DrvFs, CodeGraph should either:

- canonicalize/case-fold project roots that point to the same physical directory before deciding project identity and `.codegraph` ownership, or
- lock the SQLite database robustly across equivalent path spellings, or
- fail fast with an actionable error when the same physical `.codegraph` directory is opened through multiple path spellings.

It should not allow two MCP processes from the same client parent to treat upper/lowercase spellings of the same DrvFs path as separate project roots while writing the same SQLite database.

## Questions

1. Is `CODEGRAPH_NO_DAEMON=1` expected to fully prevent creation/use of a shared daemon in v1.1.3, or can a relocated `CODEGRAPH_DAEMON_INTERNAL=1` process still appear?
2. Does CodeGraph currently canonicalize WSL DrvFs paths by physical identity, or only by path string/realpath?
3. Would you prefer this be fixed in CodeGraph project-root identity handling, or should MCP clients like Codex avoid sending `projectPath` values that differ in casing from the server process cwd?

I can provide additional command transcripts if useful.

inth3shadows:
## Problem (#1057)

On WSL, opening a repo under `/mnt` via two different path spellings — most concretely an upper- vs lowercase variant of the same path — corrupts `.codegraph`, even with `CODEGRAPH_NO_DAEMON` set. The same class of bug hits a symlinked checkout on any platform.

## Root cause

`ToolHandler.getCodeGraph` caches each open `CodeGraph` (a live SQLite connection) in `projectCache`, and reuses the default instance, keyed on the **resolved-root path string**. Two spellings of one physical directory resolve to two different strings → two cache entries → **two SQLite connections to the same `.codegraph/codegraph.db`**. Concurrent writes across the two connections corrupt the index. (This is the same second-connection hazard already documented at the `#238` comment in that method.)

## Why not just `realpathSync` the root

That was my first attempt, and it is **insufficient** — verified on a real WSL DrvFs mount:

```
realpathSync('/mnt/c/.../MyProj') -> /mnt/c/.../MyProj
realpathSync('/mnt/c/.../myproj') -> /mnt/c/.../myproj   # casing preserved!
```

`realpathSync` resolves symlinks and `.`/`..`, but on a case-insensitive **case-preserving** filesystem it returns the caller's casing, so it cannot dedupe case-variants. Filesystem identity `(dev, ino)` is identical for every spelling and is the robust key.

## Fix

Key `projectCache` and the default-instance reuse check on `(dev, ino)` via a new `canonicalRootKey()` helper. This mirrors the inode-identity pattern this codebase already uses in `DatabaseConnection.openedInode` (`statInode`) for replace-on-disk detection, so it's idiomatic rather than novel. Minimal blast radius: `findNearestCodeGraphRoot` itself is unchanged.

## Reproduced (WSL2 / Ubuntu)

Before fix, against the built resolver:

```
[drvfs] via MyProj: /mnt/c/.../MyProj
[drvfs] via myproj: /mnt/c/.../myproj   -> two roots -> two connections (corruption)
```

After fix, `canonicalRootKey` converges:

```
symlink converges:            true
drvfs case-variant converges: true
distinct projects stay distinct: true
```

## Tests

`__tests__/root-identity.test.ts` (4 tests). The symlink case is a deterministic, filesystem-agnostic proxy for the case-insensitive-mount scenario (both produce two path strings for one inode), so it runs on case-sensitive CI. Full suite green locally (the only failures were two pre-existing CPU-contention timing flakes — `query-pool`, `mcp-daemon` — that pass in isolation).

## Scope / follow-up

This fixes the **in-process** connection cache — the reported `CODEGRAPH_NO_DAEMON` path.

`daemon-registry.ts` also hashes the path string rather than `(dev, ino)`, so two spellings can produce two discovery entries under `~/.codegraph/daemons/` for the same daemon. That's a smaller, separate bug than this PR's — the actual daemon-start arbitration (`daemon.ts`'s `acquireLockViaExclusiveOpen`, an O_CREAT|O_EXCL create on a real file inside the project's own `.codegraph/`) is already safe against path-spelling variance, since the OS resolves any spelling to the same physical file. So the with-daemon path doesn't get the same connection-level corruption this PR fixes — worst case is a cosmetic duplicate/stale entry in `codegraph list` / `stop --all`. Not expanding this PR for it.

---

Validated on WSL2 (Ubuntu). Happy to adjust naming/placement to your preference.

---
**Edit:** Softened the "Scope / follow-up" section above — my original wording called this "the same class of issue" as the corruption bug this PR fixes. On a closer look at `daemon.ts`'s lock arbitration, it isn't: the daemon-start path is already safe by construction, and the only actual gap is the cosmetic duplicate-listing one described above. Didn't want the overstatement to stand uncorrected.

Fixes #1057


inth3shadows:
Re-verified against `main` @ `5955d04`: **#1057 still reproduces and this fix is not superseded.**

`ToolHandler`'s open-connection cache is still keyed by the resolved-root **path string**, not filesystem identity:

- `src/mcp/tools.ts:1096` — default-instance reuse is `this.cg.getProjectRoot() === resolvedRoot` (string equality)
- `src/mcp/tools.ts:1099-1101` — `projectCache.get(resolvedRoot)` / `.set(resolvedRoot, cg)` ("Cache the open DB connection by RESOLVED ROOT only")

`findNearestCodeGraphRoot` returns a cased path string, so on a case-insensitive mount (WSL DrvFs `/mnt/c`, NTFS) two spellings of one repo resolve to two different strings → two `projectCache` entries → a second SQLite connection to the same `.codegraph/codegraph.db` — the exact second-connection corruption mechanism called out for #238 right above that code.

This PR's `canonicalRootKey` (`dev:ino`) collapses every spelling of one physical root onto a single cache entry, which is what prevents the second connection.

Status: the branch now **conflicts** with `main` — the cache region here, `src/directory.ts`, and `CHANGELOG.md` have all moved since it was opened — so it needs a rebase before it can merge, but the approach applies unchanged.

inth3shadows:
Rebased onto `main` @ `5955d04` and force-pushed — now `MERGEABLE`.

Conflicts were `CHANGELOG.md` plus re-applying the inode-keying to the current `freshen()`-wrapped cache block in `tools.ts`: `openSync(resolvedRoot)` still opens by the resolved path, while the default-instance check and `projectCache` get/set are now keyed by `canonicalRootKey(resolvedRoot)`.

Verified on Node 22: `tsc --noEmit` clean, the `root-identity` regression suite (4 tests) passes, and `mcp-unindexed` (8 tests) passes. Also added `Fixes #1057` so the issue closes on merge.

inth3shadows:
Rebased onto `main` @ `572d22b` (post-1.5.0) and force-pushed — `MERGEABLE` again.

Only `CHANGELOG.md` conflicted. `src/mcp/tools.ts` and `src/directory.ts` applied clean, and the keying still sits in the right place in the current `getCodeGraph`: `canonicalRootKey(resolvedRoot)` is computed after `findNearestCodeGraphRoot` and used by both the default-instance check and the `projectCache` get/set, while `openSync` still opens by the resolved path.

**Changelog entry moved to `[Unreleased]`.** It had been sitting in what is now the released `[1.5.0]` Fixes block; per CONTRIBUTING/CLAUDE guidance new entries belong under `[Unreleased]` for `prepare-release.mjs` to promote. `main`'s #1351 entry was kept as-is.

## Re-verified rather than re-asserted

Since the PR body's central claim (that `realpathSync` is insufficient) rested on a one-off probe from June, I re-ran it against the current build.

**1. On a real case-insensitive mount** (WSL2 DrvFs `/mnt/c`, 9p), two spellings of one physical directory:

| check | result |
|---|---|
| mount genuinely case-insensitive | `true` |
| `realpathSync` of `…/CgProbe` vs `…/cgprobe` | `/mnt/c/…/CgProbe` vs `/mnt/c/…/cgprobe` — **does not converge** |
| `(dev,ino)` for both spellings | `69:103019841476130290` — identical |
| `findNearestCodeGraphRoot` (the pre-fix key) | two distinct strings |
| `canonicalRootKey` (this PR) | one key for both |
| two distinct projects | stay distinct |

**2. End-to-end through `ToolHandler` itself** — real class from `dist/`, two real indexed projects, `getCodeGraph()` called with both spellings. Run twice: once as shipped, once with a `dist`-only revert of the keying line back to the path string.

| | pre-fix (path-string key) | this PR (`dev:ino` key) |
|---|---|---|
| same `CodeGraph` instance returned | `false` | `true` |
| `projectCache.size` — i.e. SQLite connections to one DB | **2** | **1** |
| cache keys | two path strings | one `dev:ino` |
| query via the second spelling | works | works |
| a distinct project still gets its own connection | — | `true` |

The `2 → 1` is the whole fix: two live connections to one `.codegraph/codegraph.db` is the corruption mechanism described in the #238 comment right above that code.

Full suite green on the rebased commit: **146 files passed / 15 skipped, 2534 tests passed, 0 failed** (Node 22, Linux). No flakes this run.

## Scope of the verification

Verified on Linux (ext4) and WSL2 DrvFs. **Not verified on native Windows/NTFS** — I don't have a Windows machine to test on. DrvFs is 9p emulating Windows' case-insensitivity, so the inode identity the fix keys on is WSL's, not NTFS's; whether `fs.statSync().ino` is equally stable across spellings on native NTFS is something I can only reason about, not demonstrate. If that matters for merging, it's worth a run on the Windows VM. The symlink case in `__tests__/root-identity.test.ts` is deterministic and filesystem-agnostic, so CI covers the mechanism on any platform even where case-insensitivity isn't reproducible.

The daemon-registry note in the original description is unchanged and still out of scope here.

inth3shadows:
Updated for the current \`main\` — merged \`main\` in (rather than rebasing) so the branch history stays append-only. Only \`CHANGELOG.md\` conflicted; \`src/mcp/tools.ts\` merged cleanly. The conflict was resolved by keeping the current `Fixes` list verbatim and appending this PR's entry to it. `tsc --noEmit` is clean and `__tests__/root-identity.test.ts` passes 4/4 on the merged tree.

inth3shadows:
Refreshed against `main` @ `2f8cce5` — `MERGEABLE` again. Fourth time, and this one is short because only `CHANGELOG.md` conflicted; `src/mcp/tools.ts` and `src/directory.ts` merged clean despite `tools.ts` having moved ~330 lines since the last refresh.

CHANGELOG resolved by taking `main`'s `[Unreleased]` wholesale and re-adding only the #1057 entry, under its `#### MCP / indexing` sub-heading. The other two this branch had been carrying — the telemetry relocation and the explore dedup — are already on `main`, so they are dropped rather than duplicated.

## Re-verified rather than re-asserted

The auto-merge succeeding is not the same as the fix still being in the right place, so I checked all three sites in `getCodeGraph`:

```
tools.ts:1773   const rootKey = canonicalRootKey(resolvedRoot);
tools.ts:1782   if (this.cg && canonicalRootKey(this.cg.getProjectRoot()) === rootKey)
                this.projectCache.get(rootKey) / .set(rootKey, cg)
                openSync(resolvedRoot)   <- still opens by the resolved path
```

End-to-end through the real `ToolHandler` from `dist/`, on a genuinely case-insensitive mount (WSL2 DrvFs `/mnt/c`), with two indexed projects and two spellings of one of them:

```
                                          main @ 2f8cce5       this branch
same instance for the two spellings                false              true
projectCache.size  (= live connections)                3                 2
cache keys                        two path strings + one    two dev:ino keys
a distinct project stays distinct                   true              true
```

**Three connections for two projects** — two of them open on the same `.codegraph/codegraph.db`. That is the second-connection mechanism described in the #238 comment directly above this code, and it is still there on today's `main`.

`tsc --noEmit` clean; `root-identity` 4/4 and `mcp-unindexed` 8/8 on the merged tree.

## One thing worth deciding alongside this

While checking whether the new writer lock (#1744) had made this redundant, I found it hasn't — and that it inherits the same weakness. `writer-lock.ts` is a cross-process pidfile acquired once per server (`daemon.ts:210`, `engine.ts:294`, `index.ts:373`); it cannot see one process holding two connections, since a second acquire from the same pid returns `{kind:'acquired'}` by design. And `getWriterPidPath` normalizes with `fs.realpathSync`, which does not converge on a case-insensitive mount:

```
dev:ino  cg1082probe / CG1082PROBE : 68:128915539333482220   (same directory)
realpath                           : …/cg1082probe  vs  …/CG1082PROBE
converges                          : false
```

So two direct-mode servers started at two spellings of one repo each write their own `writer.pid` and both acquire. I noted this on #1057 as well and offered to file it separately — happy to do that, or to fold a `canonicalRootKey`-based `getWriterPidPath` into this PR, whichever you prefer.

This has been open since June 30 and refreshed four times; if the approach isn't one you want, I would rather close it than keep it green indefinitely.

inth3shadows:
**Correcting myself: the writer-lock half of my earlier comment is wrong. Please disregard it.**

I claimed that because `getWriterPidPath` normalizes with `fs.realpathSync`, and realpath does not converge on a case-insensitive mount, two direct-mode servers started at two spellings of one repo would each write their own `writer.pid` and both acquire. I inferred that from the realpath behaviour without testing the acquire itself. Tested now, against `main` @ `b8d46d1`, two separate processes on WSL2 DrvFs `/mnt/c`:

```
process 1   /mnt/c/…/CgWriterLock   ->  kind=acquired
process 2   /mnt/c/…/cgwriterlock   ->  kind=taken      <- correctly refused
control     same spelling again     ->  kind=taken
```

The lock holds. The reason is that the pidfile path being a different *string* does not make it a different *file*: on a case-insensitive mount both spellings resolve to one inode, so the `linkSync` / `O_EXCL` create hits `EEXIST` whichever spelling asks.

```
wrote via CgWL2, readable via cgwl2      : true
O_EXCL create through the other spelling : EEXIST
```

The filesystem does the deduplication that realpath doesn't, and for a symlinked checkout realpath converges anyway. So there is no writer-lock defect here and nothing to file — my apologies for the noise.

**What this does not change:** everything else in that comment stands, and I re-verified it at `b8d46d1` rather than assuming it survived the day's commits:

```
tools.ts:1772   this.cg.getProjectRoot() === resolvedRoot     <- still a path string
tools.ts:1779   this.projectCache.get(resolvedRoot)
tools.ts:1783   this.projectCache.set(resolvedRoot, cg)
```

`projectCache` is still keyed by the path string, two spellings still produce two entries, and the measured `3 connections for 2 projects` through the real `ToolHandler` is unaffected — that path never touches the writer lock.

PR #330, opened 2026-05-22
feat: add Nix language support

uxtechie:
## Summary

Add full Nix (`.nix`) language support to CodeGraph, including symbol extraction, import resolution, and export detection.

Closes #332

**Related:** #324 — While that issue requests distributing CodeGraph as a Nix flake (packaging), this PR adds support for **parsing and indexing** Nix source files as a new language in the extraction engine.

### Symbol Extraction (`src/extraction/languages/nix.ts`)
- **Functions**: Extracts named function bindings with parameter signatures, including:
  - Simple parameters: `y = arg: arg + 1;` → signature `(arg)`
  - Destructured parameters: `z = { name }: ...` → signature `{ name }`
  - **Curried functions**: `f = a: b: c: body` → signature `a : b : c`
  - **Destructuring with `@` alias**: `{ x, y } @ args` and `args @ { x, y }` patterns
- **Variables**: Extracts non-function bindings as variable nodes
- **Inherit declarations**: Processes `inherit` and `inherit (expr)` as variable nodes
- **Call references**: Extracts function calls from `apply_expression`, distinguishing `import` expressions (emitted as `imports` references) from regular calls
- **Export detection**: Implements `isExportedNode()` helper that traverses AST ancestors to determine if a symbol is in the top-level attribute set (exported) vs. inside a `let` binding or nested attribute set (private)

### Import Resolution
- **`src/resolution/import-resolver.ts`**: Registers `.nix` and `/default.nix` extensions in `EXTENSION_RESOLUTION`
- **`src/resolution/index.ts`**: Adds file-level import resolution for `imports`-kind references — resolves directly to `file:` node IDs instead of symbol name matching. Also adjusts the pre-filter to avoid discarding valid import references

### Supporting Files
- **`src/extraction/grammars.ts`**: Registers tree-sitter-nix grammar (vendored WASM, loaded from `src/extraction/wasm/` like Pascal, Scala, Lua, and Luau)
- **`src/extraction/languages/index.ts`**: Registers Nix extractor
- **`src/types.ts`**: Adds `nix` to the `Language` type
- **`src/extraction/wasm/tree-sitter-nix.wasm`**: Bundled WASM parser (no new runtime or dev dependencies added — the grammar was compiled offline and vendored directly)

### Extraction tests (`__tests__/extraction.test.ts`)
- Basic variable and function extraction
- `inherit` and `inherit (expr)` handling
- Import expression extraction (relative paths and `builtins.import`)
- Function call reference extraction
[...]

PR #1084, opened 2026-06-30
Add Nix language support

TyceHerrman:
Supersedes #219 and #330.

## Summary
- Adds Nix .nix detection, display name, Language entry, extractor registration, and vendored tree-sitter-nix WASM with no new package dependency.
- Uses #330 as the functional base because it carried the broader extractor and resolver behavior; credits #330 for that implementation and #219 for the README/default include intent.
- Extracts Nix variables vs functions, simple/destructured/curried params, inherit bindings, static import refs, function call refs, and exported vs private attrset scope.
- Resolves only static relative Nix path imports to indexed file nodes, including ./dir to dir/default.nix. Angle-bracket channels, attribute expressions, and variable/dynamic imports do not create project file edges.
- Keeps existing import resolution behavior by scoping Nix file-path imports beside the existing C/C++ and PHP file-import branches.

## Notes
- Current main no longer has DEFAULT_CONFIG. Nix is included through EXTENSION_MAP/isSourceFile, which is the current indexing source of truth.

## Test plan
- npm run build
- npm test -- __tests__/extraction.test.ts __tests__/resolution.test.ts
- npm test was attempted locally, but the full suite entered a broad unrelated 5s timeout cascade across many heavy suites under local resource pressure, so it was stopped.


colbymchenry:
Merged in #1190, carrying this PR as the functional base — your extractor, import resolution, and tests all landed (with authorship preserved on the branch; the squash commit carries your co-author credit), plus follow-ups on top: an ABI-15 rebuild of the grammar wasm, NixOS `imports = [ ... ]` / `callPackage ./path` file edges, module-system option wiring, and Nix-scoped resolution gates. Thank you for the clean base and for consolidating #219/#330 — closing in favor of #1190. Ships in the next release.

PR #1116, opened 2026-07-01
fix(login): avoid cmd.exe when opening the login URL on Windows (#1114)

### LINKED ISSUE #1114: security(low): openBrowser routes a server-supplied URL through Windows `cmd /c start` unescaped
## Summary

On Windows, `openBrowser` (`src/reasoning/login.ts`) opens the login URL by shelling out to `cmd`:

```ts
// src/reasoning/login.ts:78-84
const [cmd, args] =
  process.platform === 'darwin' ? ['open', [url]]
  : process.platform === 'win32' ? ['cmd', ['/c', 'start', '', url]]
  : ['xdg-open', [url]];
const child = spawn(cmd, args, { stdio: 'ignore', detached: true, windowsHide: true });
```

`spawn` runs without `shell: true`, so **Node** does not shell-parse the argv — good. But `cmd.exe` then **re-parses** the arguments to its `start` built-in, so shell metacharacters (`&`, `|`, `^`, `"`) inside `url` are interpreted by cmd, not treated as a literal URL.

## How `url` gets there

`openBrowser` is called from the `codegraph login` device-flow. The URL is external input:

- `startDeviceLogin()` (`login.ts:36`) fetches `verification_uri` / `verification_uri_complete` from the dashboard (`app.getcodegraph.com`), and
- the base is overridable via the `CODEGRAPH_LOGIN_URL` env var (`loginBaseUrl()`, `login.ts:17`).

So the string reaching `cmd /c start` originates from a remote response or an environment variable — not a hardcoded constant.

## Impact

Defense-in-depth only. To exploit it an attacker needs to control that URL — i.e. MITM/compromise the first-party HTTPS endpoint, or poison `CODEGRAPH_LOGIN_URL` in the victim's environment. Given either, a crafted URL (e.g. containing `&calc`) could inject an additional command on Windows when the user runs `codegraph login`. macOS (`open`) and Linux (`xdg-open`) pass the URL as a single argv element and are not affected. Low severity, but it's untrusted input reaching a `cmd` invocation.

## Suggested fix

Either validate before spawning — reject a `url` that doesn't match `^https?://` — or open it without going through `cmd start`, e.g. `spawn('rundll32', ['url.dll,FileProtocolHandler', url])` (or `ShellExecute`), which takes the URL as a single opaque argument.

## Environment

- Found on `main` (tip `3424ff3`); Windows only. Not yet in a tagged release.

_Reported from a red-team pass. Reachability of the call chain confirmed against the current graph._

maxmilian:
Closes #1114.

## Problem

`openBrowser()` (`src/reasoning/login.ts`) launches the device-login verification URL that comes from the login server response (`verification_uri_complete`, where the server base is `CODEGRAPH_LOGIN_URL` / default `app.getcodegraph.com`). On Windows it ran:

```ts
: process.platform === "win32" ? ["cmd", ["/c", "start", "", url]]
```

`cmd.exe` (and its `start` builtin) re-parse shell metacharacters (`&  |  ^  "`). A malicious or compromised login server — or a user tricked into pointing `CODEGRAPH_LOGIN_URL` at an attacker — could return a `verification_uri_complete` such as `https://app.getcodegraph.com/device?...&calc` and get arbitrary commands executed. (`spawn` without `shell: true` does not help here: libuv only quotes args containing whitespace/quotes, so a space-free URL reaches cmd.exe unescaped.) Defense-in-depth (`low`): it needs a hostile server/env, but the browser-open path should never route a server-supplied string through a shell.

## Fix

Two layers:

1. **Launch via `rundll32 url.dll,FileProtocolHandler <url>`** instead of `cmd /c start`. `rundll32` receives the URL as a single opaque `argv` element (Node `spawn` without `shell`), so there is no shell re-parsing. The macOS (`open`) and Linux (`xdg-open`) branches already pass the URL as a lone argument and are unchanged.
2. **`https?://` scheme allowlist** before launching anything. All three OS handlers (`open` / `rundll32,FileProtocolHandler` / `xdg-open`) will happily execute a `file:///...exe`, bare, or UNC (`\\host\share\payload.exe`) path handed to them — swapping the launcher alone doesn't close that, the allowlist does, on every platform.

## Tests

`__tests__/open-browser-security.test.ts`: stubs `process.platform = "win32"`, mocks `child_process.spawn`, and asserts (a) a URL containing `&` is not launched through `cmd`/`/c`, arrives at the launcher as one intact argument, and never with `shell: true`; (b) `file:`/UNC/bare-path/`javascript:` targets are refused (no spawn at all); (c) plain `http://` still launches. Written test-first (fails on the old `cmd` path, passes after the fix). `vitest run` + `tsc --noEmit` green.

## Reachability note (please weigh in)

As of current `main`, `openBrowser` — and all of `src/reasoning/login.ts` — has **no callers**: e5897d0 removed the `codegraph login` / `logout` / `usage` command wiring and left the module orphaned. So on `main` this is hardening dead code (the vulnerable path does ship in already-released versions that had the command). If you'd rather delete `login.ts` outright than patch it, happy to rework this PR into the removal instead.

> I develop on macOS, so I could not exercise the `rundll32` path on a real Windows host. `openBrowser` is best-effort (errors are swallowed), so a Windows regression degrades gracefully rather than breaking login.

### DIFF
--- a/__tests__/open-browser-security.test.ts
+++ b/__tests__/open-browser-security.test.ts
@@ -1,0 +1,42 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+
+// openBrowser() launches a URL that originates from the device-login server
+// response (verification_uri_complete). On Windows the URL must never be handed
+// to cmd.exe, which re-parses shell metacharacters (& | ^ ") and would allow a
+// malicious/compromised login server to inject commands. See issue #1114.
+const { spawnMock } = vi.hoisted(() => ({ spawnMock: vi.fn() }));
+vi.mock('child_process', () => ({ spawn: spawnMock }));
+
+import { openBrowser } from '../src/reasoning/login';
+
+describe('openBrowser — Windows shell-injection hardening (#1114)', () => {
+  const realPlatform = process.platform;
+
+  beforeEach(() => {
+    spawnMock.mockReset();
+    spawnMock.mockReturnValue({ on: vi.fn(), unref: vi.fn() });
+    Object.defineProperty(process, 'platform', { value: 'win32', configurable: true });
+  });
+
+  afterEach(() => {
+    Object.defineProperty(process, 'platform', { value: realPlatform, configurable: true });
+  });
+
+  it('does not launch the URL through cmd.exe on Windows', async () => {
+    const url = 'https://app.getcodegraph.com/device?user_code=ABCD-1234&calc';
+
+    await openBrowser(url);
+
+    expect(spawnMock).toHaveBeenCalledTimes(1);
+    const [command, args] = spawnMock.mock.calls[0] as [string, string[]];
+
+    // cmd.exe (and its `start` builtin) re-parse metacharacters, so the launcher
+    // must not be cmd and must not carry cmd's `/c` switch.
+    expect(command.toLowerCase()).not.toContain('cmd');
+    expect(args).not.toContain('/c');
+
+    // The server-supplied URL is passed to the launcher as one intact argument,
+    // never concatenated into a shell command string.
+    expect(args).toContain(url);
+  });
+});
--- a/__tests__/open-browser-security.test.ts
+++ b/__tests__/open-browser-security.test.ts
@@ -38,5 +38,29 @@ describe('openBrowser — Windows shell-injection hardening (#1114)', () => {
     // The server-supplied URL is passed to the launcher as one intact argument,
     // never concatenated into a shell command string.
     expect(args).toContain(url);
+
+    // And never through a shell at all.
+    const options = spawnMock.mock.calls[0][2] as { shell?: boolean } | undefined;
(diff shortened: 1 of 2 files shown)


maxmilian:
Closing this — it's been overtaken by #1132, which is the better fix.

I opened #1114 reporting that `openBrowser` passed a server-supplied URL through `cmd /c start` unescaped on Windows, and this PR escaped the argument. @colbymchenry went further and removed the dead reasoning-offload modules outright (`04e2391`, "chore(security): remove dead reasoning-offload modules flagged in #1114"), deleting `src/reasoning/login.ts` altogether. Deleting the unused module beats hardening it — there's no sink left to escape.

Confirmed against current `main`: `src/reasoning/login.ts` is gone and there is no remaining `cmd.exe` / `cmd /c start` invocation under `src/`. Rebasing this branch would only reintroduce the deleted file, so there's nothing left to land. Thanks for the thorough cleanup.

Issue #1080, opened 2026-06-30
Frequent codegraph_explore calls in v1.16 make Codex slow for small tasks


qg-hs:
Hi, thanks for building CodeGraph. It is a very useful tool for understanding call chains, impact scope, and cross-module relationships.

However, after upgrading to the latest version `v1.16`, I noticed that Codex now frequently calls `codegraph_explore` even for very small tasks.

For example, for simple UI changes such as:

* removing one line of text from a card
* updating loading skeletons on a few pages
* changing labels or copy
* small style/layout changes

Codex still tends to call `codegraph_explore` first instead of using a simple `rg` / read / edit flow.

This causes a noticeable slowdown because `codegraph_explore` often returns a lot of source code, relationship information, and blast radius data. That returned content enters the model context, which increases token usage and makes Codex much slower.

In my use case, `codegraph_explore` is great for structural tasks, such as:

* call chain analysis
* data flow / state flow analysis
* permission boundary review
* impact scope before refactor
* security review
* architecture exploration

But for small known-file edits, UI text changes, skeleton loading updates, simple config changes, or literal string lookups, it feels too heavy as the default tool.

Would it be possible to improve the MCP guidance or tool selection behavior so that `codegraph_explore` is not recommended as the default tool for almost every task?

Ideally, the behavior could be something like:

```text
Known file / small edit      → read/edit directly
Literal lookup               → rg or lightweight search
Local module issue           → lightweight context/search first
Cross-module structural task → codegraph_context / trace / impact
Full source + relationship map needed → codegraph_explore
```

A configuration option to make `codegraph_explore` opt-in or lower priority would also be helpful. For example, allowing users to prefer lightweight tools by default and only use `explore` when explicitly needed.

The main issue is not that `codegraph_explore` is wrong, but that it is too heavy for small Codex tasks and causes unnecessary latency and token consumption.

Thanks again for the project. I think CodeGraph is very valuable, but this behavior makes the Codex experience much slower in day-to-day development.

Issue #975, opened 2026-06-24
Codegraph consume more tokens than just talk to claudecode.


charlietsao:
I tested on one private code repo with codegraph enabled vs disabled with claude and deepseek-v4 for 3 different code questions. The result shows that it consumes more:

enabled:
```
Total cost:            $0.4051 (costs may be inaccurate due to usage of unknown models)                                                                                                        Total duration (API):  36s   Total duration (wall): 7m 58s   Total code changes:    0 lines added, 0 lines removed   Usage by model:      deepseek-v4-flash:  44.9k input, 1.7k output, 274.7k cache read, 0 cache write ($0.4051)

Total cost:            $0.86 (costs may be inaccurate due to usage of unknown models) Total duration (API):  1m 15s Total duration (wall): 4m 1s Total code changes:    0 lines added, 0 lines removed Usage by model:    deepseek-v4-flash:  107.1k input, 4.6k output, 412.2k cache read, 0 cache write ($0.86)

Total cost:            $0.72 (costs may be inaccurate due to usage of unknown models) Total duration (API):  37s Total duration (wall): 8m 18s   Total code changes:    0 lines added, 0 lines removed   Usage by model:      deepseek-v4-flash:  108.8k input, 3.6k output, 164.9k cache read, 0 cache write ($0.72)
```

disabled:
```
Total cost:            $1.67 (costs may be inaccurate due to usage of unknown models) Total duration (API):  2m 23s Total duration (wall): 7m 49s Total code changes:    0 lines added, 0 lines removed Usage by model:    deepseek-v4-flash:  245.9k input, 11.3k output, 321.8k cache read, 0 cache write ($1.67)

Total cost:            $0.64 (costs may be inaccurate due to usage of unknown models) Total duration (API):  5m 58s Total duration (wall): 10m 22s Total code changes:    0 lines added, 0 lines removed Usage by model:    deepseek-v4-flash:  56.3k input, 1.8k output, 0 cache read, 0 cache write ($0.3270)     claude-haiku-4-5:  60.8k input, 4.4k output, 2.3m cache read, 0 cache write ($0.3150)

[...]

Issue #676, opened 2026-06-03
add CLI skills


doctorcolossus:
My use case is that I'd like to use a containerized terminal backend with `hermes-agent`. The assumption seems to be that the MCP server will run in the host environment, whereas if my agent is running commands in a container, the MCP server there won't have access to the files in the container. I could try to get hermes to run the MCP server in the container and work out how to ensure and configure a consistent URL using a network alias or something, or try to figure out some file-sharing mechanism using volume-mounts, but it's annoying to have to add such extra complexities to my setup just to get `codegraph` working with it.

It seems like it would be much simpler and more elegant if my agent could just use the excellent CLI directly without MCP.
I can see that you have skills `add-lang` and `agent-eval`. Have you considered adding skills for the CLI?

treytracedit-lab:
https://help.instagram.com/512371932629820/?helpref=uf_share

0DennisLee0:
yeah we need this

colbymchenry:
Thanks — this is a real setup friction, but there may already be a clean answer. codegraph ships a **full CLI that works completely independently of the MCP server**: `codegraph index`, `codegraph query`, `codegraph callers`, `codegraph callees`, `codegraph context`, etc. all run standalone. So inside your container you can install codegraph and run those directly against the container's filesystem — no host MCP server, no volume-sharing, no network alias needed. Point your hermes agent at the CLI as a tool/skill and it works where the code actually lives.

The one thing to manage is index freshness — run `codegraph sync` (or a git hook) in the container after edits.

If what you specifically want is a first-class, pre-packaged *skill* wrapper around the CLI for hermes (rather than wiring the commands yourself), that's a fair feature ask and I'll keep this open for it. Does running the CLI in-container unblock your use case in the meantime?

doctorcolossus:
> If what you specifically want is a first-class, pre-packaged _skill_ wrapper around the CLI for hermes (rather than wiring the commands yourself), that's a fair feature ask and I'll keep this open for it.

Yes, at least one `SKILL.md` file specifically for the CLI is exactly what I was suggesting.

[...]

PR #1117, opened 2026-07-01
docs: add community usage skill (codegraph-usage) — #1097

### LINKED ISSUE #1097: [Proposal] Community skill
## Problem

Currently, there's MCP with endpoints descriptions, that does not cover some use cases. Agent sometimes goes misusing. And it's not about the bugs, it's about poor documentation for agent.

## Proposal

Create skill (or few skills each for different toolset: explore-only or using separate tools) that is built by users experience. Edge-cases, gotchas and workarounds that people face and report can be added to that skill.

## Why

I had to create my own skill to make agent work better with codegraph. Here's an example:

```markdown
# Codegraph Usage

## Nested types

`codegraph` does NOT traverse nested struct types transitively. When searching usage of a composite struct, use a two-step approach:

1. Query the parent struct → extract nested type names from the source code in the response.
2. Query again with all nested types explicitly listed (e.g. `"config.Log config.S3 config.S2S config.Stages"`).

Do NOT fall back to grep/read for this — codegraph already returns the struct definition with nested type names in step 1.

## Impact analysis

`codegraph_impact` answers the question "what depends on this symbol?". Use it whenever you need that answer — planning a change, answering a question, or evaluating risk. Not only for refactor.

Do not reconstruct this by reading source files and tracing callers manually. `codegraph_impact` does this in one call and covers transitive dependencies that manual tracing misses.

`codegraph_impact` vs `codegraph_callers`: `codegraph_impact` answers "What depends on X transitively?", `codegraph_callers` shows only direct callers.

## Limitations

Structural search only. The graph knows calls and imports, but does NOT understand semantics. "Find code conceptually similar to X" is NOT its job — use another approach for semantic/conceptual search.
```

> Note: `## Nested types` is probably fixed in [#1064](https://github.com/colbymchenry/codegraph/issues/1064), the skill was written before (actual for v1.1.3)

Collecting more user experience we can improve tool usage by agents.

**Also:** codegraph a bit lacks documentation for human: concrete cases what is this tool for, and the main thing, what it is NOT for

Badi21:
## Summary

Adds `skills/codegraph-usage/SKILL.md` — the community field guide proposed in #1097: an agent-agnostic skill collecting real-world usage patterns, gotchas, and workarounds so agents (and humans) use CodeGraph well.

Covers:

- **When to reach for CodeGraph and when NOT to** — structural search only, not semantic/conceptual; skip it for trivial single-file edits; correctness stays with compiler/tests (the "what it is NOT for" documentation the issue asks for).
- **Tool selection table** — `explore` / `node` / `query` / `callers` / `callees` / `impact` / `affected` / `files`, including the most common mixup: `impact` (transitive blast radius) vs `callers` (one hop).
- **Community-reported gotchas** — over-calling `explore` on small tasks (#1080, #975), per-file vs global staleness banner semantics, overloaded symbol disambiguation, `projectPath` in monorepos, CLI-without-MCP for containers/subagents (#676), index freshness when scripting (`sync` vs `index`).
- **A "Contributing" section** so users can PR their own edge cases into the skill — the collect-user-experience loop the proposal describes.

Content is grounded in `src/mcp/server-instructions.ts` and `site/src/content/docs/reference/cli.md` — no invented commands or behavior.

Complements (doesn't overlap) #1019, which wraps the CLI surface itself; this one is usage guidance for the toolset.

Closes #1097

## Testing

- Every command and flag named in the skill checked against the current CLI reference and MCP tool definitions.
- Front-matter follows the existing `.claude/skills/*/SKILL.md` format.

### DIFF
--- a/skills/codegraph-usage/SKILL.md
+++ b/skills/codegraph-usage/SKILL.md
@@ -1,120 +1,92 @@
 ---
 name: codegraph-usage
-description: Community field guide for working with CodeGraph effectively — when to reach for it, which tool answers which question, known gotchas and workarounds, and what CodeGraph is NOT for. Use when working in a repo that has a .codegraph/ index, or when CodeGraph tool calls return confusing or incomplete results.
+description: How to use CodeGraph well — which tool answers which question, when to reach for it and when not, and the gotchas that trip agents up. Read before using CodeGraph tools in a project that has a .codegraph/ index, and whenever a CodeGraph call returns confusing or incomplete results.
 ---
 
-# CodeGraph usage — community field guide
+# Using CodeGraph
 
-CodeGraph is a local SQLite knowledge graph of every symbol, call edge, and
-file in an indexed project. One query returns verbatim line-numbered source
-plus the call paths between symbols and a blast-radius summary — replacing a
-grep + read loop with a single round-trip.
+CodeGraph is a local knowledge graph of a project's symbols, call edges, and
+files. One call returns verbatim line-numbered source plus the call paths
+between symbols and a blast-radius summary — replacing a grep + read loop with a
+single round-trip. Read this before reaching for CodeGraph so you pick the right
+call and avoid the mistakes below.
 
-This skill collects usage patterns, gotchas, and workarounds reported by real
-users. It complements (does not replace) the MCP server instructions.
+## When to reach for it
 
-## When to reach for CodeGraph
+- "How does X work?" / "Where is X?" / "What happens when X runs?"
+- Before editing a symbol: source + callers + blast radius in one call.
+- Tracing a flow: name both endpoints in one query (e.g. `mutateElement renderScene`) —
+  it rides dynamic-dispatch hops (callbacks, re-renders) that grep can't follow.
+- Judging impact before a change, or when reviewing risk — not only your own refactors.
 
-- "How does X work?" / "Where is X defined?" / "What happens when X runs?"
-- Before editing a symbol: get its source, callers, and blast radius in one call.
-- Tracing a flow between two points: name both endpoints in one query
-  (e.g. `mutateElement renderScene`) — it finds the path, including
-  dynamic-dispatch hops (callbacks, React re-renders) grep can't follow.
-- Impact assessment before a refactor, when answering "is this safe to change?",
-  or when evaluating a PR's risk — not only when you are the one refactoring.
+## When NOT to
 
-## When NOT to reach for CodeGraph
+- **Semantic/conceptual search.** It knows structure (calls, imports, definitions),
+  not meaning. "Code similar to X" or "where do we handle billing edge cases?" —
+  reason over results yourself or use plain search.
+- **Unindexed content.** Configs, docs, lockfiles, generated code — use Read/Grep.
+- **Trivial single-file edits.** If you know the exact file and line, editing
+  directly is faster than a graph round-trip.
+- **Correctness.** No type-checking or test-running here — the compiler, tests,
+  and linter still own "is this right?".
 
-- **Semantic/conceptual search.** The graph knows calls, imports, and
-  definitions — structure, not meaning. "Find code conceptually similar to X"
-  or "where do we handle billing edge cases?" are not its job; use your own
-  reasoning over `explore` results, or plain search.
-- **Unindexed content.** Configs, docs, lockfiles, data files, generated code
-  — use Read/Grep for those.
-- **Trivial single-file tasks.** If you already know the exact file and line
-  ("fix the typo in README", "bump the version string"), calling CodeGraph
-  first adds a round-trip for nothing. A quick edit does not need a graph.
-- **Correctness validation.** The graph has no type checking or test running.
-  The compiler, test suite, and linter still own "is this right?" — CodeGraph
-  supplements them with structural context.
+## Which call answers which question
 
-## Which tool answers which question
+Which tools you can reach depends on how CodeGraph was set up: the default MCP
+surface is `codegraph_explore` alone, other tools are enabled per config, and all
+of them exist as CLI subcommands (`codegraph <command>`). Use whichever is
+available — `explore` alone answers most questions.
 
-The default MCP surface is `codegraph_explore` alone; the narrower tools below
-are available via the CLI (`codegraph <command>`) and re-enablable over MCP
-with `CODEGRAPH_MCP_TOOLS`.
-
-| You want to know | Use | Not |
+| You want | Use | Instead of |
 |---|---|---|
 | Almost anything: how X works, where X is, survey an area | `explore` | grep + read loop |
-| One symbol's source + direct callers | `node` | reading the whole file |
+| One symbol's full source + direct callers | `node` | reading the whole file |
 | "What calls X?" (direct, one hop) | `callers` | `impact` |
-| "What does X call?" | `callees` | reading X's body and grepping each name |
-| "What depends on X **transitively**?" — blast radius | `impact` | chaining `callers` by hand |
-| "Which tests are affected by these changed files?" | `affected` | running the full suite to find out |
-| File/module layout of an area | `files` | `ls -R` + reading each file |
+| "What does X call?" | `callees` | reading X and grepping each name |
+| "What depends on X **transitively**?" (blast radius) | `impact` | chaining `callers` by hand |
 | Symbol lookup by name/kind | `query` | grep |
+| File/module layout of an area | `files` | `ls -R` + reading files |
 
-**`impact` vs `callers` is the one people mix up:** `callers` is one hop;
-`impact` walks the graph transitively and covers dependents that manual
-tracing misses. If the question is "what could break?", it's `impact`.
+`node` returns a symbol's full source or a file's contents from the index — it is
+**not** a substitute for `read` on files CodeGraph doesn't cover.
 
-## Gotchas & workarounds (community-reported)
+CLI-only (not an MCP tool): `codegraph affected <files...>` — which test files a
+change touches, without running the suite.
 
-### Don't over-call explore on small tasks
-One `explore` call usually answers the whole question — treat the source it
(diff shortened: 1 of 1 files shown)


skills/codegraph-usage/SKILL.md
```diff
@@ -0,0 +1,120 @@
+---
+name: codegraph-usage
+description: Community field guide for working with CodeGraph effectively — when to reach for it, which tool answers which question, known gotchas and workarounds, and what CodeGraph is NOT for. Use when working in a repo that has a .codegraph/ index, or when CodeGraph tool calls return confusing or incomplete results.
+---
+
+# CodeGraph usage — community field guide
```
wauxhall: I think agent should not know about community, I would remove all mentions

skills/codegraph-usage/SKILL.md
```diff
@@ -0,0 +1,120 @@
+---
+name: codegraph-usage
+description: Community field guide for working with CodeGraph effectively — when to reach for it, which tool answers which question, known gotchas and workarounds, and what CodeGraph is NOT for. Use when working in a repo that has a .codegraph/ index, or when CodeGraph tool calls return confusing or incomplete results.
```
wauxhall: To make trigger work, agent should check for `.codegraph` existence. My experience is that agents go lazy especially for skill usage. Better - "Always use before any `codegraph` tool call. Also use when `codegraph` return confusing or incomplete results"

The best approach for modal monoliths with separate `.codegraph` dirs is explicitly tell in AGENTS.md what modules use codegraph and tell to use `codegraph` here for code exploration. This should be documented as skill usage for users.

My main idea - force reading the skill BEFORE tool call, when agent wants to use it. Skill contain prevent conditions that must be read by agent before it actually makes the call.

My AGENTS.md instruction btw: **Before calling any codegraph tool, load and read the `codegraph` skill first. No exceptions.**. It works

skills/codegraph-usage/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+
+# CodeGraph usage — community field guide
+
+CodeGraph is a local SQLite knowledge graph of every symbol, call edge, and
+file in an indexed project. One query returns verbatim line-numbered source
+plus the call paths between symbols and a blast-radius summary — replacing a
+grep + read loop with a single round-trip.
+
+This skill collects usage patterns, gotchas, and workarounds reported by real
+users. It complements (does not replace) the MCP server instructions.
```
wauxhall: I think `It complements (does not replace) the MCP server instructions.` is enough

skills/codegraph-usage/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+  first adds a round-trip for nothing. A quick edit does not need a graph.
+- **Correctness validation.** The graph has no type checking or test running.
+  The compiler, test suite, and linter still own "is this right?" — CodeGraph
+  supplements them with structural context.
+
+## Which tool answers which question
+
+The default MCP surface is `codegraph_explore` alone; the narrower tools below
+are available via the CLI (`codegraph <command>`) and re-enablable over MCP
+with `CODEGRAPH_MCP_TOOLS`.
```
wauxhall: This is MCP setting made by human. Agent see eithter only `explore` or other tools user configured. This is unnecessary info for agent

skills/codegraph-usage/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+
+| You want to know | Use | Not |
+|---|---|---|
+| Almost anything: how X works, where X is, survey an area | `explore` | grep + read loop |
+| One symbol's source + direct callers | `node` | reading the whole file |
```
wauxhall: maybe "full source code?". Just to be more prescise. + should add that `node` is not replacement for `read`, maybe not here, as a separate statement

skills/codegraph-usage/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+| "What calls X?" (direct, one hop) | `callers` | `impact` |
+| "What does X call?" | `callees` | reading X's body and grepping each name |
+| "What depends on X **transitively**?" — blast radius | `impact` | chaining `callers` by hand |
+| "Which tests are affected by these changed files?" | `affected` | running the full suite to find out |
```
wauxhall: affected? didn't see it in docs. Maybe new and I missed

skills/codegraph-usage/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Don't over-call explore on small tasks
+One `explore` call usually answers the whole question — treat the source it
+returns as already read. If the task is small and the first call answered it,
+stop. Repeated exploratory calls on a task that needed none make sessions
+slower and burn tokens (reported against agent harnesses that call
+before every minor edit). Rule of thumb: call when you have a genuine
+structural question, not as a reflex before every change.
+
+### Trust results — don't re-verify with grep
```
wauxhall: Think should focus on avoiding re-verifying results already provided by codegraph - reading the same but with grep. Why: this instruction looks like a strict rule to fully rely on codegraph's results. If there's a bug in codegraph that make result more poor (I faced, but already fixed), agent should fallback to grep for symbols codegraph didn't provide. + on large codebase blast radius is truncated with "+N more". Agent can trust what it see from codegraph but search for missing.

This is not critical, need to debate over this. Maybe I'm wrong

skills/codegraph-usage/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+directly; keep trusting the graph for every file not listed. The rarer
+"auto-sync is DISABLED" banner means the whole index is frozen — Read
+directly until it's resolved.
+
+### Overloaded / same-named symbols
+A bare name that exists in several places returns every matching definition.
+Disambiguate by adding the file path or a neighboring symbol to the query
+rather than picking the first result.
+
+### Monorepos: pass projectPath
```
wauxhall: Already fixed. Can be removed. Codegraph now force projectPath as required in schema if no .codegraph dir found. Works, confirmed: I use mcpjungle as mcp gateway, the place where server started obviously has no .codegraph dir, but agent always work as expected

skills/codegraph-usage/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+If the MCP server started somewhere with no `.codegraph/` of its own, tools
+still work per project — pass `projectPath` pointing into any directory that
+has an index. No restart needed after a new `codegraph init`.
+
+### Containers & subagents: use the CLI, skip MCP
```
wauxhall: Does agent have knowledge about how it's subagents are run? Maybe that's separate skill for those who use same approach?

skills/codegraph-usage/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+`node`, `query`, `callers`, `callees`, `impact`, `affected` — same output as
+the MCP tools. If your agent runs commands in a container, install codegraph
+in the container and call the CLI against the container's filesystem; no
+host/container MCP wiring needed. Keep the index fresh with `codegraph sync`
+(or a git hook) since the watcher may not run there. Add `--json` to `query`,
+`callers`, `callees`, and `impact` for machine-readable output.
+
+### Index freshness outside agent sessions
+The file watcher keeps the graph current (~1s lag) during normal sessions.
+When scripting against the index (CI, hooks, containers), run
```
wauxhall: Didn't understand that, could you explain more detailed what conditions match sync problem?

skills/codegraph-usage/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+`codegraph sync` first — an incremental update, cheap. `codegraph index` is
+a full rebuild and only needed when things look wrong.
+
+### Not indexed? Stop asking
+If a tool reports no `.codegraph/` for a project, stop calling CodeGraph
+there for the rest of the session and use built-in tools. Indexing is the
+user's call — suggest `codegraph init` if relevant, don't run it yourself.
+
+## Contributing to this skill
```
wauxhall: This is definetely not agent instruction. Instead, you can instruct agent to report tool misuse and offer to create issue. But I think this should be moved completely outside into user's reference for using the skill. This is context noise

Badi21:
Thanks for the detailed review. I pushed a rework that leans into the "agent-facing, read before use" framing you're pointing at. What I took directly:

- Dropped the community and Contributing sections (L6, L114). You're right they're noise for the agent; that kind of "how to contribute / AGENTS.md setup" guidance fits a README, not the skill.
- Removed the `projectPath` monorepo section (L89). Thanks for confirming the schema enforces it now.
- Removed the container/subagent CLI section and the offline-sync section (L94). Agreed, those are user setup rather than per-call guidance.
- Reworked the tool table: it now says availability depends on MCP config (default is `explore` only), and marks `affected` as CLI-only (L54). It's a real command, just a CLI subcommand rather than an MCP tool, so grouping it with the MCP tools was wrong.
- Added that `node` isn't a substitute for `read` on unindexed files (L50).
- Softened the "trust results" section per L72, which I think was your most important point: don't re-verify what CodeGraph already returned, but do fall back to grep/read when the blast radius is truncated (`+N more`) or a symbol is missing. A blanket "always rely on it" is wrong when the output is incomplete.

Two I want to check before hard-coding:

- Trigger (L3). I strengthened the description toward "read before using CodeGraph tools" but stopped short of "before every call, no exceptions." The absolute version fights the "don't over-call on small tasks" guidance in the same skill, and the maintainer's own server-instructions approach. Your AGENTS.md line clearly works for your setup. Is baking that hard rule into the skill the direction you want here @colbymchenry, or keep it as "read before use" and let operators add the stricter AGENTS.md rule themselves?
- Sync banner (L105). The condition is CodeGraph prepending "⚠️ Some files … edited since the last index sync" (per file) versus "auto-sync is DISABLED" (whole index frozen). Happy to reword if that's still unclear.

@colbymchenry, the trigger question is the one design call that shapes the rest, so I'd value your read on it.

PR #1118, opened 2026-07-02
Cfml language support

ghedwards:


colbymchenry:
Thank you for this — reviewed and **merged today** via #1153 (your branch doesn't allow maintainer edits, so I couldn't push the review fixes to this PR directly; your commits carry over and the squash commit credits you as co-author).

The review verified the wasm grammars are bit-for-bit reproducible from `cfmleditor/tree-sitter-cfml` (nice to see the grammar and the integration maintained by the same org), and the extractor held up well on real code — FW/1, ColdBox, and CFWheels all index with effectively 100% component coverage. Three fixes were added on top:

- a leading UTF-8 BOM made the tag-vs-script sniffer misroute tag-based files to the cfscript grammar (17% of ColdBox's files extracted nothing) — the sniffer now skips it
- unquoted attribute values (`<cffunction name=init>`) were dropped; `tagAttr` now reads both value shapes
- functions in a component-level `<cfscript>` block now classify as `method` (the ModuleConfig shape)

One follow-up worth knowing about if you want to keep going — #1152: dotted-path `extends="coldbox.system.web.Controller"` doesn't resolve to the class node yet, which hides most inheritance on ColdBox-style apps. The extraction side you built is done; it needs a CFML-aware step in reference resolution.

CFML ships in the next release. Thanks again — closing this in favor of #1153.

PR #1177, opened 2026-07-05
fix(db): probe WAL support and fall back to DELETE on unsupported filesystems (#990)

### LINKED ISSUE #990: WAL: auto-detect filesystem mmap support and fall back to DELETE mode
### Environment
- CodeGraph 1.1.1
- Linux kernel 7.1.0-rc7, ntfs3 filesystem
- Node.js v25.9.0 (node:sqlite)

### Symptoms
`codegraph init` fails with `disk I/O error` on ntfs3 (and likely other filesystems without shared-memory mmap: WSL2 `/mnt/`, CIFS, some NFS configurations).

### Root Cause
SQLite WAL mode requires `mmap(MAP_SHARED)` on the WAL-index file. ntfs3 returns `EOPNOTSUPP` for `MAP_SHARED`.

The current code in `db/index.js:configureConnection()` does:
```js
db.pragma('journal_mode = WAL');
```

However, `PRAGMA journal_mode = WAL` **reports success** ("wal") even when the underlying filesystem can't support it. The actual failure happens on the **first write** (`CREATE TABLE ...` → `SQLITE_IOERR_SHORT_READ`, system errno 95 = `EOPNOTSUPP`), at which point the connection is irreversibly corrupted — switching to DELETE mode afterwards also fails.

The existing comment on lines 156-162 suggests the author expected a graceful fallback:
> "SQLite silently keeps the prior mode if WAL can't be enabled"

But SQLite **does not** silently fall back — it reports WAL as active but then fails on writes.

### Testing
Confirmed via direct syscall test:
```
mmap 32KB MAP_SHARED: 0xffffffffffffffff errno=95 (Operation not supported)
```
And via SQLite test:
```
PRAGMA journal_mode=WAL → OK (returns "wal")
CREATE TABLE t(x)         → disk I/O error (SQLITE_IOERR_SHORT_READ)
```

### Proposed Fix
Pre-probe WAL support with a **separate** temporary database before configuring the real connection. If the probe fails, skip the WAL pragma entirely (SQLite defaults to DELETE mode).

**Changes to `lib/dist/db/index.js`:**

1. Add `walAvailable(dir)` function:

```js
function walAvailable(dir) {
    const probePath = path.join(dir, '_cg_wal_test.db');
    try {
        const { DatabaseSync } = require('node:sqlite');
        const probe = new DatabaseSync(probePath);
        probe.exec('PRAGMA journal_mode = WAL');
        probe.exec('CREATE TABLE p(x); DROP TABLE p');
        probe.close();
        fs.unlinkSync(probePath);
        try { fs.unlinkSync(probePath + '-wal'); } catch (_e) {}
        try { fs.unlinkSync(probePath + '-shm'); } catch (_e) {}
        return true;
    } catch {
        try { fs.unlinkSync(probePath); } catch (_e) {}
        try { fs.unlinkSync(probePath + '-wal'); } catch (_e) {}
        try { fs.unlinkSync(probePath + '-shm'); } catch (_e) {}
        return false;
    }
}
```

2. Modify `configureConnection` to accept `useWal` parameter:

```js
function configureConnection(db, useWal) {
    db.pragma('busy_timeout = 5000');
    db.pragma('foreign_keys = ON');
    if (useWal) {
        db.pragma('journal_mode = WAL');
        db.pragma('synchronous = NORMAL');
    } else {
        db.pragma('synchronous = FULL');
    }
    db.pragma('cache_size = -64000');
    db.pragma('temp_store = MEMORY');
    db.pragma('mmap_size = 268435456');
}
```

3. Update call sites in `initialize()` and `open()`:

```js
configureConnection(db, walAvailable(dir));
// or: configureConnection(db, walAvailable(path.dirname(dbPath)));
```

### Why a separate probe?
Once the main connection's WAL write fails, the database connection is corrupted — even `PRAGMA journal_mode = DELETE` won't recover it. A separate temporary DB isolates the test.

### Related
- #986 (custom storage directory) — a workaround, but auto-detection is the correct fix
- #987 (shared daemon) — DELETE mode means readers can block on writers; the `codegraph status` already reports the effective journal mode, so this is transparent to the daemon

### Tested
- ntfs3: probe correctly returns `false`, falls back to DELETE, init succeeds (36,554 nodes)
- btrfs/ext4: probe returns `true`, WAL active, normal operation
- Probe file cleanup: verified no `_cg_wal_test.*` residue after init

maxmilian:
Fixes #990.

## Problem

`configureConnection` (`src/db/index.ts`) sets `journal_mode = WAL` unconditionally. On filesystems where `mmap(MAP_SHARED)` returns `EOPNOTSUPP` on the WAL-index — **ntfs3, WSL2 `/mnt`, some CIFS/NFS** — `PRAGMA journal_mode = WAL` reports `"wal"` but the **first write** fails with `SQLITE_IOERR`, and the connection is already corrupted (switching to another mode afterwards also fails). The user-visible symptom is `codegraph init` dying with `disk I/O error`.

SQLite does **not** silently fall back here — it reports WAL as active and then fails on write.

## Fix

- Add `walAvailable(dir)`: opens a throwaway database in the target directory, runs `PRAGMA journal_mode = WAL` **plus a real `CREATE/DROP` write** (the write is what actually exercises the WAL-index mmap), and cleans up its probe files. Returns `false` on any failure.
- `configureConnection(db, useWal)` now takes an explicit flag. When `useWal` is `false` it falls back to `journal_mode = DELETE` with `synchronous = FULL` (NORMAL is only crash-safe under WAL).
- The two call sites (`initialize`, `open`) compute `useWal = walAvailable(dir)`.

The probe goes through the existing `createDatabase` adapter, so it honors whatever backend the connection would actually use.

> Note: the issue proposed editing the compiled `lib/dist/db/index.js`; this PR fixes the TypeScript source `src/db/index.ts` instead, which is where the build output comes from.

## Testing

`__tests__/wal-fallback.test.ts` (all green, 72 tests pass across the touched suites):
- `walAvailable` returns `true` on a normal writable dir and leaves no probe files behind;
- `walAvailable` returns `false` when the directory can't host a database;
- `configureConnection(db, true)` → `journal_mode = wal`;
- `configureConnection(db, false)` → `journal_mode = delete` **and writes still succeed** (the point of the fallback);
- `DatabaseConnection.initialize` produces a working WAL database on a normal fs.

**Honest limitation:** the actual `EOPNOTSUPP` path can't be reproduced on APFS (macOS), so these tests exercise the branching + the DELETE-mode write path rather than forcing a real mmap failure. The reporter's syscall trace (`mmap MAP_SHARED errno=95`) plus their ntfs3/btrfs/ext4 repro substantiate the failure mode itself. Happy to add a Linux/ntfs3 CI matrix probe if that's wanted.

### DIFF
--- a/__tests__/wal-fallback.test.ts
+++ b/__tests__/wal-fallback.test.ts
@@ -1,0 +1,73 @@
+import { describe, it, expect } from 'vitest';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as os from 'os';
+import { DatabaseConnection, walAvailable, configureConnection } from '../src/db';
+import { createDatabase } from '../src/db/sqlite-adapter';
+
+/**
+ * Regression coverage for #990: WAL mode must be probed before it is enabled,
+ * and the connection must fall back to DELETE mode (still writable) when the
+ * filesystem can't sustain WAL. The EOPNOTSUPP path itself (ntfs3 / WSL2 /mnt /
+ * some CIFS/NFS) cannot be reproduced on APFS, so these tests exercise the
+ * branching logic and the fallback's write path directly.
+ */
+describe('WAL fallback (#990)', () => {
+  it('walAvailable returns true on a normal writable dir and cleans up its probe', () => {
+    const dir = fs.mkdtempSync(path.join(os.tmpdir(), 'cg-wal-ok-'));
+    try {
+      expect(walAvailable(dir)).toBe(true);
+      const leftovers = fs.readdirSync(dir).filter((f) => f.startsWith('.cg-wal-probe-'));
+      expect(leftovers).toEqual([]);
+    } finally {
+      fs.rmSync(dir, { recursive: true, force: true });
+    }
+  });
+
+  it('walAvailable returns false when the dir cannot host a database', () => {
+    const missing = path.join(os.tmpdir(), `cg-wal-missing-${Date.now()}`, 'nested');
+    expect(walAvailable(missing)).toBe(false);
+  });
+
+  it('configureConnection(useWal=true) enables WAL', () => {
+    const dir = fs.mkdtempSync(path.join(os.tmpdir(), 'cg-wal-on-'));
+    try {
+      const { db } = createDatabase(path.join(dir, 't.db'));
+      configureConnection(db, true);
+      expect(db.pragma('journal_mode', { simple: true })).toBe('wal');
+      expect(db.pragma('synchronous', { simple: true })).toBe(1); // NORMAL, safe under WAL
+      db.close();
+    } finally {
+      fs.rmSync(dir, { recursive: true, force: true });
+    }
+  });
+
+  it('configureConnection(useWal=false) falls back to DELETE and stays writable', () => {
+    const dir = fs.mkdtempSync(path.join(os.tmpdir(), 'cg-wal-off-'));
+    try {
+      const { db } = createDatabase(path.join(dir, 't.db'));
+      configureConnection(db, false);
+      expect(db.pragma('journal_mode', { simple: true })).toBe('delete');
+      expect(db.pragma('synchronous', { simple: true })).toBe(2); // FULL, durable under DELETE
+      // The point of the fallback: writes must still succeed in DELETE mode.
+      db.exec('CREATE TABLE t (x); INSERT INTO t (x) VALUES (1);');
+      const row = db.prepare('SELECT x FROM t').get() as { x: number };
+      expect(row.x).toBe(1);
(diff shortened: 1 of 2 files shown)


maxmilian:
Closing to keep my open-PR footprint here honest — not because I think the change is wrong.

I have seven PRs open in this repo and only one merged (#663). This one has had no review activity since it was opened, and I'd rather leave a short queue of things with evidence behind them than a long one you have to page through. The two I'm keeping active have verified repros on current `main` (#594's import-resolution gap, and #597's `(AX_VOID)` case that survives the #1102 salvage).

Branch stays on my fork and this is trivially reopenable if it's ever useful.

PR #1184, opened 2026-07-06
fix(sync): avoid argument-stack overflow when gathering unresolved refs

ianmage:
## Summary

- `codegraph sync` no longer aborts with `Maximum call stack size exceeded` on large projects (fresh index, big Perforce/git changelist, or the first sync after switching branches).
- `getUnresolvedReferencesByFiles` chunked the file-path `IN`-list under SQLite's parameter limit, but then spread each chunk's matched rows into a single `rows.push(...chunkRows)`. One chunk can match hundreds of thousands of `unresolved_refs` rows — enough to blow V8's argument-stack limit. Chunking bounds the SQL params, not the returned row count, so rows are now appended in a loop instead of spread.
- Adds a regression test (`getUnresolvedReferencesByFiles (large result set)`) that inserts 130k refs sharing one `file_path` and asserts no throw.
- Adds a `### Fixes` entry under `[Unreleased]` in `CHANGELOG.md`.

## Test plan

- [x] `npm run build` passes (tsc + copy-assets).
- [ ] `npx vitest run __tests__/db-perf.test.ts` — new regression test passes.
- [ ] `npm test` — full suite green (no regressions in chunking/ref tests).


ianmage:
Verified that upstream already contains this fix: #1576 (commit d8f2eea) applies the identical loop-append to both readers in `src/db/queries.ts`, plus additional hardening in `src/bin/codegraph.ts` and `src/extraction/index.ts`, and pins it with `__tests__/refs-by-files-spread.test.ts` (200k-row scenario). Both regression tests from this PR were re-run against current `main` and pass. Closing as superseded by #1576 — thanks!

PR #1215, opened 2026-07-08
fix(extraction): detect untracked files inside untracked directories (#1213)

### LINKED ISSUE #1213: `codegraph sync` silently skips all untracked files
### Summary

When project files are untracked by git, `codegraph sync` always reports "Already up to date" and never detects changes. The file watcher is equally broken. Affects both v0.9.1 and v1.3.0.

### Root Cause

`collectGitStatus()` runs `git status --porcelain --no-renames` without `-uall`. Without it, git collapses entirely-untracked directories into a single entry like `?? frontend/` instead of listing individual files. The code then skips these directory entries, so no files are ever checked.

### Reproduction

```bash
git init && echo 'function foo() {}' > app.js
codegraph index
echo 'function bar() {}' >> app.js
codegraph sync   # → "Already up to date" — should detect the change
```

### Fix

One-line change in `extraction/index.js` (line 1037 in v1.3.0):

```diff
- ['git', 'status', '--porcelain', '--no-renames']
+ ['git', 'status', '--porcelain', '--no-renames', '-uall']
```

maxmilian:
## Summary

Fixes #1213. `codegraph sync` silently skipped source files that live inside an **untracked directory**, reporting "Already up to date" and never picking them up (the file watcher missed them too).

## Root cause

`git status --porcelain --no-renames` collapses an entirely-untracked directory into a single `?? frontend/` entry instead of listing the files inside it. `collectGitStatus` (`src/extraction/index.ts`) treats such `?? dir/` entries as *possible embedded git repos* and recurses via `findNestedGitRepos` — which only returns directories that are themselves git repos. A plain untracked directory has none, so its source files are never added to `changes.added`.

(The reporter's diagnosis was against v1.3.0, where the entry was simply skipped; on `main` the untracked-dir handling now exists but only covers embedded repos, so the same files still slip through.)

## Fix

Add `-uall` so git lists individual untracked files (`?? frontend/app.js`) rather than collapsing the directory. Verified that nested untracked git repos still collapse to `?? repo/` **even with `-uall`** — git never crosses a repo boundary — so the existing embedded-repo recursion is unaffected.

One-line behavior change; root-cause analysis and the `-uall` fix were suggested by the reporter in #1213.

## Tests

`getGitChangedFiles` is exported and covered by two regression tests:
- source files inside a fully-untracked directory are detected (the bug);
- an untracked embedded git repo is still recursed into (guards against an `-uall` regression).

Both pass; `tsc --noEmit` is clean; the existing `sync` / `multi-repo-workspace` / `worktree-detection` suites remain green.

### DIFF
--- a/__tests__/git-changed-untracked-dir.test.ts
+++ b/__tests__/git-changed-untracked-dir.test.ts
@@ -1,0 +1,68 @@
+/**
+ * Regression test for #1213: `codegraph sync` silently skips untracked files
+ * that live inside an untracked directory.
+ *
+ * `git status --porcelain` collapses an entirely-untracked directory into a
+ * single `?? frontend/` entry. getGitChangedFiles must still surface the source
+ * files inside it (via `-uall`) rather than dropping the whole directory.
+ */
+
+import { describe, it, expect, afterEach } from 'vitest';
+import { execFileSync } from 'child_process';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as os from 'os';
+import { getGitChangedFiles } from '../src/extraction/index';
+
+function git(cwd: string, args: string[]): void {
+  execFileSync('git', args, { cwd, stdio: 'pipe' });
+}
+
+describe('getGitChangedFiles — untracked directories (#1213)', () => {
+  const dirs: string[] = [];
+
+  function makeRepo(): string {
+    const dir = fs.mkdtempSync(path.join(os.tmpdir(), 'codegraph-1213-'));
+    dirs.push(dir);
+    git(dir, ['init']);
+    git(dir, ['config', 'user.email', 'test@example.com']);
+    git(dir, ['config', 'user.name', 'test']);
+    fs.writeFileSync(path.join(dir, 'root.js'), 'function foo() {}\n');
+    git(dir, ['add', 'root.js']);
+    git(dir, ['commit', '-m', 'init']);
+    return dir;
+  }
+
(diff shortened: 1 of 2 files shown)


maxmilian:
Status check before this drifts further: **the linked issue #1213 was closed on 2026-07-09, but the fix never landed.**

Re-verified against current `upstream/main` (`572d22b`, 2026-07-22):

- `src/extraction/index.ts:1108` is still `['status', '--porcelain', '--no-renames']` — no `-uall`, so `git` still collapses an entirely-untracked directory into a single `?? dir/` entry.
- That entry is still routed through the embedded-git-repo recursion, which by definition returns nothing for a plain untracked directory, so its source files are still never added to `changes.added`.
- No merged PR references #1213; searching merged PRs for `untracked` turns up #218, #217, #941, #949, #980, #1038, #1066 and #1208, all of which target embedded/nested repos, gitlinks or the ignore matcher — a different path from the `?? dir/` collapse.

The close event carries no commit and no comment, so I could not tell whether it was intentional. If #1213 was closed as `wontfix` or because it is covered by something I have missed, I am happy to close this PR — just say so and I will, no argument.

If it was closed by mistake, this branch is still green and merges cleanly; the change is 2 lines of behaviour plus a regression test that fails without it.

Not pinging for a review pass — only flagging that the issue state and the code state currently disagree, since a closed linked issue makes an open PR look obsolete when it is not.

Issue #1108, opened 2026-07-01
Local-variable receiver-type inference is C++-only — instance method calls through a local don't resolve in other languages


colbymchenry:
### Summary

Method calls through a **local variable** (`const lg = new Logger(); lg.log();`) only resolve to the method node in **C++**. In every other language the call produces no `calls`→method edge, because the resolver has no way to learn the receiver variable's type.

This is the coverage gap surfaced while fixing #1079 / #1107 (same-named-method mis-resolution). It's orthogonal to that fix: #1079 was about resolving to the *wrong* method; this is about not resolving at all.

### Evidence

Two-file / single-file repros of `useX() { var lg = <construct Logger>; lg.log(); }`, indexed and inspected:

| Call shape | Resolves to the method? |
|---|---|
| `lg.log()` (local variable) | **C++ only** — Python, TypeScript, Java, C#, Go, Ruby, Rust, Swift all produce 0 method edges |
| `Logger.log()` (class/static receiver) | all languages (fixed in #1107) |
| `Logger::log()` (qualified) | C++, Rust (fixed in #1107) |
| `new Logger()` / constructor | all languages (already correct — `instantiates` edge) |

So the constructor is resolved (we know the function makes a `Logger`), but `lg.log()` on the resulting variable is dropped everywhere but C++.

### Why it matters

`getCallers`/impact/blast-radius and `codegraph_explore` flow tracing miss any call routed through a local variable — the most common way instance methods are invoked. A method whose only callers use a local receiver looks uncalled.

### Root cause

`matchMethodCall` (`src/resolution/name-matcher.ts`) infers the receiver's type only for:
- C++ locals — `inferCppReceiverType` (reads source, matches the `Type name` declarator, handles `auto` + header scan), and
- Java/Kotlin **fields** — `inferJavaFieldReceiverType` (reads the indexed field node's signature).

There is no local-variable inference for any other language. Local variables are not extracted as nodes (by design — node-explosion), so this must be source-based like the C++ path.

### Proposed fix

[...]

PR #1221, opened 2026-07-08
fix(php): resolve method calls through $this-> properties on their declared type (#1220)

### LINKED ISSUE #1220: PHP: method calls through a class property (`$this->dep->method()`) never resolve — constructor-injected dependencies produce no call edges
## Summary

In PHP, a method call whose receiver is a **class property** — `$this->greeter->greet()` — never produces a `calls` edge, regardless of whether the property is typed with a concrete class or an interface, and regardless of declaration style (PHP 8 constructor property promotion or a classic typed property assigned in the constructor).

Receiver-type inference added in #1108 (locals) and #1125 (parameters) covers the other shapes: on 1.3.0, local-variable receivers (`$g = new Greeter(); $g->greet()`) and typed-parameter receivers (`function run(Greeter $g) { $g->greet(); }`) both resolve correctly. Class-property receivers are the missing case — and in dependency-injection codebases (Symfony, Laravel constructor injection) they are the *dominant* call shape: nearly every cross-class call goes through a constructor-injected property.

## Evidence

Minimal repro — 4 call sites of `Greeter::greet()`, identical except for how the receiver is obtained:

```php
// src/Greeter.php
class Greeter {
    public function greet(string $n): string { return "hi $n"; }
}

// src/UserViaNew.php — local variable
class UserViaNew {
    public function run(): string { $g = new Greeter(); return $g->greet('new'); }
}

// src/UserViaParam.php — typed parameter
class UserViaParam {
    public function run(Greeter $greeter): string { return $greeter->greet('param'); }
}

// src/UserViaDi.php — promoted constructor property
class UserViaDi {
    public function __construct(private readonly Greeter $greeter) {}
    public function run(): string { return $this->greeter->greet('di'); }
}

// src/UserViaClassicProp.php — classic typed property
class UserViaClassicProp {
    private Greeter $greeter;
    public function __construct(Greeter $greeter) { $this->greeter = $greeter; }
    public function run(): string { return $this->greeter->greet('classic'); }
}
```

`codegraph init && codegraph callers greet` (v1.3.0, also reproduced on 1.2.0):

```
Callers of "greet" (2):

method      run
  src/UserViaNew.php:3

method      run
  src/UserViaParam.php:3
```

| Receiver shape | Resolves? |
|---|---|
| local variable from `new` | ✅ (#1108) |
| typed function parameter | ✅ |
| **promoted constructor property** (`private readonly Greeter $greeter`) | ❌ |
| **classic typed property** (`private Greeter $greeter;` + constructor assignment) | ❌ |

Same-class calls (`$this->method()`) and static calls resolve fine; this is specifically the property-receiver hop.

## Why it matters

Measured on six real PHP codebases (340 → 9 200 files, PHP 7.1 → 8.5, Symfony 3.2 → 8.0): in DI-style code every cross-class production call goes through constructor-injected properties, so `codegraph_callers` returned **0 of 13** verified production call sites across the methods we first sampled, while happily returning the unit-test call sites (local `new` receivers). The result is worse than an empty answer: a method with production callers looks *test-only/dead*, and `codegraph_impact` reports only same-name declarations, so blast-radius on a signature change misses every real consumer. (Happy to share the detailed measurement protocol.)

I have a working fix (PR incoming) that follows the existing receiver-type-inference architecture; validated on those six codebases it takes a sampled repository method from 0/7 to 7/7 production callers with 49/49 manually verified call edges and no false positives — details in the PR.

## Root cause (from reading main)

1. **Extraction** (`src/extraction/tree-sitter.ts`, `extractCall`): for `member_call_expression` the receiver becomes the node's raw text with only the leading `$` stripped, so `$this->greeter->greet()` is recorded as the reference `this->greeter.greet`. Java has a dedicated unwrap for exactly this shape — `field_access` with a `this` object unwraps to the bare field name, with the comment "so the resolver can then look up in the enclosing class's field declarations" — but the condition only matches Java's `field_access`; PHP's `member_access_expression` never takes that path.
2. **Resolution** (`src/resolution/`): nothing splits `->`, and nothing consults class property declarations for PHP — so `this->greeter` is an unmatchable receiver string and the reference is dropped.

The type information is statically available: promoted properties carry it in the constructor signature, classic typed properties (PHP 7.4+) in the declaration, and the PHP extractor already collects `property_declaration` nodes (`fieldTypes` in `src/extraction/languages/php.ts`) — it's just not used for receiver typing.

## Relationship to existing work

- #1108 (locals) and #1125 (parameters) fixed the sibling shapes; this is the third leg.
- CFML got exactly this in 1.3.0 (component properties, pseudoconstructor pattern — per the changelog "the same receiver-type inference the other object-oriented languages already had").
- PR #920 (open) implements `this.<field>.method()` resolution on the field's declared type for **TS/JS** — this issue is the PHP counterpart.

Environment: CodeGraph 1.2.0 and 1.3.0 (npm), macOS arm64, PHP grammar as bundled.

w0lan:
Fixes #1220

## Problem

A PHP method call whose receiver is a class property — `$this->dep->method()`, the dominant call shape in constructor-injection codebases (Symfony, Laravel) — produces no `calls` edge. The extractor records the receiver as raw text (`this->dep`), which no resolution strategy can type, so the reference is dropped. `codegraph_callers` then reports a heavily-used method as uncalled (or test-only, since unit tests calling through local `new` receivers DO resolve), and impact analysis misses every production consumer. Full evidence and measurement protocol in #1220.

## Fix

Three coordinated pieces, all riding existing machinery (no schema or extraction changes):

1. **`inferLocalReceiverType`** (`name-matcher.ts`): strip the PHP `this->` prefix and widen the scan to the whole file — the same treatment CFML component-scoped fields (`variables.svc` / `this.svc`) already get in this function. The existing PHP typed-parameter pattern (`Foo $prop`) then recovers the type from a **promoted constructor parameter**, a **typed property declaration**, or a **classic constructor parameter** alike; nearest-declaration-backward still lets a shadowing local win.
2. **`matchMethodCall`** (`name-matcher.ts`): resolve the `this->prop.method` shape **exclusively** through declared-type inference + `resolveMethodOnType` validation at confidence 0.9 (the same tier as the #1108/#1125 typed paths). The name-similarity strategies below never see this shape, so a property whose type can't be recovered statically (docblock-only, setter/container injection) stays unlinked rather than guessed — a wrong inference produces no edge rather than a wrong one.
3. **Defer + conformance retry** (`resolution/index.ts`): a method the property's type inherits from a supertype is resolvable only once `implements`/`extends` edges exist, so refs matching `PHP_PROP_SHAPE` are pushed to the existing deferred-chain retry and dispatched back through `matchMethodCall` in the conformance pass — the #750/#808 lifecycle, no new wiring.

## Validation

- **15-case corpus**: promoted / classic / interface-typed / parent-class / trait properties, inherited method (via the conformance retry), two same-named methods on unrelated classes (edge goes to the property's declared type, not the first-indexed homonym), local variable shadowing a property (#1108 path still wins), untyped docblock-only property (deliberately no edge), deep chain `$this->a->b->m()` (deliberately no edge, parity with other languages).
- **9 new vitest tests** (`__tests__/php-property-receiver-resolution.test.ts`), modeled on the CFML receiver-inference tests; full suite passes.
- **Six real PHP codebases** (340–9 200 files, PHP 7.1–8.5, Symfony 3.2–8.0): a sampled repository method went **0/7 → 7/7** production callers; **49/49 manually verified** property-receiver edges correct across five typed codebases, zero false positives in random samples of confidence-0.9 edges; +6 100 call edges on a 5 000-file Symfony repo (+3.8%). On an untyped symfony-1.x legacy codebase the fix simply doesn't fire (no recoverable types) — no new wrong edges there either.

## Known limitations (deliberately out of scope, candidates for follow-ups)

- Fluent/instance chains `$this->prop->inner()->outer()` — the first hop now resolves, the second needs the `<inner>().<method>` re-encoding other languages got in #645/#608.
- Nullsafe calls (`$this->prop?->method()`) — `nullsafe_member_call_expression` isn't in the PHP extractor's `callTypes` at all (affects every nullsafe call, not just property receivers).
- Types available only in `@var` docblocks, or injected via setters/containers, stay unlinked by design.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixed
+
+- PHP method calls made through a class property — `$this->dep->method()`, the dominant call shape in constructor-injection codebases (Symfony, Laravel) — now resolve to the method on the property's declared type, so `codegraph_callers` and impact analysis see production callers instead of reporting a DI-heavy method as uncalled or test-only. All three declaration shapes count: a promoted constructor parameter (`private readonly Foo $dep`), a typed property (`private Foo $dep;`), and a classic constructor parameter assigned in `__construct`. Interface-typed properties resolve to the interface method, and a method the property's type inherits from a supertype resolves through the existing conformance retry once `extends`/`implements` edges are built. Resolution is deliberately exclusive: a property whose type can't be recovered statically (docblock-only, setter/container injection) stays unlinked rather than guessed, so no name-similarity false edges are introduced. (#1220)
 
 ## [1.3.0] - 2026-07-07
 
--- a/__tests__/php-property-receiver-resolution.test.ts
+++ b/__tests__/php-property-receiver-resolution.test.ts
@@ -1,0 +1,225 @@
+/**
+ * PHP property-receiver resolution (#1108 family).
+ *
+ * `$this->prop->method()` reaches the resolver as `this->prop.method` (the
+ * extractor records the receiver's raw text with the leading `$` stripped, and
+ * — unlike a `foo()->bar()` chain — there are no `()` on the receiver). The
+ * property's declaration lives OUTSIDE the calling method: a promoted
+ * constructor parameter (`private readonly Greeter $greeter`), a classic typed
+ * property assigned in `__construct`, or a property typed by an interface. The
+ * resolver recovers the property's declared type (widening the local-receiver
+ * scan to the whole file, the treatment component-scoped fields already get)
+ * and validates it through `resolveMethodOnType`, so a property whose type
+ * can't be recovered stays UNLINKED rather than guessed — a wrong inference
+ * produces no edge instead of a wrong one.
+ *
+ * Method lookup runs EXCLUSIVELY through declared-type inference: the
+ * name-similarity fallbacks never see this shape, which is what makes the
+ * same-name-collision and no-type cases below negative. Inherited methods
+ * resolve only once `extends`/`implements` edges exist, so these refs defer to
+ * the conformance pass; the full `indexAll()` path here exercises that.
+ */
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+import * as os from 'node:os';
+import { CodeGraph } from '../src';
+import { Node } from '../src/types';
+import { ResolutionContext } from '../src/resolution';
+import { matchMethodCall } from '../src/resolution/name-matcher';
+import type { UnresolvedRef } from '../src/resolution/types';
+
+describe('PHP property-receiver resolution', () => {
+  let dir: string;
+  beforeEach(() => { dir = fs.mkdtempSync(path.join(os.tmpdir(), 'php-prop-recv-')); });
+  afterEach(() => { fs.rmSync(dir, { recursive: true, force: true }); });
+
+  const write = (rel: string, body: string) => {
+    const p = path.join(dir, rel);
+    fs.mkdirSync(path.dirname(p), { recursive: true });
+    fs.writeFileSync(p, body);
+  };
+
+  const load = async () => {
+    const cg = await CodeGraph.init(dir, { silent: true });
+    await cg.indexAll();
+    const db = (cg as any).db.db;
+    const calls: { src: string; tgt: string; tgtQn: string }[] = db
+      .prepare(
+        `SELECT s.name src, t.name tgt, t.qualified_name tgtQn
+         FROM edges e JOIN nodes s ON s.id = e.source JOIN nodes t ON t.id = e.target
+         WHERE e.kind = 'calls' AND t.kind = 'method'`,
+      )
+      .all();
+    cg.close?.();
+    return calls;
+  };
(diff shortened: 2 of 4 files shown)


colbymchenry:
Thank you for this — the report in #1220 was one of the best-diagnosed issues we've received (the receiver-shape table and the root-cause trace were both exactly right), and the fix followed the receiver-inference architecture precisely where it should.

We wanted this shipped without another review round-trip, so it landed via #1251, which carries your branch verbatim plus one hardening commit, with co-author credit to you in the merge commit and the changelog. The hardening: the whole-file scan reused the generic local-variable patterns (the CFML component-scope treatment), but in PHP a plain `$prop` local or parameter can never alias `$this->prop` — so a same-named variable in another method could mistype the property and mint a wrong 0.9-confidence edge (e.g. a helper's `$greeter = new OtherGreeter()` routing `$this->greeter->greet()` to `OtherGreeter::greet`). Property receivers now consult property-shaped declarations only, with an assignment-following fallback for classic untyped properties that keeps your multi-line-constructor and setter-injection coverage — details and the interference regression tests are in #1251.

Your three follow-up candidates (fluent second hops, nullsafe `?->` in the extractor's callTypes, docblock types) are all real — the nullsafe one especially, since it affects every `?->` call. PRs very welcome on any of them.

Closing in favor of #1251 — thanks again, this fixes the dominant call shape for every DI-style PHP codebase.

Issue #1238, opened 2026-07-10
Upgrading CodeGraph leaves old agent instructions in place


xuing:
## Problem

`codegraph upgrade` updates the binary, but does not refresh the files previously written by `codegraph install`.

I upgraded a long-lived install from 0.9.x to 1.3.1, but the CodeGraph sections in `CLAUDE.md` and `AGENTS.md` still told agents to call `codegraph_context`. That tool was removed in 1.0.0, so agents now get an “unknown tool” error and often stop using CodeGraph.

Re-running `codegraph install` fixes the sections, but users have no reason to know that they need to do this after an upgrade.

## Expected behavior

After a successful upgrade, CodeGraph should refresh the installer-owned instructions and MCP config for agents that are already configured. It should not configure new agents or change the user's permission and hook choices.

Because the instruction templates are bundled with the binary, the refresh needs to run through the newly installed binary.

PR #1239 implements this behavior.

## Environment

- CodeGraph 1.3.1, installed through npm
- Ubuntu linux-x64
- Upgraded from a 0.9.x install

Issue #1071, opened 2026-06-30
The version and the “--version” display are inconsistent


SpringYear:
<img width="1006" height="161" alt="Image" src="https://github.com/user-attachments/assets/34c2606c-bffa-4798-8558-ced48d396a2c" />
The version and the “--version” display are inconsistent

colbymchenry:
Thanks for the report — and good catch, that mismatch is confusing.

The good news is your install isn't broken: the `v1.1.4` bundle the installer downloaded really is `1.1.4`. What's happening is that a **different, older `codegraph` earlier on your PATH** is the one that actually runs when you type `codegraph`, and it reports `0.9.2`.

The usual cause is a leftover global npm install from an earlier version (`npm i -g @colbymchenry/codegraph`). That npm package launches its own version-pinned copy, so it keeps printing `0.9.2` no matter what the standalone installer drops in — and because it sits earlier on your PATH (often via a conda/node bin dir — your prompt shows `(base)`), it shadows the new install.

**To confirm and fix it on your machine:**

```powershell
# See every codegraph on your PATH (the first one is what runs):
Get-Command codegraph -All | Select-Object Source

# Remove the stale global npm copy:
npm rm -g @colbymchenry/codegraph
```

(Alternatively, put the standalone install's bin — `%LOCALAPPDATA%\codegraph\current\bin` — first on your PATH.) Open a fresh terminal afterward and `codegraph --version` should report the installed version.

**What we changed:** the installers (`install.sh` / `install.ps1`) now detect this at install time and warn you when another `codegraph` earlier on your PATH would run instead, naming the shadowing copy and how to resolve it — so this won't be a silent surprise anymore. Merged in #1072; it'll be in the next release.

Issue #303, opened 2026-05-22
no prebuilt bundle for darwin-arm64


zhanglq365:
After using npm i -g @colbymchenry/codegraph to install the package, codegraph init -i will report an error, as shown in the figure:

<img width="845" height="160" alt="Image" src="https://github.com/user-attachments/assets/dd406a3b-9adf-4dbf-a7ba-7673d451222c" />

node: v24.15.0
npm: 11.12.1

Emiyaaaaa:
```bash
npm i -g @colbymchenry/codegraph-darwin-arm64
```

zhanglq365:
> npm i -g @colbymchenry/codegraph-darwin-arm64

<img width="1238" height="236" alt="Image" src="https://github.com/user-attachments/assets/cddae1c8-0d07-4a79-b64f-f7d71c4ffcd5" />

There is still a problem

cienetzzm:
I also encountered this problem.

colbymchenry:
Root cause found and fixed. 🎯

This wasn't a packaging bug — all six per-platform packages are published correctly on the official npm registry. The problem is that the **npmmirror / cnpm mirror had not mirrored the per-platform package** (`@colbymchenry/codegraph-darwin-arm64`, etc.). The runtime ships as an *optional* dependency, and npm silently skips an optional package it can't download from the mirror — so `codegraph` ends up installed with no bundle → `no prebuilt bundle for darwin-arm64`. (I reproduced this end-to-end: a clean install from `registry.npmmirror.com` installs only the launcher, while the same install from `registry.npmjs.org` works.)

**Fixed right now:** I triggered a sync of all 6 platform packages on npmmirror — they're live there for both `0.9.2` and `0.9.3`. A fresh install should work again:

```
npm i -g @colbymchenry/codegraph
```

If your mirror still serves a stale cache, either install from the official registry:

```
npm i -g @colbymchenry/codegraph --registry=https://registry.npmjs.org
```

…or use the standalone installer (no Node/npm required):

```
curl -fsSL https://raw.githubusercontent.com/colbymchenry/codegraph/main/install.sh | sh
```

[...]

PR #1245, opened 2026-07-10
fix(upgrade): stop npm installs self-shadowing on upgrade; verify the resolved version after every upgrade

colbymchenry:
Companion to #1239 / #1238. Together they make three guarantees hold after `codegraph upgrade`: hooks refreshed, agent instruction files refreshed (#1239), and — this PR — `codegraph -v` in the **same terminal** provably serving the new version.

## Problem 1 — npm installs self-shadow on upgrade (found while validating #1239)

Reproduced live in a sandboxed `$HOME`: install 1.3.1 via `npm i -g`, run `codegraph upgrade` — and `codegraph --version` reports **1.3.1 forever**, in that terminal and every future one.

Root cause: the npm thin-installer's per-platform package (`@colbymchenry/codegraph-<platform>-<arch>`) is itself a complete bundle — vendored `node` + `bin/` launcher — living inside `node_modules`. `detectInstallMethod` checked the bundle *layout* before the `node_modules` *path*, so every npm install misdetected as a standalone bundle and `upgrade` curled install.sh into `~/.codegraph`: a **second install** that never wins the PATH race against npm's shim. The #1071 shadow pile-up, manufactured by `upgrade` itself.

This also silently broke #1239 for npm users (the issue author's own environment): its post-upgrade `codegraph install --refresh` spawn resolves the stale npm shim, which doesn't know `--refresh`.

**Fix:** path-based checks (`/_npx/`, `/node_modules/`) now run before the layout sniff — where the file lives is authoritative about *how* the user installed, whatever the artifact inside looks like. npm installs upgrade through `npm install -g …@latest` again: same shim path, so the same terminal picks up the new version immediately. The shim's GitHub-download self-heal cache (`~/.codegraph`, issue #303) is outside `node_modules` and still detects as a bundle.

## Problem 2 — nothing ever verified the upgrade took

The unix-bundle path ended with "✓ Upgrade complete. *Open a new terminal if the version looks unchanged (PATH cache)*" — an unconditional hedge for a failure mode we can detect. After a successful swap, `runUpgrade` now probes the PATH-resolved `codegraph --version` (through `cmd.exe` on Windows for the `.cmd` launcher) and reports the real outcome:

- **match** → `✓ codegraph on your PATH now reports v1.4.0 — this terminal is already using it.`
- **mismatch** → a loud warning that a stale install is shadowing the upgraded one, with `which -a codegraph` / `where codegraph` to find it. Warn-only; the upgrade itself still exits 0.
- **inconclusive** (no `codegraph` on PATH, probe fails or output unparsable) → the old soft new-terminal hint, never a scare.

Skipped for npm-local installs, whose binary PATH never serves. New `capture` dep on `UpgradeDeps` (piped stdio, 30s timeout) keeps it unit-testable; probe parsing takes the last non-empty stdout line so a stray runtime warning can't spoil it.

## Validation

- **3 detection regression tests** (thin-installer platform package → npm global / npm local / npx-cached) — verified to **fail against the pre-fix code** and pass now.
- **8 probe orchestration tests**: match / mismatch / both inconclusive shapes / last-line parsing / Windows `cmd.exe` routing / npm-local skip / no probe after a failed upgrade.
- Full suite: 2290 passed, 4 skipped (platform-gated).
- Live e2e against real binaries in a sandboxed `$HOME`: bundle 1.3.1 → upgrade → same-shell `-v` = 1.4.0 with the green confirmation; simulated shadow prints the warning. (The npm misdetection itself was reproduced live pre-fix: upgrade curled install.sh and left the shim at 1.3.1.)

## Interaction with #1239

Independent — either can merge first; whichever lands second has a trivial conflict in `runUpgrade`'s post-success block and the CHANGELOG. Ordering note for the rebase: the version probe should run (and ideally gate) *before* the `install --refresh` spawn, so the refresh never executes through a shadowed stale binary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixes
 
+- `codegraph upgrade` on an npm install now upgrades through npm again instead of quietly creating a second copy that never wins the PATH race — previously `codegraph --version` kept reporting the old version forever, no matter how many times you upgraded. (#1238)
+- After every upgrade, CodeGraph now checks that the `codegraph` command your terminal resolves actually serves the freshly installed version — confirming you don't need a new terminal, or telling you exactly which stale install is shadowing the new one. (#1071)
 - The safety watchdog no longer kills a healthy index on severely degraded storage. It used to judge liveness purely by the event loop, so one long database write on a struggling disk looked identical to a hung process and could get a valid, in-progress index terminated. During `codegraph index`/`codegraph init` the watchdog now also checks whether the index files on disk are advancing before it acts: slow-but-progressing work is left alone (bounded by a hard cap), while a genuinely hung process is still killed exactly as fast as before. (#1231)
 
 ## [1.4.0] - 2026-07-10
--- a/__tests__/upgrade.test.ts
+++ b/__tests__/upgrade.test.ts
@@ -13,6 +13,7 @@ import {
   parseLatestTagFromLocation,
   reindexAdvisory,
   runUpgrade,
+  verifyResolvedVersion,
   buildWindowsUpgradeScript,
   NPM_PACKAGE,
   type InstallMethod,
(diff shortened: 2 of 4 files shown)


colbymchenry:
## Cross-platform validation

**Linux (Docker, `node:22-bookworm`, `--rm --init`)**
- Targeted suites: 194/194. Full suite: **2290 passed, 4 skipped (platform-gated), 0 failed**.
- Live e2e in the container: real `install.sh` bundle 1.3.1 → `codegraph upgrade` → same-shell `codegraph --version` = **1.4.0**.
- Probe harness with the real `defaultCapture`/`hasCommand` against the PATH-resolved binary: match prints the green confirmation; a forced mismatch prints the shadow warning.

**Windows (Parallels VM, win32-arm64, Node 24, fresh `npm ci` at `C:\dev\codegraph` on b108500)**
- Targeted suites: **194/194** (`upgrade` 43, `installer-targets` 151).
- Live probe against a *naturally occurring* #1071 shadow found on the VM: a stale **0.9.6 bundle** launcher (`AppData\Local\codegraph\current\bin\codegraph.cmd`) sits earlier on PATH than a fresh 1.4.0 npm install, so `codegraph --version` genuinely reports 0.9.6.
  - Target v1.4.0 → **mismatch**: the shadow warning fired on a real-world true positive.
  - Target v0.9.6 → **match**: green "this terminal is already using it."
  - This exercised the `cmd.exe /d /s /c` routing, the PATHEXT `hasCommand` scan, and CRLF stdout parsing against a real `.cmd` launcher.

Issue #1247, opened 2026-07-10
C++: explicit a.operator+(b) call gets no `calls` edge to the operator method


Dshuishui:
> **Scoped** to the explicit `a.operator+(b)` form (per the discussion below). The infix `a + b` and subscript `a[i]` cases — which need receiver type inference — are tracked separately in #1258.

**codegraph version:** 1.2.0

### Summary

For C++, operator-overload methods (`operator+`, `operator/`, `operator[]`, …) are indexed as `method` nodes, but I couldn't find any `calls` edge pointing to them — whether they're invoked via infix `a + b`, subscript `a[i]`, or even the explicit `a.operator+(b)` call syntax. A structurally identical normal method call (`a.get()`) does get an edge, so this looks specific to operator overloads. Possibly expected behavior, but it surprised me — wanted to check.

### Reproduce

`optest.cpp`:

```cpp
struct V {
    int x;
    V operator+(const V& o) const { return V{x + o.x}; }
    V operator[](int i) const { return V{x + i}; }
    int get() const { return x; }
};

// positive control: plain named method call  -> edge present
int plainCaller(const V& a) { return a.get(); }

// infix operator use            -> edge not found
V infixCaller(const V& a, const V& b) { return a + b; }

// explicit operator-call syntax -> edge not found
V explicitCaller(const V& a, const V& b) { return a.operator+(b); }

// infix subscript               -> edge not found
V subscriptCaller(const V& a) { return a[3]; }
```

```bash
codegraph init
DB="$(readlink -f .codegraph)/codegraph.db"

# indexed functions/methods
sqlite3 "$DB" "SELECT name FROM nodes WHERE kind IN ('function','method');"

# call edges (caller -> callee)
sqlite3 "$DB" "SELECT s.name, t.name FROM edges e
               JOIN nodes s ON e.source = s.id
               JOIN nodes t ON e.target = t.id
               WHERE e.kind = 'calls';"

# edge-kind counts
sqlite3 "$DB" "SELECT kind, count(*) FROM edges GROUP BY kind;"
```

### Observed

- The `nodes` query lists all seven functions/methods, including `operator+` and `operator[]`.
- The `calls` query returns only `plainCaller | get`.
- Edge-kind counts: `calls | 1`, `contains | 8`.

So `operator+` and `operator[]` are present as `method` nodes but don't appear as the target of any `calls` edge, including from the explicit `a.operator+(b)` call.

### Expected

[...]

Issue #1258, opened 2026-07-12
C++: operator overloads via infix (a + b) / subscript (a[i]) get no call edge (needs receiver type inference)


Dshuishui:
Split out from #1247 per @aznikline's scoping there. #1247 now covers the explicit `a.operator+(b)` form; this issue tracks the other two cases, which share a root cause: they need receiver type inference.

### codegraph version
1.2.0

### Cases
For a C++ overloaded operator, no `calls` edge is recorded into the operator method when it is invoked via:
- infix — `a + b` (parses as `binary_expression`, which isn't in C++'s `callTypes`, so it's never treated as a call site)
- subscript — `a[i]` (parses as `subscript_expression`, same shape)

### Repro (`optest.cpp`, then `codegraph init`)
```cpp
struct V {
    int x;
    V operator+(const V& o) const { return V{x + o.x}; }
    V operator[](int i) const { return V{x + i}; }
    int get() const { return x; }
};
int plainCaller(const V& a)             { return a.get(); }  // edge present (control)
V   infixCaller(const V& a, const V& b) { return a + b; }    // no edge to operator+
V   subscriptCaller(const V& a)         { return a[3]; }     // no edge to operator[]
```
```sql
SELECT s.name, t.name FROM edges e
  JOIN nodes s ON e.source = s.id
  JOIN nodes t ON e.target = t.id
  WHERE e.kind = 'calls';
```
Observed: only `plainCaller | get`; no edge for `infixCaller -> operator+` or `subscriptCaller -> operator[]`.

### Note
Per @aznikline's analysis in #1247, resolving these needs knowing the receiver's type and that it declares the operator (receiver type inference), which the resolver doesn't do today — and the #1230 discussion flags that route as drift-prone. Filing separately so the contained explicit-form fix (#1247) isn't blocked on it, and so the common real-code form (infix) stays tracked for call-graph consumers.

PR #1260, opened 2026-07-12
fix(extract): resolve explicit C++ operator-overload calls (#1247 case 1)

aznikline:
## What

Closes the **explicit-form half of #1247**: a call to an operator-overload method written in its explicit `a.operator+(b)` form never recorded a call edge, so the `operator+` method node had zero callers and its impact radius was invisible.

This is **case (1)** from the scoping discussion in the issue. The harder cases — infix `a + b` and subscript `a[i]` — parse as `binary_expression` / `subscript_expression` (not `call_expression`) and need receiver type inference from an expression, which @Dshuishui split off into #1258. This PR does **not** touch those.

## Root cause

`tree-sitter-cpp` can't parse the `.operator<sym>` member access. For every explicit form — `a.operator+(b)`, `a.operator[](i)`, `a.operator==(b)` — the AST comes out as:

```
call_expression [a.operator+(b)]
  identifier [a]            ← namedChild(0) = receiver
  ERROR [.operator+]       ← wraps operator_name
    operator_name [operator+]
  argument_list [(b)]
```

The `call_expression` has **no `function` field** and an **ERROR node** wrapping `operator_name`. So the extractor's generic path took `namedChild(0)` (the receiver `a`) as the callee and emitted a useless `calls:"a"` ref — the `operator+` method node was never linked.

The `operator+` method node itself is indexed correctly on the definition side (named `operator+`), so this is purely a call-site extraction + resolution gap.

## Fix (C/C++ only, both parts gated)

**Extractor** (`tree-sitter.ts`): before the generic `namedChild(0)` fallback, detect a `call_expression` whose subtree contains an `operator_name` (nested under the ERROR node), read it as the callee, and qualify it with a simple-identifier receiver → `a.operator+`. For a non-identifier receiver (call result, deref, …) with no static name to route through inference, emit the bare `operator+` so `matchByExactName` still links it when the overload name is unique in scope.

**Resolver** (`name-matcher.ts`): the `recv.method` / `Class::method` method-segment gate was `\w+`, which excludes `+`/`[`/`(` — so even after the extractor emits `a.operator+`, the pattern dropped it before receiver-type inference could run. Widen the method segment to also accept `operator` + an overload-symbol tail, C/C++ only. The symbol set is the C++ overloadable operators; an unbounded `[^.]+` would over-match, so it's enumerated.

The receiver-type inference chain (`inferCppReceiverType` → `resolveMethodOnType`, which validates the method exists on the inferred type) was already in place — only the extraction gap and the pattern gate stopped it.

## Scope

- Covers: explicit `a.operator+(b)`, `a.operator[](i)`, `a.operator==(b)`, and the rest of the overloadable-operator set, in both `.method()` and `Class::method` shapes.
- Does **not** cover: infix `a + b`, subscript `a[i]` — those need expression-level receiver type inference (#1258).
- Safe by construction: the extractor branch only fires when an `operator_name` actually exists in the subtree, and only ever emits `operator`-prefixed names (no ordinary call site produces those); the resolver widening is language-gated to C/C++.

## Test

Added to `__tests__/resolution.test.ts`:

- `resolves an explicit a.operator+(b) call to the operator+ method (#1247 case 1)` — end-to-end: indexes a `Vec` with `operator+` plus an `add(Vec, Vec)` that calls `a.operator+(b)`, and asserts `add` is a caller of the `operator+` method.

Before the fix this emitted `calls:"a"` and resolved to nothing; after, it links to `operator+`.

Full suite green: **2363 passed, 4 skipped, 0 failed** (build + `vitest run`).

### DIFF
--- a/__tests__/resolution.test.ts
+++ b/__tests__/resolution.test.ts
@@ -1637,6 +1637,40 @@ func main() {
       expect(callTargets('useB')).toContain(logB.id);
     });
 
+    it('resolves an explicit a.operator+(b) call to the operator+ method (#1247 case 1)', async () => {
+      // The repro from #1247: a call to an operator-overload method in its
+      // explicit `a.operator+(b)` form never recorded a call edge. The `operator+`
+      // method node IS indexed (definition side is fine), but the resolver's
+      // method-call pattern `recv.method` used `\w` for the method segment, which
+      // excludes `+` — so `a.operator+` matched no pattern and `matchByExactName`
+      // couldn't bridge `a.operator+` to a node named `operator+`. With a receiver
+      // whose type is locally declared (`Vec a;`), the inference chain already
+      // works; only the pattern gate stopped it. This is the explicit-form case
+      // only — infix `a + b` and subscript `a[i]` parse as binary_expression /
+      // subscript_expression (not call_expression) and are tracked separately in
+      // #1258 (they need receiver type inference from an expression, not a name).
+      fs.writeFileSync(
+        path.join(tempDir, 'vec.cpp'),
+        'class Vec { public: Vec operator+(const Vec& o) const { return o; } };\n'
+          + 'Vec add(Vec a, Vec b) { return a.operator+(b); }\n',
+      );
+
+      cg = await CodeGraph.init(tempDir, { index: true });
+      cg.resolveReferences();
+
+      const opPlus = cg.getNodesByKind('method').find((n) => n.name === 'operator+');
+      const add = cg.getNodesByKind('function').find((n) => n.name === 'add');
+      expect(opPlus).toBeDefined();
+      expect(add).toBeDefined();
+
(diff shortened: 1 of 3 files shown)


aznikline:
Closing this in favor of #1268, which @colbymchenry just merged (Fixes #1247).

The merged fix goes further than this PR — it also handles pointer receivers (`p->operator+` → `p.operator+`), the `operator()` call form, and spaced call-site normalization (`operator ==` → `operator==` to match the compact definition form), plus correct handling of complex receivers that can't be type-inferred. That's the better implementation; no reason to keep this duplicate open.

The scope probe and root-cause analysis in the issue still did their job — glad to see the gap closed. I'll keep an eye on #1258 (the infix `a + b` / subscript `a[i]` cases, which need expression-level type inference) in case I can help there.

Issue #967, opened 2026-06-23
Tracking: framework-aware route & edge coverage requests


colbymchenry:
CodeGraph's value on framework code comes from connecting flows that have **no static call edge** — a route to its handler, a dispatch to its listener, a client call to its provider. We add these as framework resolvers / edge synthesizers (Express, Laravel, Rails, FastAPI, Spring, Gin, GoFrame, the dispatch-synthesizer family, …), each validated on real repos before it ships.

This issue **consolidates the individual framework-coverage requests** so they don't fragment the tracker ahead of the next releases. The individual issues are being closed and folded in here — 👍 a row below to signal demand, and PRs are very welcome (start from the closest existing resolver/synthesizer; see `docs/design/`).

**Requested framework coverage (not yet shipped):**

| Framework / pattern | Issue(s) | What's missing |
|---|---|---|
| Hono sub-routers | #307 | Routes defined on a `new Hono()` sub-router and mounted via `app.route('/prefix', sub)` aren't extracted (~82% miss). Only the mount point + direct `app.*` routes are picked up. Express-like, so close to existing support. |
| Java / MyBatis — statement-id calls | #489 | Bridge `getSqlSessionTemplate().delete(SQL_NS + ".deleteByOrderId", …)` to the XML `<delete id="…">` statement, resolving the statically-recoverable id (literal / `Class.getName()` / `+` concat). |
| Java / MyBatis — resultMap & `<sql>` | #592 | Index `<resultMap>` and `<sql>` fragments so a column/field can be traced back to the statements that reference it. |
| Java / Spring — OpenFeign → controller | #490 | Synthesize an edge from a `@FeignClient` method to the `@RestController` route with the matching HTTP path, across service modules. |
| Java / Spring — `@ConditionalOnProperty` | #491 | Emit a config-key reference from `@ConditionalOnProperty(prefix, name, …)` the way `@Value` / `@ConfigurationProperties` already do. |

Prioritization is roughly by 👍 + PR availability. Each new synthesizer/resolver is validated on ≥2 real repos plus a control before it lands.

This is distinct from the other consolidation trackers: **#648** (language support), **#649** (agent / IDE install targets), **#750** (chained factory/singleton call resolution).

KnifeOfLife:
+1, reported concrete-class @Resource field injection case in #1187

antoineFrau:
## Route-extraction backlog (plugin issues)

[...]

Issue #648, opened 2026-06-02
Tracking: language support requests (post-1.0 roadmap)


colbymchenry:
codegraph parses code with **tree-sitter**, so adding a language is mostly: wire a grammar `.wasm` + a small per-language extractor, then validate retrieval on real repos. There's an `/add-lang` workflow and a contributor cookbook (see #97 / `docs/ADDING-A-LANGUAGE.md`).

This issue **consolidates the individual language-support requests** so they don't fragment the tracker ahead of 1.0. The individual issues are being closed and folded in here — 👍 a language below to signal demand, and PRs are very welcome (start from the closest existing extractor).

**Currently supported:** TypeScript/JavaScript, ArkTS (HarmonyOS), Python, Go, Rust, Java, Kotlin, Scala, C/C++, CUDA, C#, Visual Basic .NET, Swift, Objective-C, Metal, PHP, Ruby, Dart, Lua/Luau, R, Erlang, CFML, COBOL, Solidity, Terraform/OpenTofu (HCL), Nix, Svelte, Vue, Astro, Razor/Blazor, Liquid, Pascal/Delphi.

**Requested (not yet supported):**

| Language | Issue(s) | Notes |
|---|---|---|
| BrightScript | #559 | |
| ReScript | #524 | |
| PLC Structured Text (SCL) | #441 | |
| Lazarus / Free Pascal | #340 | Delphi (`.pas` / `.dfm`) is already supported; Lazarus FPC is close — please test current behavior |
| Verilog / SystemVerilog | #262 | |
| Julia | #229 | |
| Fortran | #200 | |
| Groovy | #169 | |
| dbt (Databricks) | #88 | |
| SQL | #85 | |
| HTML / Jinja2 templates | #431 | FastAPI server-side rendering |
| AUTOSAR C/C++ (heavy macros) | #557 | More a C/C++ macro-expansion gap than a new grammar |
| GLSL | #685 | Graphics shaders; mature community tree-sitter grammars exist |
| QML | #759 | Qt C++/QML apps — the C++↔QML bridge (`qmlRegisterType`, signals/slots) is the valuable part |
| ABAP / abapGit | #694 | Serialized `.clas.abap` / `.intf.abap` / `.prog.abap` objects |
| Slint | #916 | |
| MQL5 | #55 | MetaTrader; C++-like, reporter offered a PR |
| Shell / Bash (& PowerShell) | #239 | Mature Bash tree-sitter grammar exists |
[...]

PR #1264, opened 2026-07-12
feat: Elixir language support (.ex/.exs)

w0lan:
Follow-up to #1220/#1221 and the Elixir entry in #648 — this adds Elixir as a supported language.

### What this adds

- **Modules as namespaces** — `defmodule`, including nested modules with their full dotted name (`Foo.Bar.Baz`), `defprotocol`/`defimpl` blocks.
- **Functions** — `def`/`defp`/`defmacro`/`defguard` with correct visibility; multiple clauses of the same function (pattern matching, guards, default-arg heads) merge into one node. `@spec` becomes the signature, `@doc`/`@moduledoc` the docstring, `defstruct` the fields.
- **Cross-module call edges** — `Module.fun(...)` resolves through the file's `alias` table (including `alias Foo.{A, B}` and `alias ..., as: X`), `__MODULE__.fun` links within the module, `&Mod.fun/arity` captures and `%Struct{}` literals become `references`, `@behaviour`/`use` become `implements`, `import`/`require`/`alias` become import edges. Calls inside module-attribute values (`@key Config.fetch!(...)`) are extracted; `@spec`/`@doc`-style metadata attributes are consumed so `String.t()` and friends never become edges.
- **Bare local calls resolve same-file only.** In Elixir a bare lowercase call can only target the enclosing module or an import. Letting bare names fall through to repo-wide matching produced systematic false edges on real code (every Mix `config :app, ...` linked to someone's `config/0`; ConnTest's imported `get(conn, path)` linked to an HTTP client's `get`). The rule removed ~3.2k guessed edges on the first validation repo with zero recall cost on same-module calls. Import-based resolution is deliberately out of scope for v1.
- **Silence over guessing** — per the project's "a wrong edge is worse than no edge" rule, dynamic dispatch with no static target (variable receiver, `apply/3`, `GenServer.call(pid, ...)`) is left unlinked.

31 extraction/resolution tests added; full suite green (2393 passed / 0 failed on Node 22). The branch also went through two independent external code reviews before submission; all nine findings (guard-expression calls, `%__MODULE__{}` / `&__MODULE__.fun/arity` references, one-line `defmodule ..., do:` bodies, dynamically-named `defmodule` bodies — with static recovery of literal `Module.concat(A, B)` names, function-local `alias`, declaration/Kernel-macro call noise) were reproduced by a failing test first and are fixed in the last commit.

### Validation on real code

Validated on **6 production Elixir repositories** (~840 Elixir files) from one company codebase, deliberately spanning eras: Elixir ~1.4 (2017) → 1.15, plain Plug, Phoenix 1.4, Phoenix 1.7 + LiveView, Ecto 1–3, umbrella apps, and Elixir+TypeScript/JS mixed repos. Method as in #1220: grep-established ground truth vs graph for statically-resolvable calls (recall), manual audit of random `calls` edges (precision), plus a framework-macro false-edge sweep.

| Metric | Result |
|---|---|
| Recall (statically-resolvable calls) | 174/174 after the module-attribute fix (173/174 before) |
| Precision (random `calls` edges, manual audit) | 60/60 |
| False edges from Phoenix/Ecto/Mix macros | 0 |
| Same-named functions across modules (e.g. one name defined in 9 modules) | all resolved to the correct module |

Results were uniform across the 1.4→1.15 spectrum — resolution rides on static module names (`alias`/fully-qualified), so old code works as well as new.

**Fair coverage** (methodology from the README table, calibrated by reproducing two published rows with the same build — requests 100.0% vs 100%, ripgrep 86.8% vs 86.7%): **`elixir-ecto/ecto` 92.9%** (52/56; the 4-file residual is all genuine frontier: the OTP application entry, compile-time conditional protocol impls, and the two query-DSL pseudo-modules `Ecto.Query.API`/`WindowAPI`). README row added in this PR. As an application-shaped stress test, the same measurement on `plausible/analytics` gives 67.4% on shipping Elixir — the residual there is structural (controllers/LiveView/workers reachable only through routing and scheduling; the router alone holds 419 unresolved refs because `get`/`post`/`live` macros don't bind their module argument), which is precisely what the Phoenix resolver below is for.

### Known limitations (v1)

- **Imported bare calls** (`import Foo` + bare `bar()`) resolve same-file or not at all — no import expansion yet.
- **Dynamic dispatch** (module in a variable, `apply/3`, GenServer/process messaging, `Application.get_env`-configured modules) — intentionally silent.
- **User macros / DSLs** are not expanded (Phoenix router, Ecto schema DSL — see roadmap below), `.heex` templates are not parsed. This includes a bare module alias standing as a declarative-macro argument (`field :secret, MyApp.EncryptedBinary`, `belongs_to :site, Site`) — no reference is emitted for it yet (deliberate: emitting every capitalized mention risks noise; scoping it to known declarative positions is part of the Phoenix/Ecto follow-up).
- **`defdelegate`** does not yet emit an edge to its `to:` target; **`@derive`** does not emit an `implements` edge (its targets are almost always out-of-repo protocols like `Jason.Encoder`).
- Bare local calls resolve to a function of the **caller's own module** only: in a file defining two+ modules, a bare call is matched against same-file *and* same-module candidates, so it can no longer bind to a same-named function in a sibling module. When the call site's module can't be determined, or the callee is defined only in another module (or file), the call stays unresolved rather than pointing at the wrong function.

### Roadmap / follow-ups

This PR is the language core — the first step of a planned path:

1. **Phoenix router resolver** (`src/resolution/frameworks/phoenix.ts`) — route → controller action / LiveView edges (`get "/x", StatsController, :funnel` with `scope`-alias nesting), in the spirit of the Laravel/Rails resolvers. **Next PR, in preparation**; it will be registered in #967. Measured headroom on plausible/analytics above; validation will run on production Phoenix apps plus a canonical OSS app + control, per the bar stated there.
2. **`.heex` templates** — component/assign references, precedent in `docs/design/template-markup-parser.md`.
3. **Behaviour/GenServer callback edges** — the dispatch-synthesizer family; after feedback on 1–2.

### Relation to the open Elixir PRs (#228, #871, #1229, #1250)

This is an independent implementation — all four open PRs' code was read before submitting, and this section is meant to make consolidation easier, not to relitigate them. In one line each: **#228** takes the same architectural approach as this PR (macro-name dispatch in a visitNode hook) with a narrower feature set; **#871/#1250** are one lineage (#1250 = #871 rebased + an `as:`-alias fix, validated by its authors on a ~3,400-file production monorepo); **#1229** pairs a thinner core with two genuinely valuable pieces no one else has — a tested Phoenix router resolver and HEEx handling.

What this PR adds over the extraction-focused ones (verified against #1250's branch):

- **Default and multi-alias resolution.** `alias MyApp.Repo` + `Repo.get(...)` and `alias MyApp.Accounts.{User, Profile}` resolve to the right module here (per-file alias table, memoized); in #1250 only the explicit `alias X, as: Y` form resolves — the import node is stored under the canonical name, so the common default-alias style yields no edge.
- **Bare-call precision gate.** #1250 has no Elixir rule in the name matcher, so bare lowercase calls fall through to repo-wide name matching — on the validation repos that produced systematic false edges (Mix `config`, imported ConnTest `get`). The same-file-only rule here removed ~3.2k guessed edges on the first validation repo.
- **Attribute handling** (`@spec` → signature, `@doc` → docstring, META vs custom attribute split so `String.t()` never becomes an edge but calls inside `@key Config.fetch!(...)` do), clause merging, `defguard`, protocol/impl blocks, captures and `%Struct{}` references.
- **Validation methodology, in the spirit of #1220:** ground truth established by grep per function (recall measured against what *should* be in the graph, not by inspecting what is), precision on random edge samples, and a framework-macro false-edge sweep — across 6 production repos + the calibrated fair-coverage measurement above. This is what caught the default-alias and bare-call classes in the first place.

@waseigo's #1229 (Phoenix resolver + HEEx) is complementary to this core and overlaps with follow-up 1 above — the router resolver there looks portable onto this core, and coordinating/crediting there beats duplicating it.

The branch is built on current `main` (no conflicts) with a clean, reviewable history — it's intended to be mergeable as-is. If you'd rather consolidate the Elixir PRs yourself (as with ArkTS/Nix), it works as the reference implementation + validation corpus too — take whatever is useful, same as #1221 — though the commits landing as they are would of course be the preferred outcome.

Scope adjustments or a different split are fine on request.


w0lan:
While validating this branch on a production Elixir repo I noticed missing `calls` edges in ExUnit test modules (one module lost 9 of 26 calls to the same helper). It is **not** an extraction issue with this PR: since `describe`/`test` are macros rather than scopes, all N calls attribute to the enclosing module node, producing N refs with an identical `(from_node, name, kind)` key differing only by line/col — and extraction emits all of them correctly.

The loss happens later, in the resolver: when such a same-key group straddles the 5000-row resolution batch boundary, the per-batch cleanup (keyed on that 3-tuple, without line/col) deletes the whole group, including rows a later batch hasn't read yet. That's exactly #1265, fixed in #1266 — with #1266 applied on top of this branch, the repo goes from 17/26 to 26/26 edges for that module, with no other changes (+9 edges total, node count identical). Elixir test suites make this pre-existing bug much more likely to fire (many identical-key calls per module), so #1266 is worth landing alongside or before this PR.

colbymchenry (commented):
## CodeGraph review

**Overall risk: 🟡 Low**

Adds full Elixir language support to the extractor, including module, function, macro handling and alias resolution. The most important risk is that the generic call extraction may not correctly handle Elixir call nodes, potentially missing call edges or emitting incorrect references.

<details>
<summary>Blast radius: 37 files affected beyond the diff · 274 symbols · 12 test files selected</summary>

Tests to run:
- `__tests__/exclude-config.test.ts`
- `__tests__/extension-mapping.test.ts`
- `__tests__/extraction.test.ts`
- `__tests__/frameworks.test.ts`
- `__tests__/grammar-wasm-bytes.test.ts`
- `__tests__/include-config.test.ts`
- `__tests__/multi-repo-workspace.test.ts`
- `__tests__/mybatis-extractor-robustness.test.ts`
- `__tests__/object-literal-methods.test.ts`
- `__tests__/pr19-improvements.test.ts`
- `__tests__/resolution.test.ts`
- `__tests__/security.test.ts`

</details>

---

w0lan:
Thanks for the review. Ran the 12 selected test files at this PR's head under Node 22: **12/12 files, 1060 passed / 2 skipped, 0 failed**.

On the flagged risk (generic call extraction mishandling Elixir call nodes): Elixir doesn't ride the generic path blindly — call nodes go through a language hook first (`ELIXIR_DECL_MACROS` / `ELIXIR_SPECIAL_FORMS` dispatch in the visitNode hook plus a dedicated extractor for `def`-family, attributes, `defstruct`, aliases), and only genuine calls fall through to `extractCall`. That area is also what the validation in the description measured directly: 174/174 recall on statically-resolvable calls, 60/60 manual precision audit, and 0 false edges from Phoenix/Ecto/Mix macros across 6 production repos, plus the 92.9% fair-coverage row on `elixir-ecto/ecto`.

w0lan:
Two updates pushed:

1. **Bare-call resolution is now qualified by the enclosing module** (new commit, TDD'd). Previously — as disclosed in Known limitations — a bare call in a file defining two+ modules with the same function name could bind to a sibling module's function, i.e. a wrong edge. A bare call now matches same-file *and* same-module candidates only, staying unresolved otherwise; single-module files behave identically. That limitation bullet no longer applies (description updated).
2. **Rebased onto current `main`** (includes #1270). Full suite on the rebased branch: 139 files, 2409 passed, 0 failed (Node 22). Revalidated on the production Elixir codebase from my earlier report: the ExUnit module that indexed 17/26 call edges now indexes **26/26** on top of your row-id fix — +9 edges total, node count unchanged, exactly the batch-boundary tail.

liz-baker:
Thank you for calling out #1250 in your writeup — I've closed it in favor of this PR. I verified the bare-call false-edge issue you described against our own production Elixir monorepo (~3,400 files): a bare `authorize(...)` call relying on an `Exop.Operation`-injected macro was resolving to an unrelated private `authorize/2` in a completely different module under #1250's build. Rebuilt and reindexed against this branch and the same call correctly falls into `unresolved_refs` instead — no false edge. Nice catch, and appreciate the much more rigorous validation methodology here.

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### New Features
```
colbymchenry, line 12: **🟠 Medium — “Changelog and release contract” Changelog entry violates release contract**

The added CHANGELOG entry includes internal macro names (defmodule, def, defp, etc.) and extractor implementation details, which are not user‑facing as required by the changelog contract.

⛔ Violates the business rule **Changelog and release contract**.

colbymchenry (commented):
## CodeGraph review

@w0lan — 1 finding · 1 business rule violated on this change.

**Overall risk: 🟠 Medium**

Adds full Elixir language support to CodeGraph, including WASM grammar loading, file‑extension mapping, and a dedicated extractor. The most important risk is that the CHANGELOG entry leaks internal implementation details, violating the release‑contract rule.

### Findings

| Severity | Finding | Where |
|---|---|---|
| 🟠 Medium | “Changelog and release contract” Changelog entry violates release contract _(rule: Changelog and release contract)_ | `CHANGELOG.md:12` — see inline comment |

### Worth double-checking

- **CHANGELOG entry** — Verify that the new changelog entry is user‑facing and does not expose internal macro names or extractor implementation details.
- **Elixir extractor registration** — Confirm that `elixirExtractor` is correctly added to `EXTRACTORS` and that module‑stack state (`resetModuleStack`, `moduleNameStack`) is correctly scoped per file.
- **Extension mapping** — Ensure that the `.ex` and `.exs` extensions map to the `elixir` language and that `isSourceFile` recognises them.

### Business rules (1 violated · 1 honored · 3 not applicable)

| Status | Rule | Note |
|---|---|---|
| ⛔ Violated | Changelog and release contract | CHANGELOG entry contains internal implementation details, breaching the user‑facing requirement. |

<details>
<summary>QA checklist — 2 things to verify in the running product</summary>

- [ ] **cli** — Run `npm test` and ensure all Elixir extraction tests in __tests__/extraction.test.ts pass. _(Before this PR Elixir was unsupported, so those tests would fail; passing confirms the new extractor works.)_
- [ ] **cli** — Run `codegraph init` on a sample project containing an .ex file and verify no parse errors are reported. _(Without the change the language would be reported as unsupported; successful init shows the new language is recognized.)_

</details>

<details>
<summary>Blast radius: 38 files affected beyond the diff · 276 symbols · 12 test files selected</summary>

Tests to run:
- `__tests__/exclude-config.test.ts`
- `__tests__/extension-mapping.test.ts`
- `__tests__/extraction.test.ts`
- `__tests__/frameworks.test.ts`
- `__tests__/grammar-wasm-bytes.test.ts`
- `__tests__/include-config.test.ts`
- `__tests__/multi-repo-workspace.test.ts`
- `__tests__/mybatis-extractor-robustness.test.ts`
- `__tests__/object-literal-methods.test.ts`
- `__tests__/pr19-improvements.test.ts`
- `__tests__/resolution.test.ts`
- `__tests__/security.test.ts`

</details>

---
[Full report](http://localhost:8365/impact/641a0053-d1af-4755-bdea-0a01a8439330)

w0lan:
Fair point — the entry described the extractor, not the user-visible behavior. Rewrote it in user-facing terms (what gets indexed and how calls resolve), dropping the macro-by-macro internals — the `docs(changelog): rewrite Elixir entry in user-facing terms` commit at the branch tip. (Commit SHAs shift as we keep the branch rebased on current main whenever the Unreleased changelog section moves, so the branch stays conflict-free and ready to merge.)

w0lan:
Some context on our side: we're rolling CodeGraph out across our company repos. A good chunk of our stack is Elixir too, which is why we put this PR together and validated it as carefully as we did.

We're just not sure how to plan around it. We're happy to wait if Elixir support is something you'd consider merging or reimplementing yourself, and we'll keep the branch rebased and green in the meantime. But if adding Elixir isn't on your roadmap, we'd genuinely rather just know — no hard feelings either way, it only helps us decide how to set things up internally.

gabrielrinaldi:
@w0lan this is great, thank you, would love to see this merged as well

tlarevo:
@w0lan thanks a lot for adding elixir support 🙏 , @colbymchenry please consider approving this 💙

humberaquino:
This is great! I hope it gets considered and approved soon 🙏 @colbymchenry

Miskamyasa:
Howdy! Only one question - When?

Issue #1240, opened 2026-07-10
Incremental sync misses cross-file relations after a new export is added


loadcosmos:
### Summary

On CodeGraph **v1.3.1**, incremental `sync` can leave the graph semantically stale when a changed file introduces an export that satisfies an existing import/call in an unchanged file.

The index reports a clean state (`pendingChanges = 0`, `pendingRefs = 0`, `state = "complete"`, `reindexRecommended = false`), but a full index of the exact same source snapshot adds missing edges.

### Minimal TypeScript reproduction

Initial `a.ts`:

```ts
import { greet } from './b';

export function run(): number {
  return greet();
}
```

Initial `b.ts`:

```ts
export function other(): number {
  return 1;
}
```

1. Run a full initial index:

```sh
codegraph init --force .
```

2. Replace `b.ts` with:

```ts
export function greet(): number {
  return 42;
}
```

3. Run incremental sync and inspect the graph:

```sh
codegraph sync --quiet .
codegraph status --json .
```

### Actual result after `sync`

- `fileCount: 2`
- `nodeCount: 5`
- `edgeCount: 4`
- edge kinds: `contains: 3`, `imports: 1`
- `pendingChanges: { added: 0, modified: 0, removed: 0 }`
- `pendingRefs: 0`

Then run a full index without changing any source file:

```sh
codegraph index --force --quiet .
codegraph status --json .
```

### Result after full index

- `fileCount: 2`
- `nodeCount: 5`
- `edgeCount: 6`
- edge kinds: `calls: 1`, `contains: 3`, `imports: 2`
- `pendingChanges: { added: 0, modified: 0, removed: 0 }`
- `pendingRefs: 0`

### Expected result

For an unchanged source snapshot, `sync` should produce the same resolved cross-file graph as a full index. In this case, the import/call in unchanged `a.ts` should be resolved after `b.ts` adds `greet`.

### Impact

The MCP file watcher uses incremental sync, so `impact`, caller/callee, and flow results can silently omit cross-file relationships after a normal edit. Status does not indicate that a full re-index is needed.

This appears consistent with reference resolution being scoped to changed files: the unresolved reference in unchanged `a.ts` is not revisited after `b.ts` gains the matching export.

### Environment

- CodeGraph: `1.3.1`
- Linux x64
- TypeScript

colbymchenry:
[...]

Issue #1187, opened 2026-07-06
Java: `codegraph_explore` blast radius misses callers for concrete-class `@Resource`/`@Autowired` field injection (package-correlated)


KnifeOfLife:
### Summary

`codegraph_explore` reports **3 callers** for a method that grep shows is called from **10 files (15 call sites)**. The method is invoked exclusively through Spring `@Resource` / `@Autowired` field injection where the **field type is a concrete `@Component` class, not an interface**. The 7 missed callers cluster in specific packages (`controller/`, `facade/impl/`, `notify/`, `support/center/`), while the 2 correctly-resolved external callers are both in `service/**/impl/` subpackages.

### Environment

- codegraph: **`v1.2.0`**
- Language: Java 1.8
- Framework: Spring Boot 2.3.0 + Dubbo 2.6.x
- Project: large monorepo (~118 MyBatis mappers, 59 Dubbo facade impls, 25 service subpackages)
- Index method: `codegraph init` + file-watcher sync

### Reproduction

**Target method** — concrete `@Component` class, no interface:

```java
@Component
public class MemberDescriptionSupport {
    public String getSuperVipName() { ... }
}
```

**Caller pattern (missed)** — every missed call site looks like this:

```java
@Component  // or @RestController / @Service
public class SomeClass {
    @Resource
    private MemberDescriptionSupport memberDescriptionSupport;  // CONCRETE class, not interface

    public void someMethod() {
        memberDescriptionSupport.getSuperVipName();  // <- edge dropped
    }
}
```

### Expected vs Actual

| Tool | Reported callers |
|---|---|
| `codegraph_explore` blast radius | 3 files |
| `grep "getSuperVipName("` (ground truth) | **10 files, 15 call sites** |

**Full breakdown:**

| Caller file | Package | Call sites | Seen by codegraph? |
|---|---|---|---|
| `MemberDescriptionSupport.java` (self-call `this.getSuperVipName()`) | `support/description` | 1 | ✅ |
| `MemberPersonalCenterServiceImpl.java` | **`service/center/impl`** | 2 | ✅ |
| `MemberOpenNoticeServiceImpl.java` | **`service/rights/impl`** | 1 | ✅ |
| `MemberAppController.java` | `controller/gw` | 2 | ❌ |
| `MemberEnhancedController.java` | `controller/gw/enhanced` | 1 | ❌ |
| `MemberRightsCSCFacadeImpl.java` | `facade/impl` | 1 | ❌ |
[...]

PR #1266, opened 2026-07-12
fix(resolution): stop silent edge loss when a same-key reference group straddles the resolution batch boundary

### LINKED ISSUE #1265: Resolver drops `calls`/`references` edges when a same-key reference group straddles the 5000-row resolution batch boundary
### Summary

On `codegraph index` (batched resolution path), edges are silently lost when a
group of references sharing `(from_node_id, reference_name, reference_kind)` —
i.e. one caller invoking the same resolvable target from many call sites — is
split across the `batchSize` (default 5000) boundary. The first batch resolves
its slice and creates its edges, then deletes the *entire* key group, including
the rows a later batch has not read yet. Those references vanish before they can
be resolved; their edges are never created. The index reports a clean, complete
state (`pending=0`), so the loss is silent.

Found while validating the Elixir support work on a production codebase (a test
module calling one helper 26 times produced only 17 `calls` edges, and the count
changed with batch size), but the bug is language-agnostic and reproduces on
clean v1.4.1 in TypeScript.

### Environment

- codegraph: **v1.4.1** (also present on current `main`, HEAD ecc8b30)
- Language-agnostic (reproduced in TypeScript; observed in the field on Elixir)
- Path: `codegraph index` → `resolveAndPersistBatched`, default `batchSize=5000`

### Root cause

`getUnresolvedReferencesBatch` (src/db/queries.ts:1873) reads pending rows with
`LIMIT/OFFSET` and **no `ORDER BY`**, so rows come back in insertion (rowid)
order; the batch loop (src/resolution/index.ts:1243-1288) always reads `offset 0`
and relies on processed rows leaving the pending set.

After each batch it calls `deleteSpecificResolvedReferences`
(src/db/queries.ts:1988):

```sql
DELETE FROM unresolved_refs
WHERE from_node_id = ? AND reference_name = ? AND reference_kind = ?
```

This deletes by the 3-tuple **without `line`/`col`**. But edge identity *includes*
line/col (`idx_edges_identity`, migration v6), so each call site is a distinct
edge. When a same-key group of N call sites straddles the batch boundary, batch #1
resolves k of them (k edges), then this DELETE removes all N rows — the unread
(N−k) tail is gone before batch #2 reads `offset 0`. `markReferencesFailed`
(src/db/queries.ts:2010) shares the same 3-tuple key. The failed-refs retry path
(`resolveAndPersistListYielding`, #1240) uses the same delete with no status
filter, so retrying a subset of a key group can also wipe that key's
still-pending rows (confirmed by a failing test while fixing).

### Minimal reproduction (deterministic, TypeScript, clean v1.4.1)

`target.ts`: `export function target(): number { return 1; }`

`caller.ts`: `run()` that calls `target()` K times, one call per line:

```ts
import { target } from './target';
export function run(): number {
  let n = 0;
  n += target();   // ... repeated K times, distinct lines
  return n;
}
```

```sh
# K = 8000 (total pending > 5000): batch boundary splits the single group
codegraph init --force .
# -> 4998 `calls` edges run->target instead of 8000; 3002 silently lost;
#    unresolved_refs: pending=0, failed=0 (deleted, not parked)

# K = 4998 (total pending = 5000, one batch): 4998 edges — no loss
```

Loss begins the instant total pending exceeds `batchSize`:

| K (call sites) | total pending | `calls` edges run→target | lost |
|---|---|---|---|
| 4998 | 5000 | 4998 | 0 |
| 4999 | 5001 | 4998 | 1 |
| 5000 | 5002 | 4998 | 2 |
| 8000 | 8002 | 4998 | 3002 |

Confirmed single-variable: on the identical 8000-call project,
`batchSize=5000` → 4998 edges, `batchSize=1000000` → 8000 edges.

<details>
<summary>Generator + edge-counter scripts (node ≥22)</summary>

`generate.mjs <outDir> <K>`:

```js
import fs from 'fs';
import path from 'path';

const outDir = process.argv[2];
const K = parseInt(process.argv[3], 10);
if (!outDir || !Number.isFinite(K)) {
  console.error('usage: node generate.mjs <outDir> <K>');
  process.exit(1);
}

fs.rmSync(outDir, { recursive: true, force: true });
fs.mkdirSync(outDir, { recursive: true });

fs.writeFileSync(
  path.join(outDir, 'target.ts'),
  'export function target(): number {\n  return 1;\n}\n'
);

const lines = [];
lines.push("import { target } from './target';");
lines.push('');
lines.push('export function run(): number {');
lines.push('  let n = 0;');
for (let i = 0; i < K; i++) {
  lines.push('  n += target();');
}
lines.push('  return n;');
lines.push('}');
fs.writeFileSync(path.join(outDir, 'caller.ts'), lines.join('\n') + '\n');

fs.writeFileSync(
  path.join(outDir, 'tsconfig.json'),
  JSON.stringify({ compilerOptions: { target: 'ES2020', module: 'commonjs', strict: true } }, null, 2) + '\n'
);
fs.writeFileSync(
  path.join(outDir, 'package.json'),
  JSON.stringify({ name: path.basename(outDir), version: '0.0.0', private: true }, null, 2) + '\n'
);
```

`count.mjs <projectDir>`:

```js
import { DatabaseSync } from 'node:sqlite';
import path from 'path';

const proj = process.argv[2];
const db = new DatabaseSync(path.join(proj, '.codegraph', 'codegraph.db'), { readOnly: true });

const nodes = db.prepare("SELECT id, name FROM nodes WHERE name IN ('run','target')").all();
const run = nodes.find((n) => n.name === 'run');
const target = nodes.find((n) => n.name === 'target');

console.log(JSON.stringify({
  run_calls_target_edges: db.prepare("SELECT COUNT(*) c FROM edges WHERE source = ? AND target = ? AND kind = 'calls'").get(run.id, target.id).c,
  unresolved_pending: db.prepare("SELECT COUNT(*) c FROM unresolved_refs WHERE status='pending'").get().c,
  unresolved_failed: db.prepare("SELECT COUNT(*) c FROM unresolved_refs WHERE status='failed'").get().c,
}, null, 2));
db.close();
```

</details>

### Suggested fix

Delete resolved rows by their full call-site identity instead of the 3-tuple —
add `line`/`col` to the `WHERE` (they already flow through `createEdges` as
`ref.original.line/column`; both columns are `NOT NULL` in `unresolved_refs`, so
plain equality works), or delete by the row `id` carried through the batch.
Either way only the rows actually resolved in the batch are removed; a straddling
group's tail stays pending and is resolved by the next batch. This preserves the
#1187 drain-termination invariant (every processed row still leaves `pending`).
Apply the same keying to `markReferencesFailed` for symmetry.

I have this fix implemented with regression tests (straddling group, exact-fit
control, retry-path clobber — each watched fail on the unfixed code first) —
PR follows shortly.

### Impact

Silent, non-deterministic-looking edge loss on large repos: any resolvable
same-key group large enough to cross a 5000-row boundary loses its tail. Rare in
ordinary code (few functions call one resolvable target dozens/hundreds of times)
but unbounded per incident (generated code, dispatch tables, repetitive test
files) — and invisible in `status`, which reports the index as complete.

### Related

#1187 (orphan sweep — its drain-termination invariant is preserved by the fix),
#1240 (failed-refs retry — shares the affected delete), #1001 (chunking of a
different delete). None covers the batch-boundary loss itself.

w0lan:
Fixes #1265.

### Summary

On `codegraph index` (the batched resolution path, default `batchSize=5000`),
edges are silently lost whenever one caller invokes the same resolvable target
from more call sites than fit in the remainder of a batch. The first batch
resolves its slice of the group and creates its edges — then deletes the
**entire** key group from `unresolved_refs`, including the rows a later batch
has not read yet. Those references vanish before resolution; their edges are
never created, and the index reports a clean, complete state (`pending=0`,
`failed=0`), so nothing signals the loss.

Found while validating the Elixir support PR on a production codebase: a test
module calling one helper 26 times produced only 17 `calls` edges on the
batched path, and the count changed with batch size. The bug itself is
language-agnostic and reproduces on clean v1.4.1 in TypeScript — full analysis,
deterministic reproduction, and loss table in #1265.

### Fix

Key `deleteSpecificResolvedReferences` and `markReferencesFailed` on the full
call-site identity — `(from_node_id, reference_name, reference_kind, line,
col)`. `line`/`col` are `NOT NULL` in `unresolved_refs`, so plain equality
suffices; they already flow through the resolver (`ref.original.line/column`).
All three call-site paths updated (`resolveAndPersist`,
`resolveAndPersistListYielding`, `resolveAndPersistBatched`). A straddling
group's tail now stays pending and is resolved by the next batch, which
preserves the drain-termination invariant from #1187 (every processed row
still leaves the pending set) and composes with the #1240 failed-refs retry.

The retry path turned out to be affected too, not just latently: the 3-tuple
delete has no status filter, so `resolveAndPersistListYielding` resolving a
subset of a same-key group also wiped that key's still-pending rows (pre-fix
it dropped 8 of 10 remaining pending rows in the new test).

No supporting index added: the existing `idx_unresolved_from_name`
`(from_node_id, reference_name)` narrows the delete to the key group and
SQLite filters line/col within it; adding a composite index would cost a
schema-migration version bump for marginal gain outside pathological group
sizes. Happy to add it here or as a follow-up if you prefer.

### Verification

- Regression test `__tests__/batch-boundary-edge-loss.test.ts`, three
  scenarios: a 15-call-site same-key group with an injected `batchSize=10`
  (8/15 edges before the fix, 15/15 after); an exact-fit control (group fills
  the batch exactly; passes pre- and post-fix); and the retry-path scenario
  above. Each bug-pinning scenario was watched fail on the unfixed code first.
- Full suite green, build/typecheck clean.
- End-to-end: the K=8000 generator project from #1265 on the rebuilt CLI
  produces 8000/8000 edges (4998 before), `pending=0, failed=0`.
- The change also went through an independent external adversarial code review
  of the final diff (verdict: approve). It independently reproduced the
  regression test's red-on-revert and confirmed via grep that no other
  3-tuple-keyed mutation of `unresolved_refs` remains; its one non-blocking
  note is the composite-index follow-up mentioned above.

### Impact

Rare but unbounded and invisible: ordinary code seldom has one function
calling one resolvable target thousands of times, but generated code, dispatch
tables, and repetitive test files do (observed in the field: 26→17 edges on a
production Elixir repo; synthetic: 3002 edges lost at K=8000) — and `status`
reports the index as complete throughout.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixes
+
+- Batched reference resolution no longer silently drops call and reference edges when one caller reaches the same target from more call sites than fit in a single resolution batch. Each call site is its own edge (identity includes line and column), but the per-batch cleanup deleted resolved refs keyed only on caller + name + kind, so the first batch's delete wiped the whole group — including the call sites a later batch hadn't read yet. Those edges never got built, while the index reported a clean, fully-resolved state, so callers and impact analysis quietly under-reported heavy call sites. The cleanup now keys on the full call-site identity, so a group that straddles a batch boundary keeps every edge. (#1265)
+
 
 ## [1.4.1] - 2026-07-10
 
--- a/__tests__/batch-boundary-edge-loss.test.ts
+++ b/__tests__/batch-boundary-edge-loss.test.ts
@@ -1,0 +1,227 @@
+/**
+ * Batch-boundary edge loss in resolveAndPersistBatched.
+ *
+ * The batched resolver reads pending refs in LIMIT/OFFSET pages (offset always
+ * 0) and, after each batch, deletes the resolved rows. The delete used to key
+ * only on (from_node_id, reference_name, reference_kind) — but edge identity
+ * ALSO includes line/col (idx_edges_identity, migration v6), so every call site
+ * is a distinct edge. When a same-key group of resolvable refs (one caller
+ * invoking the same target from many call sites) is larger than the batch size,
+ * the first batch resolves its slice, then the 3-tuple delete wipes the WHOLE
+ * group — including the tail rows a later batch hasn't read yet. Those edges are
+ * silently lost and the index still reports pending=0.
+ *
+ * This pins the fix: the per-batch delete must remove only the rows actually
+ * resolved in that batch (full call-site identity, line/col included), so a
+ * straddling group's tail stays pending and is resolved by the next batch.
+ */
+
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as os from 'os';
+import CodeGraph from '../src/index';
+import type { Edge } from '../src/types';
+
+describe('Batch-boundary edge loss (resolveAndPersistBatched)', () => {
+  let testDir: string;
+  let cg: CodeGraph;
+
+  beforeEach(() => {
+    testDir = fs.mkdtempSync(path.join(os.tmpdir(), 'codegraph-batch-boundary-'));
+  });
+
+  afterEach(() => {
+    if (cg) {
+      cg.destroy();
+    }
+    if (fs.existsSync(testDir)) {
+      fs.rmSync(testDir, { recursive: true, force: true });
+    }
+  });
+
+  function findFunction(name: string) {
+    const hit = cg
+      .searchNodes(name)
+      .find((r) => (r.node.kind === 'function' || r.node.kind === 'method') && r.node.name === name);
+    expect(hit, `expected an indexed definition of ${name}`).toBeDefined();
+    return hit!.node;
+  }
+
+  it('keeps every call-site edge when a resolvable same-key group straddles the batch boundary', async () => {
+    const srcDir = path.join(testDir, 'src');
+    fs.mkdirSync(srcDir, { recursive: true });
+
+    fs.writeFileSync(path.join(srcDir, 'target.ts'), 'export function target(): number { return 1; }\n');
+
+    // One caller invoking `target()` from CALL_COUNT distinct call sites. This
+    // is a single same-key group (from=run, name=target, kind=calls) of
+    // CALL_COUNT rows with distinct line/col — CALL_COUNT expected `calls`
+    // edges.
+    const CALL_COUNT = 15;
+    const lines: string[] = ["import { target } from './target';", 'export function run(): number {', '  let n = 0;'];
+    for (let i = 0; i < CALL_COUNT; i++) {
+      lines.push('  n += target();');
+    }
+    lines.push('  return n;', '}', '');
+    fs.writeFileSync(path.join(srcDir, 'caller.ts'), lines.join('\n'));
+
+    cg = CodeGraph.initSync(testDir);
+    await cg.indexAll();
+
+    // Re-extract the caller without resolving, so all CALL_COUNT refs are back
+    // in the pending set for a fresh batched pass.
+    fs.appendFileSync(path.join(srcDir, 'caller.ts'), '\n// requeue\n');
+    await cg.indexFiles(['src/caller.ts']);
+    expect(cg.getPendingReferenceCount()).toBeGreaterThan(0);
+
+    // batchSize 10 < CALL_COUNT 15: the group straddles the boundary.
+    const resolver = (
+      cg as unknown as {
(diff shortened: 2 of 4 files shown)


w0lan:
Field confirmation on a production Elixir codebase (the repo where this was originally noticed): an ExUnit test module calling the same helper from 26 call sites had 17 edges before this fix and exactly 26 after (total edges +9, node count unchanged, `pending=0`). ExUnit modules are a natural trigger for this bug — `describe`/`test` are macros, so every call in the file attributes to the module node and shares one `(from_node, name, kind)` key.

w0lan:
Superseded by #1270, which fixes the same bug (#1265 / #1269) by row id — the cleaner of the two directions suggested in #1265. Closing.

PR #1270, opened 2026-07-13
fix(resolution): clean up processed refs by row id so batch boundaries can't drop sibling call sites

### LINKED ISSUE #1269: Resolution batching silently drops call edges when one caller's same-named call sites straddle a batch boundary
Found while validating #1268 on nlohmann/json — an unexplained deterministic ±1 in `calls` edge counts between otherwise-equivalent index runs traced back to this.

### Root cause

`deleteSpecificResolvedReferences` (`src/db/queries.ts`) cleans up resolved refs by

```sql
DELETE FROM unresolved_refs WHERE from_node_id = ? AND reference_name = ? AND reference_kind = ?
```

— no `line`/`col`. The batched resolution loop (`resolveAndPersistBatched` in `src/resolution/index.ts`) reads pending refs in batches, resolves a batch, inserts its edges, then deletes the resolved refs by that three-column key.

When one caller contains **several call sites to the same callee** and a batch boundary splits them, resolving the first batch's sites deletes **every** row with that (caller, name, kind) key — including sibling rows in later batches that were never resolved. Their edges are silently never created. `markReferencesFailed` uses the same key shape, so the parked-as-failed path over-deletes identically.

### Concrete repro (observed)

`nlohmann/json`: `detail::binary_writer::write_cbor` calls `to_char_type` at 11 distinct lines. With the default batch size, the boundary lands between sites: the last call site's row (line 172) is deleted by its siblings' cleanup and its edge is missing — deterministically, across repeated `codegraph init` runs. Which specific edges vanish shifts with total project ref count (anything that changes ref totals moves the boundary), which is why it surfaced as ±1 noise in an A/B diff.

Effect: callers/impact silently under-report multi-call relationships. `SELECT count(*)` of edges for such pairs depends on where batch boundaries fall, not on the code.

### Fix direction

Delete by `unresolved_refs.id` (carry the row id through resolution), or include `line`/`col` in the delete/fail key. Row id seems cleanest — the table has an `INTEGER PRIMARY KEY` already, and it also dedups correctly if two refs share (caller, name, kind, line, col).

colbymchenry:
Fixes #1269

## The bug

Post-batch cleanup deleted resolved refs (and parked failed ones) by `(from_node_id, reference_name, reference_kind)` — no line/col. When one caller had several call sites to the **same callee** and a resolution batch boundary split them, the first batch's cleanup removed every row with that key — including later-batch siblings that were never attempted. Their edges were silently never created, deterministically, and *which* edges went missing shifted with unrelated changes to total ref count (which is how it surfaced as ±1 noise in #1268's A/B validation).

## The fix

Refs loaded from `unresolved_refs` now carry their row id through resolution (`UnresolvedReference.rowId` → `UnresolvedRef.rowId`), and all three persist paths — sync `resolveAndPersist`, the yielding failed-ref retry pass (#1240), and the batched drain loop — delete / mark-failed by exactly that row id (`deleteReferencesByRowIds` / `markReferencesFailedByRowIds`). The key-tuple methods remain only as the fallback for hand-built refs passed through the public API.

Failed-parking gains the same precision: resolution outcome can differ per call site (receiver-type inference reads the ref's line), so a sibling row must not inherit another row's failure.

Also untracks the `zz-scratch*.test.ts` local scratch files that accidentally slipped into #1268 and gitignores the pattern.

## Validation

- **Red-green regression test**: 5 same-named call sites, batch size 2 — old code produced 2 edges, fix produces all 5. Plus failed-path drain and legacy no-rowId fallback tests.
- **Real-codebase check (nlohmann/json)**: re-index with the fix is a **strict superset** of the previous edge set — 0 edges lost, **422 real call edges recovered** (e.g. `write_cbor`'s 38 `to_char_type` calls previously indexed as 11; recovered lines spot-checked against source).
- Full suite green: 2377 tests, `tsc --noEmit` clean.


w0lan:
FYI for cross-reference: this is the same bug reported ~4h earlier in #1265 (with a deterministic TypeScript reproduction and loss table) and fixed in #1266 via the other direction sketched there (full call-site identity in the cleanup keys). Your row-id approach is the cleaner of the two — it was listed in #1265's "Suggested fix" as the alternative, and it also avoids the composite-index question entirely. Closing #1266 as superseded and #1265 as resolved by this. One datapoint from the field that may still be useful: on a production Elixir codebase the bug dropped 9 of 26 same-key call edges in one ExUnit module (ExUnit `describe`/`test` are macros, so whole test files share one `(from_node, name, kind)` key — Elixir repos hit this unusually often).

PR #1273, opened 2026-07-13
fix(affected): honor custom filter glob semantics

Jseca666:
`codegraph affected --filter 'tests/**/*.py'` misses tests directly inside `tests/`, while `--filter 'tests/*.py'` can also select `other/tests/...` or a changed `.py.bak` file. This change uses the existing `picomatch` dependency to match glob patterns correctly.

Filters containing `/` match the whole project-relative path, and `**/` matches zero or more directories. Filename-only filters such as `*_test.go` continue to match at any depth, and explicit custom filtering still includes dotfiles.

This PR now focuses on the remaining custom-filter fix from #1273. The default test-file recognition is already covered by #1803; its shared `isTestPath` classifier and existing Go/Python/JVM regression tests are retained. The branch is synchronized with `main` and preserves the original contribution history.

### Validation

- `npm run build` passed on Windows with Node 24.14.1.
- 50 tests passed across `cli-affected-filter`, `cli-affected-paths`, `cli-affected-test-conventions`, `is-test-file`, and `graph`.
- The new CLI regression cases reproduced missing root-level matches and incorrect directory matches against the upstream implementation before the fix.
- `git diff --check` passed.
- `npm audit` reports 10 existing dependency advisories (1 critical, 4 high, 5 moderate). Dependency manifests and lockfiles match upstream; this PR adds or updates no dependencies.


maxmilian:
Drive-by note from #1507, in case it helps this get looked at: **this PR also fixes a bug nobody has connected to it.**

#1507 (opened today) reports `codegraph affected` missing `math_test.go` for a Go file. That is the same inline `defaultTestPatterns` block you delete here — Go's underscore suffix matches none of the six JS/TS patterns, and the shared `isTestFile` you import matches it on `/[._-](test|tests|spec|specs)\.[a-z0-9]+$/`. So this PR closes at least two issues, not one.

I ran both predicates against the same inputs at `upstream/main` (`c65d56c`) to be sure rather than reading the regexes:

| path | inline (today) | shared (this PR) |
|---|---|---|
| `math_test.go` | **false** | true |
| `tests/test_math_utils.py` | **false** | true |
| `FooTest.kt` | **false** | true |
| `math.test.ts` | true | true |
| `math.go` | false | false |

The JVM/Swift CamelCase shapes (`FooTest.kt`, `BarTests.swift`, `QuxTestCase.java`) are broken today too, and your change fixes those as well — the inline list has no CamelCase rule at all. `math.go` staying false on both is the check that the swap doesn't over-match ordinary sources into "tests".

The one gap: your two new tests are both Python, so nothing here would catch a Go regression. If you'd like, I can send a commit adding `math.go` / `math_test.go` to `__tests__/cli-affected-paths.test.ts` against your branch — say the word and I'll open it, or ignore this if you'd rather keep the PR as-is. Not trying to take it over; it's your fix and it's been sitting for three weeks.

Jseca666:
Thanks for the careful validation and for connecting this to #1507. Please go ahead and add math.go / math_test.go regression coverage to this branch — I agree it should be included before merge.

Jseca666:
Hi @maxmilian, thanks again for offering to add the Go regression coverage. Just checking in — are you still planning to send the follow-up commit for math.go / math_test.go? No worries if timing is tight; I’m happy to add it on my side instead. Please let me know if you need anything from me to help move this forward.

Jseca666:
Hi @colbymchenry, I noticed #1507 was closed via #1803, based on #1688.

My PR #1273 has been open since July 13 and already proposed replacing the duplicated test-file detection. On August 5, @maxmilian verified that it also fixes #1507 and linked it in both discussions.

Could you review this earlier contribution and clarify how it should be acknowledged? My PR also includes a separate `--filter` glob fix that appears to remain outstanding. I’d like to get that contribution merged—would you prefer that I update #1273 to focus on the remaining fix?

Thanks.

Issue #540, opened 2026-05-29
[Bug] codegraph sync fails with "too many SQL variables" during code parsing


JimmyChung208:
Description:
When executing the codegraph sync command, the synchronization process crashes at the end of the "Parsing code" stage with a database variable limit error.

Steps to Reproduce:

Open the terminal and navigate to the project repository.

Run the command codegraph sync.

The "Scanning files" step completes successfully.

The "Parsing code" step reaches 100%, and the command exits with an error.

Expected Behavior:
The synchronization should complete successfully regardless of the project size. If the repository is large, the tool should ideally chunk or batch the database transactions to avoid hitting SQLite limits.

Actual Behavior:
The process fails with the following output:
https://meee.com.tw/8JkYdbd

Environment:

OS: Ubuntu 24.04

CodeGraph Version: 0.9.6

Project Size: AOSP 1TB

Additional Context:
This appears to be related to the SQLite MAX_VARIABLE_NUMBER limitation. It seems the tool is attempting to insert or query too many records in a single SQL statement when parsing a large codebase. Chunking the operations into smaller batches might resolve this issue.

PR #1283, opened 2026-07-14
fix(sync): avoid argument-stack overflow gathering unresolved refs on large syncs

ianmage:
## Problem

`codegraph sync` aborts with `RangeError: Maximum call stack size exceeded` on large projects — reproduced on a large Perforce workspace (Unreal Engine C++/Python, ~6k files, ~1.1M references), and reproducible on any fresh index, large changelist, or first sync after a branch switch.

Captured stack:

```
RangeError: Maximum call stack size exceeded
    at QueryBuilder.getUnresolvedReferencesByFiles (dist/db/queries.js)
    at CodeGraph.sync (dist/index.js)   // resolution git-fast-path
```

## Root cause

Two unresolved-ref readers chunk their `IN (...)` list under SQLite's parameter limit, then collect the matched rows with `rows.push(...chunkRows)`. Chunking the file-path / name list bounds the **SQL parameters**, not the returned **row count**: one 500-item chunk can match hundreds of thousands of rows, and spreading them as call arguments exceeds V8's argument-stack limit (~125k args). Affected:

- `getUnresolvedReferencesByFiles` — pending refs for changed files (the sync git-fast-path)
- `getRetryableFailedReferences` — the failed-ref retry path (#1240)

## Fix

Append row-by-row (`for (const row of chunkRows) rows.push(row)`) in both readers. No behavior change; only removes the arg-spread. The existing file-path / name chunking (for the SQLite parameter limit, #540) is unchanged.

## Tests

Two regression tests in `__tests__/db-perf.test.ts` drive each reader past the arg-spread ceiling. Both throw `Maximum call stack size exceeded` on the pre-fix spread and pass on the fix. Full `db-perf` + `pr19-improvements` suites green (55 tests). Verified end-to-end: a forced full re-sync of the ~1.1M-ref workspace that previously threw now completes.


ianmage:
Rebased onto latest main — conflict resolved. Ready for review @colbymchenry.

ianmage:
Verified that upstream already contains this fix: #1576 (commit d8f2eea) applies the identical loop-append to both readers in `src/db/queries.ts` (`getUnresolvedReferencesByFiles` and `getRetryableFailedReferences`), plus additional hardening in `src/bin/codegraph.ts` and `src/extraction/index.ts`, and pins it with `__tests__/refs-by-files-spread.test.ts` (200k-row scenario). Both regression tests from this PR were re-run against current `main` and pass. Closing as superseded by #1576 — thanks!

Issue #1043, opened 2026-06-29
Support for C++ classes inheriting templated classes in C++


ryancu7:
CodeGraph is not working very well on our C++ codebase. We have a lot of classes that inherit from templated types. [DeepWiki (AI generated)](https://deepwiki.com/search/if-i-have-a-c-project-that-has_96e61b32-c9dc-4083-90c7-325b981e5a33?mode=deep) says this could be the reason [(more detail)](https://deepwiki.com/search/if-i-have-a-c-project-that-has_510872a0-6a79-4a66-94a8-9d01aaab3eb9?mode=codemap). Can CodeGraph support following classes that inherit from templated classes?

colbymchenry:
Fixed in #1048 (merged to `main`).

**Root cause:** when a class derived from a template — `class Derived : public Base<int>`, a CRTP base like `class App : public CRTPBase<App>`, or a `struct` inheriting a template — CodeGraph recorded the base as the full instantiation text `Base<int>`. That never matched the template, which is indexed under its bare name `Base`, so the inheritance edge was dropped and the derived class looked like it inherited from nothing (impact/callers analysis stopped at that boundary).

**Fix:** the template arguments are now stripped from the base name (`Base<int>` → `Base`, `ns::Tpl<int>` → `ns::Tpl`), so a templated base resolves exactly like a non-templated one. This covers plain template bases, CRTP, structs, and multi-base clauses, for bases in the same file or same namespace.

One case is still not linked: a base in a **different namespace referenced with its qualifier** (`class Q : public other_ns::Tpl<int>`). That's a separate, pre-existing namespace-resolution limitation — the non-templated `other_ns::Plain` doesn't resolve either — so it's not specific to templates. If that pattern is common in your codebase, open a new issue and we'll track it separately.

This will go out in the next release. Thanks for the report (and the DeepWiki diagnosis — it pointed right at it)!

PR #1287, opened 2026-07-14
fix(extraction): strip template args from out-of-line C++ method qualifiers (#1286)

### LINKED ISSUE #1286: C++: out-of-line template method `qualified_name` keeps template params (`Box<T>::get`), can overflow NAME_MAX
`stripCppTemplateArgs` (added in #1043 to drop `<…>` from templated **base-class** references so `extends` resolves) is not applied to **method qualifiers**. An out-of-line template method definition therefore keeps the class template parameter list in `qualified_name`.

### Reproduction

```cpp
template <typename T>
class Box { public: T get() const; void set(T v); private: T value; };

template <typename T> T Box<T>::get() const { return value; }
template <typename T> void Box<T>::set(T v) { value = v; }
```

After `codegraph init`, the `nodes` table has:

| `name` | `qualified_name` | expected |
|---|---|---|
| `get` | `Box<T>::get` | `Box::get` |
| `set` | `Box<T>::set` | `Box::set` |

Notes:
- `name` is already clean (`get`); only the qualifier leaks.
- **Inline** method bodies inside the class are unaffected (they come out as `Box::get`); only the **out-of-line** definition syntax leaks the params — so the same method gets a different `qualified_name` depending on where it's defined.
- The `<T>` here is the primary template's parameter *name* (always `T`), not a specialization, so it carries no disambiguating value.

### Severe manifestation — `qualified_name` can exceed `NAME_MAX` (255)

With long, multi-line template parameter names (this is the shape of ICU's `capi_helper.h`), codegraph stores the entire `Class<…params…>::method` — newlines included — as `qualified_name`:

```
qualified_name (272 bytes):
  ApiHelper<IntegerConversionSourceCTypeParameterForTheHelperTemplateClassInstanceLong,
            IntegerConversionDestinationCPlusPlusTypeParameterForTheHelperTemplateClassLong,
            kMagicValidationSentinelConstantForTheHelperTemplateClassInstanceGuardLong>::validate
```

A consumer that derives a filename from `qualified_name` then fails with `OSError: [Errno 63] File name too long` (272 > 255).

### Impact

Same failure mode #1043 fixed for `extends`: `name-matcher` compares the qualifier before the last `::`, and `Box<T>` does not match the class node indexed as `Box`, so a method resolved via its out-of-line definition can fail to link to its class.

### Root cause / fix

`extractCppReceiverType` in `src/extraction/languages/c-cpp.ts` builds the receiver by joining the `::`-parts of the declarator's qualified-id but does not strip template arguments; the receiver then becomes `qualifiedName = \`${receiverType}::${name}\`` in `tree-sitter.ts`. Applying the existing `stripCppTemplateArgs` to the receiver (it already turns `Outer<int>::Inner` → `Outer::Inner`) fixes both the normal and the long-name case. PR incoming.

Reproduced on 1.4.x.

Dshuishui:
Fixes #1286.

## Problem
`stripCppTemplateArgs` (added in #1043 to drop `<…>` from templated base-class refs so `extends` resolves) isn't applied to method qualifiers. An out-of-line template method keeps the class template params in `qualified_name`:

```cpp
template <typename T> class Box { public: T get() const; };
template <typename T> T Box<T>::get() const { /* ... */ }
// → qualified_name = "Box<T>::get"   (expected "Box::get")
```

- `name` is already clean (`get`); only the qualifier leaks.
- Inline method bodies are unaffected (`Box::get`) — the same method got a different `qualified_name` depending on where it's defined.
- Long/multi-line template params (ICU `capi_helper.h` shape) push the `qualified_name` past NAME_MAX (255) → consumers deriving a filename hit "File name too long".
- Same failure mode #1043 fixed for `extends`: `name-matcher` compares the qualifier before the last `::`, and `Box<T>` ≠ the class node `Box`.

## Fix
Apply the existing `stripCppTemplateArgs` to the receiver in `extractCppReceiverType`, before the `::` split (also tolerates `::` inside template args like `Box<std::string>::get`). Reuses the #1043 helper.

## Tests
- `__tests__/extraction.test.ts`: out-of-line `Box<T>::get`/`set` now extract as `Box::get`/`Box::set`; no node carries `<` in its qualified name.
- All template-related tests (incl. #1043) pass.

### DIFF
--- a/__tests__/extraction.test.ts
+++ b/__tests__/extraction.test.ts
@@ -4074,6 +4074,35 @@ class Both : public Base<char>, public Plain {};
     });
   });
 
+  describe('C++ out-of-line template method qualifier (#1286)', () => {
+    // An out-of-line template method definition
+    // (`template<typename T> T Box<T>::get()`) recorded its class qualifier as the
+    // full instantiation `Box<T>`, so the method's qualifiedName was `Box<T>::get`.
+    // That never name-matched the class node indexed as `Box`, and long/multi-line
+    // template params could push the qualifiedName past the 255-byte filename
+    // limit. The `<…>` args are stripped so the qualifier is the bare `Box`,
+    // exactly like #1043 did for templated base classes. Inline method bodies were
+    // already unaffected.
+    it('strips template args from an out-of-line template method qualifier', () => {
+      const code = `
+template <typename T>
+class Box { public: T get() const; void set(T v); private: T value; };
+
+template <typename T> T Box<T>::get() const { return value; }
+template <typename T> void Box<T>::set(T v) { value = v; }
+`;
+      const nodes = extractFromSource('box.cpp', code).nodes;
+      const get = nodes.find((n) => n.name === 'get');
+      const set = nodes.find((n) => n.name === 'set');
+
+      // Qualifier is the bare class, not the `Box<T>` instantiation.
+      expect(get?.qualifiedName).toBe('Box::get');
+      expect(set?.qualifiedName).toBe('Box::set');
+      // No node still carries angle brackets in its qualified name.
+      expect(nodes.find((n) => n.qualifiedName?.includes('<'))).toBeUndefined();
+    });
+  });
+
   describe('C++ stack-allocation construction (#1035)', () => {
     // `Calculator calc(0)` (direct-init) and `Widget w{1, 2}` (brace-init) carry
     // the constructor args directly on the declarator — no call/new node — so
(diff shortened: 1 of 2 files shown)


Dshuishui:
Closing — this was fixed upstream in #1309 (commit `4dd29ea`), which applies `stripCppTemplateArgs` to the method qualifier in `extractCppReceiverType`, the same fix this PR proposed. No longer needed. Thanks!

PR #1288, opened 2026-07-15
fix(resolution): skip oversized files when chasing imports/re-exports — a single large asset import OOMs the whole index

li1164267803:
# fix(resolution): skip oversized files when chasing imports/re-exports — a single large asset import OOMs the whole index

## Problem

`codegraph init` dies with `FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory` during "Resolving refs", reproducibly at the same ref, on a mid-size React/TS project (1,161 files, ~54k refs). Raising the heap doesn't help — an 8 GB heap (`NODE_OPTIONS=--max-old-space-size=8192`, honored by the vendored runtime) dies at the same ~65%. The background daemon OOMs the same way on every session, so the project can't be indexed at all. Affects 1.3.1 and 1.4.1; 0.9.9 was fine.

**Minimal repro — two files:**

```bash
mkdir -p repro/src && cd repro && git init
dd if=/dev/urandom of=src/big.mp4 bs=1m count=240
printf 'import video from "./big.mp4";\nexport function play(): string {\n  return video;\n}\n' > src/app.ts
codegraph init   # OOM at "Resolving refs"
```

## Root cause

Reference resolution follows import targets verbatim. An asset import — `import Emilia from "@/assets/Emilia.mp4"` in the real project (a 243 MB video), `./big.mp4` above — hands the resolver a path that extraction never parsed. Extraction has a `MAX_FILE_SIZE` guard (1 MB) precisely so vendored blobs and bundles stay out of the graph, but resolution's `readFileCached()` has no such guard: the import/re-export chase (`getReExports`, `getImportMappings`) reads the target **whole**, decodes hundreds of MB of binary as UTF-8, and regex-scans the resulting string for re-exports (for a non-JS extension the language re-key falls back to the consumer's language, so the mp4 bytes are scanned as TypeScript). The transient strings and match arrays blow through any heap limit. Since the file was never parsed, the read can never produce a resolution — the cost buys nothing.

Instrumentation pinpointing it (per-ref log, 2 GB heap): heap is at 110 MB entering the `Emilia` import ref, then the process dies inside that single `resolveOne` call.

## Fix

Mirror extraction's guard in `readFileCached()`: `stat` before reading, and treat anything over 1 MB as unreadable (cached `null`). Oversized files become exactly as invisible to resolution as they already are to extraction — no behavior change for any file the graph can actually contain.

## Tests

Two regression tests in `__tests__/resolution-oversized-file.test.ts`:

- an oversized (>1 MB) but syntactically valid barrel must **not** have its re-export chain chased — pre-fix this produced a `calls` edge through a file the graph doesn't contain (that's the read the OOM rides in on); post-fix the chase stops at the barrel
- a project importing a >1 MB binary asset indexes cleanly

Both fail before the fix and pass after. Full suite: 139 files, 2,374 passed / 4 skipped — green.

Verified end-to-end: the 240 MB two-file repro and the original 1,161-file project both index completely with a **2 GB** heap post-fix (previously OOM'd 8 GB); index integrity confirmed (13,633 nodes / 31,397 edges — 232 more resolved edges than 0.9.9 produced on the same repo).


colbymchenry (commented):
## CodeGraph review

**Overall risk: 🟡 Low**

Adds a MAX_RESOLUTION_FILE_SIZE constant and a guard in readFileCached to skip reading files larger than 1 MiB, mirroring the extraction guard and preventing OOM during indexing. The main risk is ensuring the guard matches the extraction limit and does not unintentionally skip legitimate files.

### Worth double-checking

- **resolution guard** — Verify that MAX_RESOLUTION_FILE_SIZE (1 MiB) matches the extraction MAX_FILE_SIZE constant and that the guard correctly skips files larger than this size while allowing normal files.
- **readFileCached behavior** — Confirm readFileCached still returns file content for files ≤ MAX_RESOLUTION_FILE_SIZE and caches null for oversized files as intended.

### Business rules (12 not applicable)

No rules violated or at risk.

<details>
<summary>QA checklist — 2 things to verify in the running product</summary>

- [ ] **cli** — Run `codegraph init` on a project that imports a >1 MiB binary asset and verify the command completes without error or OOM. _(Validates the new guard prevents OOM during indexing.)_
- [ ] **cli** — After indexing, run a query (e.g., `codegraph query getNodesByKind function`) and ensure no functions from the oversized file appear. _(Ensures the guard correctly excludes oversized files from resolution.)_

</details>

<details>
<summary>Blast radius: 80 files affected beyond the diff · 150 symbols · 78 test files selected</summary>

Tests to run:
- `__tests__/adaptive-explore-sizing.test.ts`
- `__tests__/android-res-exclusion.test.ts`
- `__tests__/batched-ref-cleanup.test.ts`
- `__tests__/c-fnptr-synthesizer.test.ts`
- `__tests__/celery-dispatch-synthesizer.test.ts`
- `__tests__/cfml-inheritance-resolution.test.ts`
- `__tests__/cfml-receiver-inference.test.ts`
- `__tests__/closure-collection-synthesizer.test.ts`
- `__tests__/concurrent-locking.test.ts`
- `__tests__/config-secret-redaction.test.ts`
- `__tests__/context-ranking.test.ts`
- `__tests__/context.test.ts`
- `__tests__/cooperative-yield.test.ts`
- `__tests__/db-reopen-on-replace.test.ts`
- `__tests__/dynamic-boundaries.test.ts`
- `__tests__/erlang-behaviour-synthesizer.test.ts`
- `__tests__/explore-blast-radius.test.ts`
- `__tests__/explore-corroboration-ranking.test.ts`
- `__tests__/explore-nl-stopword-collision.test.ts`
- `__tests__/explore-output-budget.test.ts`

</details>

---
[Full report](http://localhost:8365/impact/824fcf91-7cd3-4601-bdbe-afad7df82611)

colbymchenry (commented):
## CodeGraph review

**Overall risk: 🟡 Low**

Adds a MAX_RESOLUTION_FILE_SIZE guard and skips reading files larger than 1 MiB during reference resolution, preventing OOM crashes when indexing large binary assets or oversized barrel files. The main risk is that callers of readFileCached now receive null for such files, which could cause unexpected null handling if not accounted for.

### Worth double-checking

- **Resolution guard** — Confirm the new MAX_RESOLUTION_FILE_SIZE constant (1024 * 1024) matches the extraction MAX_FILE_SIZE threshold and that the guard logic (`fs.statSync` size check) is correctly placed before reading the file.
- **readFileCached return handling** — Verify that downstream uses of `readFileCached` (e.g., `context.readFile`, `getFileLines`) correctly handle a null return; ensure no code assumes a non‑null string after the guard.
- **Cache behavior** — Check that the fileCache stores `null` for oversized files and that this does not corrupt LRU cache semantics.

### Business rules (12 not applicable)

No rules violated or at risk.

<details>
<summary>QA checklist — 2 things to verify in the running product</summary>

- [ ] **cli** — Run `codegraph init` on a project that imports a >1 MiB binary asset (e.g., a .mp4) and verify the command finishes without crashing or OOM. _(The new guard should skip reading the large asset, so the indexing process completes where it previously would have crashed.)_
- [ ] **cli** — Run `codegraph init` on a project with a barrel file >1 MiB that re‑exports a symbol, then query the graph for symbols in that barrel file and confirm none are present. _(The guard causes the oversized barrel to be excluded from the graph, changing the observable result compared to the pre‑PR behavior.)_

</details>

<details>
<summary>Blast radius: 80 files affected beyond the diff · 150 symbols · 78 test files selected</summary>

Tests to run:
- `__tests__/adaptive-explore-sizing.test.ts`
- `__tests__/android-res-exclusion.test.ts`
- `__tests__/batched-ref-cleanup.test.ts`
- `__tests__/c-fnptr-synthesizer.test.ts`
- `__tests__/celery-dispatch-synthesizer.test.ts`
- `__tests__/cfml-inheritance-resolution.test.ts`
- `__tests__/cfml-receiver-inference.test.ts`
- `__tests__/closure-collection-synthesizer.test.ts`
- `__tests__/concurrent-locking.test.ts`
- `__tests__/config-secret-redaction.test.ts`
- `__tests__/context-ranking.test.ts`
- `__tests__/context.test.ts`
- `__tests__/cooperative-yield.test.ts`
- `__tests__/db-reopen-on-replace.test.ts`
- `__tests__/dynamic-boundaries.test.ts`
- `__tests__/erlang-behaviour-synthesizer.test.ts`
- `__tests__/explore-blast-radius.test.ts`
- `__tests__/explore-corroboration-ranking.test.ts`
- `__tests__/explore-nl-stopword-collision.test.ts`
- `__tests__/explore-output-budget.test.ts`

</details>

---
[Full report](http://localhost:8365/impact/824fcf91-7cd3-4601-bdbe-afad7df82611)

li1164267803:
Friendly ping — this has been open since mid-July, and I just re-verified that the bug is still live on `main`.

Checked against `3ed73bc` (v1.6.0, the current npm release):

- `readFileCached()` in `src/resolution/index.ts` still does a bare `fs.readFileSync(fullPath, 'utf-8')` with no size check, and there is no size guard anywhere else in `src/resolution/` — so nothing fixed this by another route in the meantime.
- Dropping this PR's regression test onto a clean `main` worktree, the re-export case still fails: a `calls` edge is resolved *through* an oversized barrel that extraction never put in the graph.

```
× does not chase re-exports through a barrel extraction skipped for size
  → expected [ { …(7) } ] to have a length of +0 but got 1
✓ indexes a project importing a >1MB binary asset without dying
```

(The second one passes on `main` only because it uses a 2 MB fixture — the real-world trigger was a ~240 MB `.mp4` import, which OOM's an 8 GB heap at "Resolving refs" ~65% and kills every `codegraph init` on the project. The threshold is what the test pins; the scale is what bites.)

The branch is still conflict-free against `main` (GitHub reports `MERGEABLE`), and the change is 22 lines in one function plus the test file. Happy to rebase, adjust the constant to import `MAX_FILE_SIZE` from extraction directly instead of re-declaring it, or split the test out — whatever makes it easier to take. Thanks!

PR #1303, opened 2026-07-16
fix(cli): honor NO_COLOR, add --no-color, disable colors when stdout is not a TTY (#1281)

### LINKED ISSUE #1281: NO_COLOR / --no-color not honored — ANSI escapes ship in piped output of all list commands
## Context

`status`, `query`, `callers`, `callees`, `impact`, and `files` embed ANSI color codes even when stdout is a pipe, and setting `NO_COLOR=1` produces byte-identical output (verified on v1.4.1, darwin-arm64: identical `wc -c` with and without `NO_COLOR=1`). `explore` output is clean markdown; `--json` variants are ANSI-free.

For agent/tooling consumers the escapes are wasted bytes and transcript noise — and the standard escape hatch (https://no-color.org) doesn't work. The current workaround is using `--json` everywhere, which for `query` is ~5× the bytes of the human format (6,707 vs 1,395 bytes for the same 10 hits).

## Ask

Honor `NO_COLOR` (and/or add `--no-color`), and ideally default to no color when stdout is not a TTY — the common convention. The human-format output is the most token-efficient shape you ship; it just needs to be clean when piped.

maxmilian:
Fixes #1281

## Problem

Human-format output (`query`, `status`, `callers`, `callees`, `impact`, `files`, …) embedded ANSI color codes unconditionally: piping to a file or another tool captured the escapes, and the standard escape hatch (https://no-color.org) was ignored — `NO_COLOR=1` produced byte-identical output. For agent/tooling consumers the escapes are wasted bytes and transcript noise, and the workaround (`--json`) is ~5× the bytes of the human format.

## Fix

All CLI coloring already flows through the `colors` palette in `src/bin/codegraph.ts` (plus a mini-palette in `src/upgrade/index.ts`), so the fix is one shared gate — `colorEnabled()` in a new `src/ui/colors.ts`:

1. `--no-color` anywhere on the command line disables color — it's stripped from argv **before** commander parses (same pre-parse intercept pattern as `-v`), so it also works after subcommands (`codegraph query foo --no-color`) where commander would otherwise reject it as an unknown option; still registered on the program so it shows in `--help`
2. a non-empty `NO_COLOR` environment variable disables color (an **empty** `NO_COLOR` does not, per the spec)
3. otherwise color renders only when stdout is a TTY — piped/redirected output is escape-free by default

With color disabled the palette collapses to empty strings, so every existing `colors.X` interpolation and `chalk.X()` helper becomes a plain-text no-op — **no call sites change**.

## Verification

- New `__tests__/cli-no-color.test.ts` (5 tests, end-to-end against the built binary, red before / green after): piped `query`/`status`/`callers` output must be escape-free; `NO_COLOR=1` honored; `--no-color` accepted after a subcommand; empty `NO_COLOR` does **not** disable color
- Full suite: **142 files / 2405 passed** (`npm run build && npm test`)
- TTY regression check: under a pseudo-TTY (`script -q /dev/null …`), `query` output still renders colors; `--help` lists the new flag
- CHANGELOG entry under `[Unreleased]` → Fixes

## Notes

- `src/ui/shimmer-worker.ts` (index progress animation) still writes ANSI cursor control when piped — that's a rendering concern distinct from command *output* and untouched here; happy to follow up separately if wanted.

https://claude.ai/code/session_01XWwajmLbgJbK6JBCY8nKfk


maxmilian:
Superseded by #1306, which landed on main and closes #1281.

#1306 takes the more thorough route — a shared `ansiColorsEnabled()` in `src/ui/color.ts` with full precedence (`--no-color` > `--color` > `NO_COLOR` > `FORCE_COLOR` > TTY > CI) applied across the CLI, upgrade, and progress paths. It covers the same three behaviors this PR addressed (NO_COLOR env, `--no-color` flag, non-TTY/piped stdout) plus the progress-animation control chars, so this PR is now redundant. Closing it — thanks @colbymchenry.

PR #395, opened 2026-05-25
feat(haskell): add Haskell support

ichxorya:
(note: this PR was done with Claude's help - like many other PR; feel free to nudge me about this since I'm just a novice Haskell dev and want to have a tool that can help me with my work)

Edit: See https://github.com/colbymchenry/codegraph/pull/1337 for the most recent continuation of the PR. Thank you for your work!

## Summary

Adds first-class **Haskell (`.hs`)** indexing to CodeGraph, backed by the
upstream [`tree-sitter-haskell`](https://github.com/tree-sitter/tree-sitter-haskell) for its WASM file.

**What gets extracted:**
- Functions — incl. parameterless top-level binds (`main = …`, `pi = 3.14`),
  multi-clause definitions, and operator defs (`(===) x y = …`, `(===) = …`)
- Type classes as `interface`; instance methods scoped to their receiver type
- ADTs / `newtype`s / **GADTs** as `enum` + constructors as `enum_member`
- Record-syntax accessors as `field` nodes
- Type synonyms as `type_alias`; dotted-module imports preserved
- Call edges from both `match`-body **and** `where`-clause bindings
- `implements` from `instance C T` and `data … deriving (C)`
- `extends` from class superclass constraints (incl. tuple `(Eq a, Show a) =>`)
- **Higher-order call synthesis** — `map area xs` emits a `calls` edge to
  `area`. Allowlist covers 39 *function-first* combinators (`map`/`fmap`/
  `filter`/`foldr`/`mapM_`/`traverse`/`zipWith`/`sortBy`/…). Data-first
  variants like `forM_`/`for_` are deliberately excluded.

## What's deferred

- Monadic-bind / operator-as-call flows (`xs >>= f`, `f <$> xs`)
- Orphan instances where the receiver type lives in another file
- `.lhs` / `.cabal` / `package.yaml` parsing
- Pattern synonyms

## Test plan

- [x] `npm test` — 843 tests pass, including **27 new Haskell tests** in
      `__tests__/extraction.test.ts` (positive + negative coverage for every
      feature above)
- [x] Real-repo re-index with call-edge gains:
      xmonad `1a875b34` +66%, shellcheck `aac0823e` +314%,
      pandoc `f1e06147` +99%, purescript `c4a35b34` +199%
- [x] Structured probe (29/29) covering false-positives (lambdas, sections,
      composed fns, data-first HOFs, pattern synonyms) and false-negatives
      (where-clauses, GADTs, do-blocks, operator defs, superclass `extends`)

PR #1337, opened 2026-07-17
feat(haskell): add module-aware extraction and resolution

michelguillaume:
## Credit and lineage

This PR continues and supersedes #395 by @ichxorya. The branch intentionally preserves the four original commits (`f441a09`, `c0ff295`, `1c52378`, and `4abb846`) with their original authorship, then merges current `main` and layers the extraction, resolution, performance, sync-safety, and regression work on top.

Thank you to @ichxorya for the grammar integration, initial extractor, HOF/`where`/`do`/operator/GADT work, and original evaluation data.

### Extraction

- Registers `.hs` and packages the pinned, MIT-licensed ABI-14 `tree-sitter-haskell` WASM with a reproducible rebuild/health-check recipe.
- Extracts modules, grouped equations/signatures, point and pattern bindings, constants/functions, lexical helpers, ADTs/newtypes/GADTs (including grouped constructors), data families/instances, record and pattern-synonym selectors, type aliases/families, classes, instances, standalone deriving, superclasses, and FFI declarations.
- Handles prefix/infix/backticked/qualified/dotted operators without corrupting operator payloads such as `^.`, `==.`, `<-.`, or `<.>`.
- Models constructor patterns separately from runtime constructor applications and removes duplicate semantic edges for nullary constructors, sections, and higher-order arguments.
- Covers `do`, qualified `mdo`, `rec`, lambda/case/view patterns, recursive `let`/`where`, list comprehensions, pattern guards, record dot/update syntax, and lexical scopes for same-named helpers.
- Models executed operands for monadic, applicative, and alternative operators in infix and prefix form (`>>=`, `=<<`, `>>`, `<*>`, `<**>`, `<|>`, `*>`, `<*`, `<$>`, `<&>`, `$>`, `<$`). Continuations remain function references rather than fabricated direct calls.
- Supports standalone/grouped/bidirectional pattern synonyms, Haddock attachment, bundled pattern exports, builder/matcher dependencies, and top-level destructuring bindings with per-name signatures.

### Resolution

- Resolves Haskell modules from indexed module headers rather than filesystem-like import paths.
- Supports qualified/unqualified imports, `ImportQualifiedPost`, aliases, explicit lists, operators, `Type(..)`, selected children, `hiding`, export lists, and restricted/aliased/wildcard re-exports.
- Tracks parent ownership for `DuplicateRecordFields` and propagates ambiguity/restrictions through re-export chains instead of choosing an arbitrary field.
- Resolves simple annotated `OverloadedRecordDot` receivers by parent type and deliberately leaves unknown/ambiguous projections unresolved.
- Resolves local/imported/self-qualified fields, constructors, class selectors, functions, and constants while keeping instance implementations and lexical helpers scoped.
- Disambiguates duplicate module headers by conventional `Module/Path.hs` suffix and importer proximity, while declining ties.
- Gives same-line homonymous lexical declarations stable distinct IDs.

### Incremental correctness and performance

- Replays stamped Haskell import edges atomically when module/import/export topology changes, with a durable recovery marker for interruptions before, during, or after invalidation.
- Persists a schema-v9 Haskell topology fingerprint per file. Comment/body-only edits re-index and reattach incoming edges without replaying every Haskell import reference; module/import/export/re-export changes retain the conservative global replay.
- Old indexes with no fingerprint perform one conservative replay and self-heal.
- Caches signature, export, lexical-binding, and import parsing state. Signature extraction scales linearly in the regression benchmark.

## Validation

- `npm run build`
- Focused merged Haskell extraction/resolution/sync matrix: 76/76 passed.
- Full suite on the final tree: **148 test files passed, 4 skipped; 2,546 tests passed, 49 skipped; 0 failed**.
- Independent extractor and resolver audit agents both reached zero remaining reproducible Haskell P1/P2 findings.
- Re-export/import fixtures validated with GHC 9.6.7.
- `npm pack --dry-run`: 775 package entries; Haskell JS, WASM, schema migration, and grammar license included.
- `git diff --check` passed.

## Deliberate limits

This remains a static syntax/scope graph, not GHC. It models monadic/applicative syntax, lexical binders, constraints, and import/export visibility, but does not run dictionary elaboration, infer concrete monad stacks, or perform compiler-grade typeclass/instance selection. Cabal component graphs, CPP/Template Haskell expansion, `.lhs`, and `.hsc` are not modeled yet; genuinely ambiguous static references are intentionally left unresolved.

The fork keeps a daily/manual upstream merge workflow that validates build, tests, and package contents before updating this branch while the upstream PR remains open.


develop7:
hey @michelguillaume thank you for picking up the flag! you may want to keep an eye on merge conflicts and have them resolved proactively.

dpwiz:
Anything blocking in particular?

michelguillaume:
Nothing technical is blocking it. It’s ready for review, and the only remaining requirement is an approving review from someone with write access. @colbymchenry, would you be able to review it when you have time?

luochen1990:
This is a huge PR!  Thanks for your job ❤️

But I really want this!

I'll deploy it to replace the master version as a realworld test since I happen to be developing several Haskell projects.

And use LLM to review it (many rounds and lots of tokens) if this project welcomes AI contribution.  Just confirm this and I'll rush.

luochen1990:
image

It works well on my side till now. Will keep monitoring it.

luochen1990:
I ran an independent review pass over this PR (at head `bf9247e`) and sent the follow-up fixes to the branch as a cross-fork PR: [TontineTrust/codegraph#4](https://github.com/TontineTrust/codegraph/pull/4) — merging it here will update this PR automatically.

Highlights (11 linear commits, ~70 new tests, +850/−164):

- **Import-list namespaces**: bare uppercase items are dual-namespace per Haskell 2010 §5.3 — removes 33 wrong edge rows on pandoc (`import Data.Text (Text)` was wiring readers to a never-imported xml-light constructor) and makes `import M (Just)` + `Just x` resolve.
- **Operator-binding signatures**: LHS no longer split on the first `=` (`(==)` was stored as `"("`).
- **Function refs in data positions**: bare names in tuple/list/record values (+662 real reference edges on pandoc, e.g. Lua marshaling member tables).
- **Tokenizer discipline for non-Haskell queries** (`app's` / `(->)` / `https` / `stToken` fragments) and **re-export walk gating** back to pre-PR depth for non-Haskell.
- Real-repo A/B on xmonad + pandoc: more correct edges, flagship flow chains identical, indexing slightly faster. Full accounting in the PR description.

All credit to @michelguillaume and @ichxorya for the underlying work — these are incremental hardening fixes on top of it. A merge commit (not squash) when this lands would keep the attribution chain (#395 → #1337 → follow-up) intact.

PR #1375, opened 2026-07-20
fix(index): stop verified daemon before recreating database

### LINKED ISSUE #1325: `codegraph index` fails with "database file is in use" when MCP server is still running
### Environment

- **OS**: Windows 11 (Git Bash)
- **Codegraph**: 1.4.1

### Steps to Reproduce

1. Install codegraph MCP into Claude Code: `codegraph install`
2. Use Claude Code normally
3. Exit Claude Code (either via `/exit` or by closing the window directly)
4. Try to rebuild the index: `codegraph index`

### Expected

The index rebuilds successfully.

### Actual

```
[ERR] Failed to index: Could not rebuild the index — the database file is in use
(EPERM, Permission denied: \\?\E:\workspace\aod\.codegraph\codegraph.db)
```

### Additional Context

After exiting Claude Code (both `/exit` and closing the window), the MCP server process remains running on my machine. `codegraph index` continues to fail until I manually end the "Node.js JavaScript Runtime" process via Task Manager. After doing so, `codegraph index` succeeds.

hbui290:
## Summary
- Stop a verified CodeGraph daemon before `codegraph index` recreates its SQLite database.
- Fail closed when daemon identity cannot be verified, so an unrelated PID is never signaled.
- Keep unverified daemon state intact and tell the user to stop it manually.

## Root cause
On Windows, a surviving MCP daemon can retain the database handle after the client exits. Recreating the database then fails with `EPERM: database file is in use`.

## Validation
- `npm run build`
- `npx vitest run __tests__/daemon-registry.test.ts __tests__/index-command.test.ts __tests__/query-pool.test.ts __tests__/mcp-daemon.test.ts` (35 passed)

## Scope
Automated tests ran on macOS. The fix uses Node IPC/PID behavior intended to be cross-platform, but Windows runtime verification is still welcome.

Fixes #1325

### DIFF
--- a/__tests__/daemon-registry.test.ts
+++ b/__tests__/daemon-registry.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expect, beforeEach, afterEach } from 'vitest';
-import { spawn } from 'child_process';
+import { execFileSync, spawn } from 'child_process';
 import * as fs from 'fs';
+import * as net from 'net';
 import * as os from 'os';
 import * as path from 'path';
 import {
@@ -9,8 +10,10 @@ import {
   registerDaemon,
   deregisterDaemon,
   listDaemons,
+  stopDaemonAt,
   type DaemonRecord,
 } from '../src/mcp/daemon-registry';
+import { encodeLockInfo, getDaemonPidPath, getDaemonSocketPath } from '../src/mcp/daemon-paths';
 
 /** A pid that's guaranteed dead: spawn a trivial process, let it exit, reap it. */
 async function deadPid(): Promise<number> {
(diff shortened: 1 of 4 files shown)


bompus:
Verified on Windows 11 against current `main` (b9ca4b7).

This PR's branch predates the #1553 daemon-identity rewrite on main, so it no longer merges (unrelated histories) and `git apply --3way` conflicts in `daemon-registry.ts` and its test. I verified the intent by applying only the `codegraph.ts` hunk on top of main's `stopDaemonAt`, with the guard adapted to main's outcomes (`term`/`kill`/`not-running`/`no-daemon`; main no longer reports `unverified`/`still-running`).

Behaviour with a real daemon spawned the way the launcher does (`serve --mcp` with `CODEGRAPH_DAEMON_INTERNAL=1`), then `codegraph index`:

| | main | + hook |
| --- | --- | --- |
| `index` exit | 1 (`EPERM ... codegraph.db in use`, the #1325 symptom) | 0, `Stopped the CodeGraph daemon (pid …) before re-indexing` |
| daemon alive after | yes, lock file left | no, lock file removed |

The PR's own `refuses to recreate while a live daemon cannot be verified` test fails against main by design: main's `stopDaemonAt` treats an unverifiable pid as `not-running`, cleans the artifacts and lets `index` proceed, which seems the better outcome now that identity is proven by the hello. Suggest rebasing this PR down to the `codegraph.ts` hook plus a test that a real daemon is stopped; happy to push that as a fresh PR if the author prefers.

PR #1405, opened 2026-07-22
Add QML

hospitaler17:
- Implemented a custom QML extractor to parse Qt QML files, extracting components, properties, signals, and functions.
- Integrated the QML extractor into the existing extraction framework.
- Added a Qt framework resolver to handle C++ and QML specific patterns, including Q_PROPERTY and signal/slot connections.
- Updated the types to include 'qml' as a supported language.
- Enhanced the resolution process to link QML signal handlers to corresponding C++ signals.

### DIFF
--- a/__tests__/qml.test.ts
+++ b/__tests__/qml.test.ts
@@ -1,0 +1,557 @@
+/**
+ * QML Extractor Tests
+ *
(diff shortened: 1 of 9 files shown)


hospitaler17:
I'm sorry, my mistake

Issue #1110, opened 2026-07-01
Ruby: receiver.method calls drop the method name, so instance method calls produce no edge


colbymchenry:
### Summary

The Ruby extractor drops the method name from a `receiver.method` call: `lg.log()` was recorded as a call to `lg` (the bare receiver), not `lg.log`. Since `lg` matches no symbol, the reference resolved to nothing and **no method-call edge was ever produced** — so a Ruby method invoked through a receiver had no recorded callers and was invisible to impact/blast-radius and `codegraph_explore` flow traces.

This is the Ruby-specific blocker noted while adding local-variable receiver-type inference (#1108): that fix couldn't help Ruby because the call reference itself was missing.

### Root cause

`extractCall` (`src/extraction/tree-sitter.ts`) recognizes receiver-bearing calls by the `object`/`name`/`function` fields used by Java/JS/Python/etc. tree-sitter grammars. tree-sitter-ruby's `call` node instead uses `receiver` + `method` fields, so it fell through to the generic fallback, which takes the first named child (the receiver) as the callee. The method name was discarded:

```
lg.log()   ->  calls "lg"       (should be "lg.log")
Logger.new ->  calls "Logger"   (relied on calls->instantiates promotion)
```

### Fix

Handle Ruby's `call`/`method_call` nodes explicitly in `extractCall`: build `receiver.method` (so the resolver and local-variable type inference can link it), keep bare `foo(...)` as just the method name, and emit `Foo.new` as an `instantiates` reference to the class. A capitalized (constant) receiver also gets a `references` edge so a class used only via its class methods still records a dependent.

With this plus #1108, `lg = Logger.new; lg.log()` resolves `lg.log` to `Logger#log`, and the two-file same-name case is same-file-correct (#1079).

Issue #388, opened 2026-05-25
Go: `codegraph_callers` severely under-reports cross-package qualified calls (`pkg.Func(...)`) — <1% recall on a large Go monorepo


wnzzer:
## Summary

On a large Go monorepo (~4,260 `.go` files, 93k nodes, 256k edges, single language), `codegraph_callers` returns **0–2 callers for cross-package functions that grep finds hundreds to thousands of call sites for**. In-package / unqualified call resolution works correctly.

This affects the central value proposition of the tool for Go projects — `callers` / `impact` / `callees` / `trace` all build on the same call edges. For a layered Go codebase (`handler/` → `service/` → `domain/` → `dao/` packages), where the vast majority of calls cross package boundaries, these tools effectively return empty results.

## Environment

- `@colbymchenry/codegraph@0.9.4` (npm-shim, darwin-arm64)
- macOS 24.6.0
- Project: Go monolith, 4,260 `.go` files (+4 yaml), single Go module
- Index status:
  - Files indexed: 4,264
  - Nodes: 93,703
  - Edges: 255,667
  - DB size: 177.47 MB
  - Journal mode: WAL
- Index is fresh (just ran `codegraph init -i`), no concurrent writes during query

## Observed Behavior — Recall Numbers

All symbol names below are anonymized; `pkgA` is a high-fan-in utility package (logging/error helpers), `pkgB` is a configuration/rights package. The shapes are real, only names are scrubbed.

| Symbol | Call form in source | Actual call sites (`grep -rn`) | `codegraph_callers` returned | Recall |
|---|---|---|---|---|
| `pkgA.FuncX` (e.g. error helper) | `pkgA.FuncX(ctx, ...)` (qualified, cross-package) | **5,303** | **1** | ~0.02% |
| `pkgA.FuncY` (e.g. info helper) | `pkgA.FuncY(ctx, ...)` (qualified, cross-package) | 603+ files | 2 | <0.5% |
| `pkgB.FuncM` | qualified cross-package | 2 | 0 | 0% |
| `pkgB.FuncN` | qualified cross-package | 4 | 0 | 0% |
| `ok` (in-package helper, multiple defs) | bare call, same file | n/a (aggregated query) | 3 returned + 10-symbol aggregation note ✅ | works |

Querying with or without the package prefix in the `symbol` argument both fail equivalently:

- `codegraph_callers symbol="FuncY"` → `No callers found`
[...]

PR #1453, opened 2026-07-25
fix(resolution): resolve Python module members through aliased imports

br41nlet:
## Summary

Python calls and imports through an aliased module (`from pkg import module as alias`) produce no `calls` edge and no file→file `imports` edge. `codegraph callers` on such a target returns nothing.

## Problem

`resolvePythonModuleMember` and `resolveModuleImportToFile` name the target submodule by joining the import source with `imp.localName`. Under `from pkg import module as alias`, `localName` is the alias, so the lookup asks for `pkg.alias` — which names no module. The member never resolves and the edge is dropped.

`localName` and `exportedName` coincide only when the import is unaliased. That is why the unaliased form resolves and the aliased one silently does not, and why this reproduces in real trees but not in minimal repros — small repros rarely alias.

Reported in #899 as still failing after #578 / #715; this appears to be the remaining cause.

## Reproduction

```
pkg/__init__.py      (empty)
pkg/module.py        def func(): return 1
main.py              from pkg import module as mod_alias

                     def caller():
                         return mod_alias.func()
```

`caller` has no outgoing `calls` edge, and `codegraph callers func` reports none. Dropping `as mod_alias` makes both appear.

## Fix

Both sites use `exportedName` (the real submodule), falling back to `localName` when it is absent or `*`. The `import x as y` namespace branch is untouched — it binds the module at `source` itself and returns before this path.

## Tests

Added `resolves Python module-attribute calls after \`from pkg import module as alias\` (#899)` in `__tests__/resolution.test.ts`, beside the existing #578 unaliased case. It asserts the call resolves to the submodule function, and that an aliased stdlib module (`import os as operating_system`) still creates no edge.

The test earns its place: with the fix reverted it is the only failure in that file (1 failed / 176 passed); with the fix, 177/177.

## Validation

- `npx vitest run __tests__/resolution.test.ts` — 177/177
- `npx vitest run __tests__/arkts-resolution.test.ts __tests__/cfml-inheritance-resolution.test.ts __tests__/php-property-receiver-resolution.test.ts __tests__/same-name-disambiguation.test.ts` — 38/38
- `npm run build` — clean
- `npm test` — the failing-test set is byte-identical to unmodified `main` (55 failures, all in CLI / MCP-daemon / version-affordance suites, pre-existing on this machine); passing count goes 2468 → 2469, which is the added test. No new failures, none removed.

Measured on a 991-file Python project that imports modules under aliases throughout: a target with 6 call sites across 5 enclosing functions went from 0 callers to 5; another went 2 → 4; +824 edges index-wide. Unaliased and class-based resolution unchanged.

Scope is Python only — both sites are language-gated. The Go (#388) and Ruby (#1110) caller gaps have different root causes.

### DIFF
--- a/__tests__/resolution.test.ts
+++ b/__tests__/resolution.test.ts
@@ -1243,6 +1243,57 @@ def external_caller():
       expect(externalCalls).toHaveLength(0);
     });
 
+    it('resolves Python module-attribute calls after `from pkg import module as alias` (#899)', async () => {
+      // #578 fixed the unaliased form; the aliased one still dropped its edge.
+      // resolvePythonModuleMember joined `imp.source` with `imp.localName` to
+      // name the submodule, but under `from pkg import module as alias`
+      // `localName` is the alias — so it looked for `pkg.alias`, which names no
+      // module, and the member never resolved. The two names coincide only when
+      // the import is unaliased, which is why the unaliased case passed on a
+      // scratch project while real trees (which alias) kept reporting no callers.
+      fs.mkdirSync(path.join(tempDir, 'pkg'));
+      fs.writeFileSync(path.join(tempDir, 'pkg', '__init__.py'), '');
+      fs.writeFileSync(
+        path.join(tempDir, 'pkg', 'module.py'),
+        'def func():\n    return 1\n'
+      );
+      fs.writeFileSync(
+        path.join(tempDir, 'main.py'),
+        `from pkg import module as mod_alias
+import os as operating_system
+
+
+def caller():
+    return mod_alias.func()
+
+
+def external_caller():
+    return operating_system.getcwd()
+`
+      );
+
+      cg = await CodeGraph.init(tempDir, { index: true });
+
+      const caller = cg.getNodesByKind('function').filter((n) => n.name === 'caller')[0];
+      expect(caller).toBeDefined();
+      const calls = cg.getOutgoingEdges(caller!.id).filter((e) => e.kind === 'calls');
+      expect(calls).toHaveLength(1);
+      const target = cg.getNode(calls[0]!.target);
+      expect(target?.name).toBe('func');
+      expect(target?.filePath.replace(/\\/g, '/')).toBe('pkg/module.py');
+
+      // An aliased *stdlib* module must still produce no edge — resolution only
+      // matches real in-repo module files, alias or not.
+      const externalCaller = cg
+        .getNodesByKind('function')
+        .filter((n) => n.name === 'external_caller')[0];
+      expect(externalCaller).toBeDefined();
+      const externalCalls = cg
+        .getOutgoingEdges(externalCaller!.id)
+        .filter((e) => e.kind === 'calls');
+      expect(externalCalls).toHaveLength(0);
+    });
+
     it('attaches Go methods to their receiver type across files (#583, cross-file half)', async () => {
       // In Go a type's methods are commonly declared in a different file from the
       // `type` declaration (`type Box` in box.go, `func (b *Box) Get()` in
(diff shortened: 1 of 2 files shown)


inth3shadows:
Independent confirmation on a real tree, plus a **minimal repro that does fire** — the description says this "reproduces in real trees but not in minimal repros", and I think that is only true at the `calls` level. The file→file `imports` half reproduces in five files.

## Minimal repro (1.5.0, tip `d6d1728`)

```bash
#!/usr/bin/env bash
set -eu
T="$(mktemp -d)/aliasrepro"; mkdir -p "$T/pkg"; cd "$T"
: > pkg/__init__.py
printf 'def connect():\n    return "db"\n'                                              > pkg/db.py
printf 'def append():\n    return "ledger"\n'                                           > pkg/ledger.py
printf 'from . import db as dbmod\n\n\ndef use_aliased():\n    return dbmod.connect()\n' > pkg/aliased.py
printf 'from . import ledger\n\n\ndef use_plain():\n    return ledger.append()\n'        > pkg/plain.py
codegraph init . >/dev/null 2>&1
python3 - <<'PY'
import sqlite3
c = sqlite3.connect('.codegraph/codegraph.db')
q = """SELECT COUNT(*) FROM edges e
       JOIN nodes s ON e.source=s.id JOIN nodes t ON e.target=t.id
       WHERE e.kind='imports' AND s.file_path=? AND t.file_path=? AND t.kind='file'"""
print("plain.py   --imports--> ledger.py  (from . import ledger)      :",
      c.execute(q, ('pkg/plain.py', 'pkg/ledger.py')).fetchone()[0])
print("aliased.py --imports--> db.py      (from . import db as dbmod) :",
      c.execute(q, ('pkg/aliased.py', 'pkg/db.py')).fetchone()[0])
PY
```

Output:

```
plain.py   --imports--> ledger.py  (from . import ledger)      : 1
aliased.py --imports--> db.py      (from . import db as dbmod) : 0
```

The two files are identical apart from the alias. Dumping all 9 edges shows `file:pkg/plain.py -[imports]-> file:pkg/ledger.py` present and no corresponding `pkg/aliased.py -> pkg/db.py` edge at all — matching the description's "no file→file `imports` edge" exactly. (Neither arm produces a function-level `calls` edge at this size, which is probably why minimal repros looked clean: if you measure `callers connect` you see 0 for both and conclude nothing reproduces.)

## Real-tree magnitude

A ~1.5k-line stdlib-only Python package, all five modules imported the same way (`from . import X [as Y]`) and called the same way (`X.attr()`). Cross-file inbound edges against call sites counted from source:

| module | bound as | `alias.attr()` call sites | cross-file inbound edges |
|---|---|---|---|
| `registry.py` | `reg` | **38** | **0** |
| `db.py` | `dbmod` | 18 | **0** |
| `select.py` | `sel` | 6 | 1 |
| `ledger.py` | `ledger` | 10 | **44** |
| `integrity.py` | `integrity` | 3 | **15** |

38 call sites resolving to zero edges, against 10 resolving to 44. The alias is the only difference between the rows.

## Why it was expensive downstream

I build a static "which user journeys could this change break" map on top of the graph. Because `db.py` and `registry.py` are the two aliased modules, seeding **every node in either file** produced a closure that never left the file — so a change rewriting the entire graph-traversal layer answered *"no journeys affected"*, confidently and with no degrade signal. The two modules with the most call sites in the package were the two the graph could not see into.

Worth noting for whoever reviews: `select.py` keeps **one** edge despite being aliased, so the effect is not strictly all-or-nothing — a regression test asserting exactly zero might be more brittle than asserting the unaliased/aliased differential.

Happy to run this against a candidate build if that is useful.

inth3shadows:
Heads-up: this looks resolved on `main` now. #1785 (`7be699c`, "resolve Python aliased module imports (#1626)") landed on 2026-09-08.

To check, I ran this PR's own test (`resolves Python module-attribute calls after \`from pkg import module as alias\` (#899)`) as a standalone test against current `main` (`ba3c21e`). It passes: `mod_alias.func()` resolves to `pkg/module.py:func`, and the aliased stdlib call still produces no edge.

This one can probably be closed unless there's a case it covers that #1785 doesn't.

PR #1472, opened 2026-07-28
Fix Windows watcher false edits from NTFS atime

### LINKED ISSUE #1451: File reads flagged as edits on Windows (NTFS atime)
**Env:** codegraph 1.4.1 (win32-x64), same code at `main` (v1.5.0). Windows 11, NTFS with last-access updates enabled (`fsutil behavior query disablelastaccess` → `2, ENABLED`; the Windows default since 10 1803).

## Symptom

Nearly every `codegraph_explore` response flags the files it just rendered:

```
⚠️ Some files referenced below were edited since the last index sync ...
  - src/.../ServiceAccountPermissionAssignment.cs (edited 13ms ago, pending sync)
```

The files were never edited: `LastWriteTime` weeks old, git clean. Only their `LastAccessTime` changed, to the moment the daemon read them to serve the query.

## Repro

With the daemon watching, read indexed files from any other process without modifying them:

```bash
for f in $(find src -name '*.cs' | head -40); do cat "$f" > /dev/null; sleep 0.3; done
```

Run any explore while the loop runs. The merely-read files appear as pending, with ages matching the read cadence:

```
(Note: 12 file(s) elsewhere in this project are pending index sync ...
  - src/.../AddCustomerSettingsTests.cs (edited 4706ms ago)
  - src/.../AssignResourceToCustomerTests.cs (edited 4274ms ago)
  - src/.../AssignServiceRegionToCustomerTests.cs (edited 3799ms ago)
  ...)
```

## Root cause

`handleChange` ([src/sync/watcher.ts#L573-L592](https://github.com/colbymchenry/codegraph/blob/main/src/sync/watcher.ts#L573-L592)) marks every watcher event as a pending edit, with no mtime/size/hash check. On Windows, libuv's `ReadDirectoryChangesW` filter includes `FILE_NOTIFY_CHANGE_LAST_ACCESS`, so with atime updates enabled a mere **read** emits a `'change'` event identical to an edit. The daemon's own render-time reads then flag the files of the very response being built. (NTFS re-updates atime only when >1h stale, which is why each query flags a fresh batch. macOS FSEvents and Linux inotify don't report reads, so only Windows is affected.)

The debounced sync that follows is a harmless no-op (the size+mtime pre-filter rejects it); the damage is the chronic false "read these files directly" advice, which wastes agent calls and drowns out real pending edits.

JJordan0C:
## Summary

- ignore Windows watcher events when the file's current size and modification time still match its indexed metadata
- preserve deletion handling and fail open when filesystem or database metadata cannot be checked
- add unit coverage plus a Windows-only regression test for NTFS access-time updates
- document the fix in the changelog

Fixes #1451

## Validation

- `npm run build`
- `npx vitest run __tests__/watcher.test.ts __tests__/mcp-staleness-banner.test.ts` — 40 tests passed on Windows
- real Windows watcher check with `DisableLastAccess = 2`: an access-time-only update produced no pending edit, while a real write remained detectable
- full Node 22 Linux validation in a clean Docker container: `npm ci && npm run build && npm test`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixes
+
+- On Windows systems with NTFS last-access updates enabled, reading indexed source files no longer makes CodeGraph report them as edited or trigger needless auto-syncs. (#1451)
 
 ## [1.5.0] - 2026-07-21
 
--- a/__tests__/watcher.test.ts
+++ b/__tests__/watcher.test.ts
@@ -59,8 +59,16 @@ describe('FileWatcher', () => {
 
   // Inert by default — unit tests drive events via __emitWatchEventForTests
   // and never depend on real OS watch delivery.
-  const newWatcher = (syncFn: SyncFn, opts: WatchOptions = {}) =>
-    new FileWatcher(testDir, syncFn, { inertForTests: true, ...opts });
+  const newWatcher = (
+    syncFn: SyncFn,
+    opts: WatchOptions = {},
+    isFileStateCurrent?: (relativePath: string) => boolean
+  ) => new FileWatcher(
+    testDir,
+    syncFn,
+    { inertForTests: true, ...opts },
+    isFileStateCurrent
+  );
 
   beforeEach(() => {
     testDir = fs.mkdtempSync(path.join(os.tmpdir(), 'codegraph-watcher-'));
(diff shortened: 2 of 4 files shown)


bompus:
Verified on Windows 11 (NTFS) against current `main` (b9ca4b7) plus our fork's markdown/literal extras.

Merges onto main with a CHANGELOG-only conflict. No index change. `watcher.test.ts`: 37 pass, 1 fail on the first run (`should auto-sync when files change while watching`, `EPERM` from `rmSync` in the suite's `afterEach` on the temp dir, a Windows handle-still-open race), then 38/38 on two reruns, and 38/38 on main. So a flake, not a regression, but it appeared only on the PR branch in 3 runs, so the new watcher code path may hold the directory handle slightly longer on Windows.

I could not reproduce the original issue's stale-banner symptom on this volume either before or after the PR (see my note on #1451), so this is a build/test verification only, not a behavioural confirmation.

bompus:
Correction to the note above: the `rmSync` EPERM in `watcher.test.ts` also showed up on a later full-suite run of a branch WITHOUT this PR (main plus three unrelated PRs), and passed on rerun there too. So it is a host-side Windows teardown flake, not something this PR's watcher change introduced. The rest of the comment stands.

bompus:
Validated on September 10, 2026, and resolved the integration conflicts against current upstream `3ed73bc127323e63153bf6ec8354afa82ce36aaf`.

The resolved branch is [bompus/codegraph:fix/windows-atime-integration](https://github.com/bompus/codegraph/tree/fix/windows-atime-integration), at **`e4b6a00049e454d502c231aaf5cb18f4c9d770b9`**. It preserves the original `2977f063` commit. Only `CHANGELOG.md` conflicted; source and tests merged cleanly. The branch contains this fix and upstream history, without our consolidated fork changes.

Tests ran in a Windows-local checkout on **Windows 11 build 26200, Node 24.16.0, NTFS last-access updates enabled (`DisableLastAccess = 2`)**:

| Check | Original merge-base `572d22b` | PR head `2977f06` | Resolved upstream `e4b6a00` |
| --- | --- | --- | --- |
| TypeScript build and copied assets | Passed | Passed | Passed |
| Watcher + MCP staleness tests | 37 passed | 40 passed | 46 passed |
| Read-only trials producing a false banner | 3/3 | 0/3 | 0/3 |
| Read-only responses producing a false banner | 12/12 | 0/12 | 0/12 |
| Real-edit responses retaining the banner | 4/4 | 4/4 | 4/4 |

The reproduction uses actual MCP daemons and a separate process doing only `readFileSync`. Read-only size, SHA-256, and exact indexed modification time stayed unchanged. A real append remained detectable.

Daemon startup can refresh atime and conceal the bug. The fixture ages only access time after startup, lets that setter event debounce for three seconds, verifies a clean unrelated explore response and unchanged indexed mtime, then starts the read-only process. No timestamp setter runs during the measured read action.

Timing limitation: one resolved-upstream trial first responded at 384 ms, outside the 300 ms quick-check window; the other two first responded within 100 ms. A subsequent integration into our fork (`cd3e482`) reproduced zero false banners with all three first responses at 75–110 ms and retained all four real-edit banners. Its targeted suite passed 47 tests. Those fork results supplement, rather than replace, the exact upstream-branch validation.

All targeted runs exited successfully. Existing watcher teardown warnings also appeared on the original base. We did not run the full suite or viewer build on the resolved upstream branch. The original base/head package version is 1.5.0; the current upstream integration is 1.6.0. Separately, the installed 1.6.0 bundle reproduced the original false-warning behavior before this fix.

The PR is still open and conflicting at `2977f063`; we cannot push to its author's repository. To adopt the resolution, the author can run from the PR branch:

```sh
git switch fix/windows-ntfs-atime-events
git fetch https://github.com/bompus/codegraph.git refs/heads/fix/windows-atime-integration
git merge --ff-only e4b6a00049e454d502c231aaf5cb18f4c9d770b9
```

Then push that branch to the repository backing this PR. No force push is needed for the current PR head.

Issue #1482, opened 2026-07-30
TypeScript: `export const alias = fn` breaks caller/impact traversal — callers/impact of the implementation omit every consumer calling through the alias (2-file repro)


valkyriweb:
## Summary

When a function is re-exported under a value alias (`export const alias = fn`), CodeGraph indexes the call edge to the *alias* and the binding `alias = fn`, but never composes them. `callers` and `impact` on the implementation therefore omit every consumer that calls it through the alias — and report a complete-looking answer with a precise count and no uncertainty marker.

Verified on a 2-file, 6-node repro (below) and reproduced in a 16k-file TypeScript monorepo, where `impact` on a real function omitted both of its production call sites and listed only the definition file and its tests.

## Minimal reproduction

Two files, fresh repo, nothing else:

`impl.ts`
```ts
export function realImpl(): string {
  return "real";
}

export const aliasName = realImpl;
```

`consumer.ts`
```ts
import { aliasName } from "./impl.js";

export function consumerFn(): string {
  return aliasName();
}
```

```bash
git init . && git add -A && git commit -m repro
codegraph init .          # 6 nodes, 8 edges
```

## Actual

```
$ codegraph callers realImpl

Callers of "realImpl" (1):

file        impl.ts
  impl.ts:1

$ codegraph impact realImpl

Impact of changing "realImpl" — 2 affected symbols:

impl.ts
  function    realImpl:1
  file        impl.ts:1
```

`consumer.ts` does not appear in either. Changing `realImpl` breaks `consumerFn`, and impact analysis reports the blast radius as the definition file alone.

The control query shows the missing half of the join is present in the graph:

```
$ codegraph callers aliasName

Callers of "aliasName" (2):

function    consumerFn
  consumer.ts:3

file        consumer.ts
  consumer.ts:1
```

So the index contains `consumerFn → aliasName` and `aliasName = realImpl`. Only the composition is missing.

## Expected

`callers realImpl` and `impact realImpl` should include `consumerFn` / `consumer.ts`, by resolving value aliases whose initializer is a direct reference to an indexed function. Failing that, the output should mark the result as potentially incomplete rather than presenting a count.

## Why this is more than a missing edge

The failure is a false negative shaped like a complete answer. There is no "exact match not found" or "possible missing edges" signal, so a consumer cannot distinguish "nothing depends on this" from "the traversal stopped early".

[...]

PR #1484, opened 2026-07-30
fix(resolution): resolve `.js` import specifiers to their TS source

valkyriweb:
Fixes the first of the two root causes behind #1482.

## The bug

`EXTENSION_RESOLUTION` only ever **appends** a suffix. So a NodeNext-style relative specifier is tried as:

```
./x.js  ->  x.js.ts   x.js.tsx   x.js.d.ts   ...   x.js
```

`x.ts` is never tried. `resolveRelativeImport` returns `null`.

That matters more than it looks, because `"moduleResolution": "NodeNext"` **requires** the emitted extension on relative imports. In such a project *no relative import resolves at all* — every reference falls through to Strategy 3, the name-matcher. Name-matching cannot cross a rename, so a default import or an aliased import silently reports **0 callers** with full confidence.

One 16k-file TypeScript project I measured has **32,732 `.js` specifiers and 0 extensionless ones**. For that repo, import-based resolution was effectively off.

## The fix

A small rewrite table, tried **only after the literal path**:

| specifier | candidates |
|---|---|
| `.js` | `.ts`, `.tsx`, `.d.ts` |
| `.jsx` | `.tsx`, `.jsx`, `.d.ts` |
| `.mjs` | `.mts`, `.d.mts` |
| `.cjs` | `.cts`, `.d.cts` |

Applied in `resolveRelativeImport` and in `resolveAliasedImport`'s `tryWithExt`, gated to TS-family languages.

Ordering is the safety property: a real emitted `.js` sitting on disk beside its source still wins, so **no currently-resolving import changes target**. There is a test for exactly that.

## Tests

`__tests__/js-specifier-ts-resolution.test.ts`, 3 cases:

1. default import through `./impl.js` — resolves to `impl.ts`
2. `.mjs` / `.cjs` -> `.mts` / `.cts`
3. `./emitted.js` where **both** `emitted.js` and `emitted.ts` exist — the real `.js` wins

Each fixture imports under a *different local name* than the declaration, so the name-matcher cannot rescue the edge; the assertion isolates import resolution.

On `main` cases 1 and 2 fail and case 3 passes; with the fix all three pass.

## Verification

- full suite on this branch: **2,533 passed, 0 failed, 175 skipped**
- the 175 skips are the `kernel-*-parity` files, expected without a native kernel build (`CODEGRAPH_KERNEL=0`)
- real-world: a 18.4k-file TS/polyglot repo re-indexed in 1m30s, edges **1,262,985 -> 1,268,309 (+5,324)**, and a `callers` query that previously returned nothing now returns both true call sites

## Not covered

I did not build the Rust kernel, so **the kernel path for this change is unvalidated**. Both changed functions live in the TS resolution layer above it, but you'll want to confirm that.

---

Companion PR for the second root cause (alias bindings) is independent of this one and can land in either order; both touch `import-resolver.ts` in different regions, so whichever lands second may need a trivial rebase.

### DIFF
--- a/__tests__/js-specifier-ts-resolution.test.ts
+++ b/__tests__/js-specifier-ts-resolution.test.ts
@@ -1,0 +1,107 @@
+/**
+ * `.js` specifiers naming an on-disk `.ts` source.
+ *
+ * `"moduleResolution": "NodeNext"` REQUIRES the emitted extension on relative
+ * imports (`./impl.js`), while the file on disk is `./impl.ts`.
+ * EXTENSION_RESOLUTION only ever APPENDS a suffix, so such a specifier was tried
+ * as `impl.js.ts`, `impl.js.tsx`, … then bare `impl.js` — never `impl.ts`.
+ * Import resolution returned null for EVERY relative import in the project and
+ * the resolver fell through to the name-matcher, which cannot cross a rename,
+ * so an aliased or default import reported a false 0 callers.
+ *
+ * Each fixture below imports under a DIFFERENT local name than the declaration,
+ * so the name-matcher cannot rescue the edge — the assertion isolates
+ * import-based resolution.
+ */
+import { describe, it, expect, afterEach } from 'vitest';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as os from 'os';
+import CodeGraph from '../src/index';
+
+describe('.js specifiers resolve to their TS source', () => {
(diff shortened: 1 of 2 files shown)


valkyriweb:
Companion: #1485 (alias bindings). Independent — either order; whichever lands second may need a trivial rebase of `import-resolver.ts`.

Issue #1447, opened 2026-07-24
godot gdscript support


EndrrSB:
https://docs.godotengine.org/en/stable/about/introduction.html
https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_basics.html

Godot  GDScript  PLZ

BetterToAutomateTheWorld:
Bumping this issue, working in a small team on gscript, that would be great to have it supported in codegraph! :)

virtualzavie:
Duplicate of #1218

Issue #1445, opened 2026-07-24
Please support GDScript for godot engine


Muxczf:
Please support GDScript for godot engine

BetterToAutomateTheWorld:
Bumping this PR, working in a small team on gdscript, that would be great to have it supported in codegraph! :)

virtualzavie:
Duplicate of #1218

Issue #1218, opened 2026-07-08
Add GDScript Support (Godot Game Engine)


PadraicPeterson:

Unknownuserfrommars:
Yes please add Godot suport

Derillo:
Feature Request: GDScript (Godot 4) support — tree-sitter grammar exists and is actively maintained

### Language

**GDScript** — Godot Engine's primary scripting language (Godot 4.x+, dynamically typed, indentation-based, Python-like). Extension: `.gd`.

### Context

GDScript is one of the most widely used languages in game development, shipped with every Godot project (the engine had ~1.6M unique downloads in 2023 and the user base keeps growing). There is currently no static-analysis / call-graph tool in the Godot ecosystem that does cross-file symbol resolution — which is exactly the gap CodeGraph fills so well for TypeScript, Python, and Lua.

I'm working on a Godot 4.7 project (`osolemio-godot` — a music learning game) and CodeGraph currently silently ignores every `.gd` file: `codegraph status` reports "Files: 30 / Nodes: 0 / Edges: 0" because the 30 indexed files are all YAML configs, none of the GDScript is picked up. For a Godot project, CodeGraph effectively degrades to "not installed".

### Tree-sitter grammar — already exists and battle-tested

Unlike several languages CodeGraph would have to bootstrap, GDScript does **not** need a grammar written from scratch:

| Grammar repo | Stars | Activity | Notes |
|---|---|---|---|
| [`PrestonKnopp/tree-sitter-gdscript`](https://github.com/PrestonKnopp/tree-sitter-gdscript) | ~109 | **Active** — last push Jul 13 2026 | De-facto standard JavaScript grammar; MIT |
| [`GDQuest/GDScript-formatter`](https://github.com/GDQuest/GDScript-formatter) | ~395 | Active | Production GDScript formatter (Rust) built **on top of tree-sitter** — proves the grammar holds up on real-world codebases |
| [`neovim-treesitter/nvim-treesitter-queries-gdscript`](https://github.com/neovim-treesitter/nvim-treesitter-queries-gdscript) | — | Active | Existing highlight queries — useful starting point for the extraction queries |

So both the parser and a first set of tree-sitter queries already exist, are maintained, and have production consumers. The upstream tree-sitter org itself doesn't list it under "official parsers", but CodeGraph already ships community grammars (e.g. Luau), so this isn't a blocker.

### What GDScript nodes/edges CodeGraph would want to extract

[...]

PR #1499, opened 2026-08-02
feat: add GDScript language support

cosmicnag:
## Summary

Add GDScript support for Godot Engine projects.

## Changes

- Add tree-sitter-gdscript WASM grammar (PrestonKnopp/tree-sitter-gdscript, MIT)
- Wire gdscript into grammars.ts (WASM map, EXTENSION_MAP, VENDORED_WASM_LANGS)
- Implement gdscript extractor: class_name_statement, function_definition, enum_definition
- Handle preload/load as import edges, extends as extends edges
- Register extractor in languages/index.ts
- Add GDScript display name

## Testing

Tested on Godot 4.7 project : 
- 228 GDScript files indexed
- 2,749 nodes, 4,404 edges extracted
- class_name declarations, functions, enums, preload/load edges all working

Addresses #1218, #1445, #1447.


BetterToAutomateTheWorld:
Bumping this PR, working in a small team on gdscript, that would be great to have it supported in codegraph! :)

src/extraction/wasm/tree-sitter-gdscript.wasm
virtualzavie, line 1: This is the part of the PR that makes me the most uncomfortable.
How can this binary be trusted? If it comes from somewhere else, where is the checksum to validate that it is legitimate?

cosmicnag: Fair push-back, and you're right to raise it — the binary in this PR had no recorded provenance at all, which is a problem regardless of whether the binary itself was fine. I've replaced it with a reproducible build and documented the source.

**What the binary is now:** an unpatched, deterministic build of [`PrestonKnopp/tree-sitter-gdscript`](https://github.com/PrestonKnopp/tree-sitter-gdscript) at `8ecb27e` (MIT) — `v6.1.0` plus 13 commits.

**Why a source build and not a package:** GDScript isn't in `tree-sitter-wasms` (36 grammars, no GDScript), so there was no drop-in build. The obvious fallback, npm `tree-sitter-gdscript@6.1.0`, is 13 commits behind master and fails to parse two files that master parses — `addons/gut/error_tracker.gd` and `test/unit/ui/test_world_health_bar_3d.gd` (447/453 vs 449/453 clean on the test corpus). Upstream's `package.json` still reads `6.1.0` on master, so **the version number cannot identify the source — only checksums can.** That's why they're recorded.

**Checksums to validate against:**

| | |
|---|---|
| `tree-sitter-gdscript.wasm` | `5c65071d058206320fd9d38d55748e268b71cb7b3586cde0fd68f94ffea5728a` (338,854 B, ABI 14) |
| `src/parser.c` @ `8ecb27e` | `48a64c3b52b48dc544be2625c6f4b2f007d52075339a24ea95c3cbde4afb0922` |
| `src/scanner.c` @ `8ecb27e` | `175188924474f3265c49ac364695036136eb69ced21ac5e6a23e3a9d9c4a2e75` |

The two source checksums are the ones that let you confirm the wasm corresponds to the claimed upstream source, rather than being some other binary that happens to parse GDScript.

**You can rebuild it rather than trust my copy.** `docs/grammars/tree-sitter-gdscript.md` records the exact recipe, pinned to `tree-sitter-cli@0.25.10` and `emscripten/emsdk:4.0.4` (`sha256:47d573d5a86379a06f850de200d69407e6baa2d2f9c19d9e156a67db57f80f2f`). The build is bit-for-bit reproducible — every rebuild produced an identical sha256, so a rebuild-and-compare is a real check. `parser.c` is committed upstream (no `generate` step), which is why it reproduces cleanly.

**Behavioural check on top of the hash:** `node scripts/add-lang/check-grammar.mjs <wasm> <sample.gd> 20` → ABI 14, 20/20 clean parses, `PASS`. And swapping this wasm in for the previous ABI-15 binary in this branch changes nothing in the index: same 360 files, 4,835 nodes, 9,465 edges, same 4 pre-existing parse failures. The previous binary parses identically, so it appears to have come from the same source — but built with an unpinned toolchain, which is the actual thing you were objecting to.

If you'd prefer the provenance in a different form (e.g. one manifest covering the whole `src/extraction/wasm/` directory instead of a per-grammar doc), say the word and I'll restructure it.

PR #1503, opened 2026-08-03
fix(installer): stop swapping symlinked config files for detached copies (shared AGENTS.md setups)

meganemura:
## Summary

`codegraph install` breaks symlinked config files. The installer writes a temp file and then renames it onto the destination. But `rename` replaces the symlink itself, not the file it points to. So the symlink turns into a plain copy.

This hurts setups that share one `AGENTS.md` across agents:

```bash
ln -sf ~/AGENTS.md ~/.claude/CLAUDE.md
ln -sf ~/AGENTS.md ~/.codex/AGENTS.md
```

After install, the link is gone, and edits to the shared file no longer reach the agent. Dotfiles-managed configs break the same way.

The fix: resolve the symlink first, then write to the real file. The write stays atomic (temp file + rename in the target's own directory). All installer writes go through this one helper, so JSON, TOML, and markdown files are all covered. If several agents link to the same shared file, the block is written only once; the other installs report `unchanged`.

## Relationship to #433

@0x1306a94 found this problem first and proposed the same core fix in #433 — full credit to them. This PR differs in three ways:

- It follows whole symlink chains, including dangling links (links whose target does not exist yet). #433 resolves only one hop there.
- It leaves uninstall unchanged: when the file would become empty, we still remove the link, never the user's real file.
- Tests target today's API (`upsertInstructionsEntry` with the real markers) and include the shared-`AGENTS.md` case. A CHANGELOG entry is included, crediting #433.

## Test plan

- [x] `npx vitest run __tests__/installer-targets.test.ts` — all pass (6 new symlink tests), macOS
- [x] `tsc --noEmit` clean


davdittrich:
Confirmed independently against `@colbymchenry/codegraph` 1.5.0 on Linux. This remains reproducible in the released installer.

A real setup used shared global agent instructions through symlinks, including:

```text
~/.claude/CLAUDE.md -> shared instructions
~/.codex/AGENTS.md -> shared instructions
~/.gemini/GEMINI.md -> shared instructions
```

Running `codegraph install` replaced the Claude and Codex symlink directory entries with regular files. The resulting Claude file contained the previous shared instructions plus the CodeGraph marker block; the Codex file contained the CodeGraph block. An earlier installer pass had done the same to the Gemini link.

The observed filesystem timestamps align with the installer run, and the written block is byte-identical to `CODEGRAPH_INSTRUCTIONS_BLOCK`. The released `atomicWriteFileSync` still writes `<path>.tmp.<pid>` and then calls `renameSync(tmpPath, filePath)`, which replaces the symlink inode rather than writing through it.

This has security-relevant integrity impact: users can silently lose a single shared policy source and end up with divergent per-agent instruction files without any warning. PR #1503 addresses the exact failure mode and its full-chain handling also covers shared and dangling-link configurations. Please consider this an independent real-world confirmation that the fix is still needed in 1.5.0.

meganemura:
@davdittrich Thank you for the independent confirmation on 1.5.0. Reproducing it on Linux with three shared links at once says far more than my own report could. Your timestamp check and the byte-identical block match what the tests in this PR cover.

@colbymchenry Given the report above, this still affects released 1.5.0. Please consider it for merging when you have time.

Issue #1086, opened 2026-07-01
getCallersRecursive/getCalleesRecursive: depth guard returns before visited.add, causing duplicate results


inth3shadows:
**File:** `src/graph/traversal.ts`
**Lines:** 246–249 (`getCallersRecursive`), 296–299 (`getCalleesRecursive`)

**Defect:** The depth guard (`if (currentDepth >= maxDepth || visited.has(nodeId)) return;`) at L246/L296 returns without executing `visited.add(nodeId)` at L249/L299, so when `currentDepth >= maxDepth` fires the caller/callee node is never marked visited; a node reachable via multiple edges from the same parent is then pushed to `result` once per edge.

**Failure trace:**
- Function `X` has two `calls` edges from `Y` (e.g., `Y` calls `X` twice in different expressions).
- `getCallersRecursive("X", 1, 0, result, visited)`: depth 0 < maxDepth 1 → continue; `visited.add("X")`.
- Push loop, edge e1 (`Y→X`): `!visited.has("Y")` = true → `result.push({Y,e1})`; recurse `("Y", 1, 1, …)` → depth 1 ≥ maxDepth 1 → **return without `visited.add("Y")`**.
- Push loop, edge e2 (`Y→X`): `!visited.has("Y")` = true (Y was never added) → `result.push({Y,e2})` → **duplicate**.
- With default `maxDepth=1` (as used in the public `getCallers`/`getCallees` at `src/index.ts`) any function called by the same caller via two edges produces a duplicate entry.

**Expected fix:**
Move `visited.add(nodeId)` to immediately after the `visited.has(nodeId)` arm of the guard and before the depth arm fires — or, equivalently, split the guard into two separate checks and add to `visited` unconditionally after the `visited.has` check:

```ts
if (visited.has(nodeId)) return;
visited.add(nodeId);                // mark before depth check
if (currentDepth >= maxDepth) return;
```

Alternatively, add `visited.add(callerNode.id)` (resp. `calleeNode.id`) in the push loop before the recursive call, to prevent the same node being pushed more than once per result set.

inth3shadows:
---

**Verification & metadata** (added after filing)

**Anchor:** Introduced in `cc6e7a5c` (2026-01-18, the original graph-traversal implementation); present in every release through **v1.1.6** and still on `main` @ `712a406`. Not a recent regression.

**These 5 issues are really 3 root defects:**
- *Depth guard returns before `visited.add`* → #1086, #1089
[...]

Issue #693, opened 2026-06-05
CodeGraph does not index call edges from anonymous/lambda functions


Cyclone1070:
The call graph indexer misses call edges originating from anonymous/lambda function bodies. Functions called only from inline closures (e.g., cobra `RunE` handlers, goroutine literals, callback closures) appear to have zero callers in the graph.

### Steps to Reproduce

Given a codebase like:

   ```go
   // cmd/factory.go
   func Wire() (*Deps, error) { ... }

   // cmd/root.go
   var rootCmd = &cobra.Command{
       RunE: func(cmd *cobra.Command, args []string) error {
           deps, err := Wire()  // <-- call from anonymous closure
           ...
       },
   }
 ```

codegraph_callers("cmd.Wire") → returns "No callers found"

colbymchenry:
Fixed in #744 (merged to `main`) — thanks for the clear cobra example, it pinpointed the gap exactly.

The call edge from the closure was actually being captured, but it leaked to the **file** node instead of a meaningful symbol: Go's top-level `var` initializer was walked with no scope, so a `Wire()` call inside `var rootCmd = &cobra.Command{ RunE: func(){ Wire() } }` attributed to `root.go` rather than to a caller you'd recognize. (JS/TS didn't have this gap — an arrow-in-const becomes a named node whose calls attribute correctly.)

Now the initializer walk is scoped to the declared symbol, so a call inside **any** anonymous `func_literal` at package level — a cobra `RunE` handler, a goroutine literal, a callback closure, nested closures, struct-field/slice/map closures — attributes to the enclosing `var`. So `codegraph_callers("Wire")` returns `rootCmd`, and `impact` follows it.

Validated on **cli/cli** (858 Go files, the canonical cobra app): node and edge counts are unchanged and file-level dependents are byte-identical to before (no regression), while 62 top-level-closure calls correctly moved from file-attributed to var-attributed.

Ships in the next release; re-index (`codegraph index -f`) to pick up the corrected attribution.

Note: in-function closures already attributed to their enclosing function and still do; this was specifically about closures at package scope.

PR #1511, opened 2026-08-05
Walk declaration initializers scoped to the declared symbol — Kotlin, Java, TS/JS, Scala, Rust, Python (#693 for the other six languages)

### LINKED ISSUE #1510: Declaration initializers are still unwalked (or unscoped) outside Go — #693 was fixed for Go only
### Summary

`#693` ("CodeGraph does not index call edges from anonymous/lambda functions") was fixed for **Go only**, by #744. The same gap is still open in **Kotlin, Java, TypeScript/JavaScript, Scala, Rust and Python** — and in most of them it is worse than in Go: the calls are not merely attributed to the file node, they never reach the graph at all.

The consequence is the one #693 described: a function reached only through a declaration's initializer looks like it has **no callers**, and `impact` under-reports its blast radius. On Android/MSDK code, on Scala, and on any Python module that wires itself up at import time, that is most of the wiring.

One correction to #693's closing note, which says JS/TS didn't have this gap: that holds only for `const f = () => …` (an arrow value delegates to `extractFunction` and gets its own node). For every other JS/TS initializer shape the gap is real — see the table.

### Repro, one file per language

```kotlin
class C {
    private val fieldLambda: () -> Unit = { target() }        // no caller
    private val samField = Runnable { target() }              // no caller
    private val plain = compute()                             // no caller
    private val delegated by lazy { compute() }               // no caller
    val sameLine: Int get() = compute()                       // no caller
    init { val cfg = load() }                                 // `load` vanishes
    val (a, b) = makePair()                                   // `makePair` vanishes
}
object Holder { val cb: () -> Unit = { hit() } }              // no caller
```

```java
class T {
    private final Runnable fieldLambda = () -> target();                      // no caller
    private final Runnable l = new LocationListener() { … target() … };       // class + method invisible
    private final int eager = compute();                                      // no caller
}
```

```ts
const cfg = load();                                  // caller recorded as the FILE node
const obj = { handler: () => target() };             // nothing at all — no member node, no edge
const list = [() => target()];                       // caller recorded as the FILE node
export const ok = { handler: () => target() };       // works (exported object-of-functions path)
```

```scala
class C {
  val fieldLambda: () => Unit = () => target()   // no caller
  val direct = compute()                         // no caller
  lazy val lazily = compute()                    // no caller
}
```

```rust
const LEN: usize = compute_len();                        // no caller
static REGISTRY: Lazy<Cfg> = Lazy::new(|| build_cfg());  // no caller
```

```python
APP = compute()                        # no caller
handler = lambda: target()             # no caller
MAPPING = {"a": compute()}             # no caller
first, second = compute(), other()     # no caller, and no symbol either
```

### Where each one goes wrong

| language | mechanism | result |
|---|---|---|
| Kotlin | the `property_declaration` hook returns `true`; the dispatcher only runs `scanFnRefSubtree` | initializer, delegate and accessor bodies never walked |
| Java | `extractField` mints the field node and stops | `variable_declarator`'s `value` never walked; an anonymous class in a field initializer is never extracted at all |
| TS/JS | `extractVariable`'s walk runs with only the FILE on `nodeStack`; object literals are excluded from it outright | calls attributed to the file; non-exported object literals contribute nothing |
| Scala | the `val`/`var` hook returns `true` | `value` field never walked |
| Rust | `const_item`/`static_item` ride `extractVariable`'s generic fallback | `value` field never walked |
| Python | the `assignment` branch mints the node and stops | `right` field never walked |

### How much it costs, measured

Each number is one extractor arm (wasm) diffed against upstream `main` over a real tree, counting refs that appear/disappear/change owner:

| language | corpus | lost | re-attributed | new |
|---|---|---|---|---|
| Kotlin | 113-file Android app (MSDK callbacks) | 0 | 0 | **+368** |
| Java | 409-file Android SDK sample tree | 0 | 0 | **+651** (+21 nodes: anonymous listener/`Parcelable.Creator` classes) |
| TS/JS | 499 files across three projects | 0 | **780** (file → declaring constant) | +228 |
| Scala | 32-file SpinalHDL project | 0 | 0 | **+812** |
| Rust | 282 files across three projects | 0 | 0 | +86 |
| Python | 799 files across three projects | 0 | 0 | **+3057** |

Node and edge counts are unchanged everywhere except Java (+21 nodes, the anonymous classes that were previously invisible).

### Two smaller defects found while measuring, not part of this family

- **Rust phantom node.** `extractVariable`'s generic fallback mints a node for *every* direct `identifier` child, so `const MAX: u32 = OTHER;` produces `MAX` **and** a spurious `OTHER`.
- **C# is wider than a field initializer.** Beyond `variable_declarator` (whose initializer is a bare child, with no `value` field), `extractProperty` walks neither the accessor bodies, nor `=> expr`, nor `{ get; } = …`. For decompiled/reverse-engineered C#, where logic often lives in property getters, that is the larger share of the recall loss.

Happy to split either of those out into its own issue.

### Fix

PR incoming — Kotlin, Java, TS/JS, Scala, Rust and Python, both extractor arms (TS + the Rust kernel), kernel parity preserved per language. C# deliberately left out; it needs its own shape and its own validation.

danusha2345:
Closes #1510. Ports #693's fix (merged as #744 for Go) to **Kotlin, Java, TypeScript/JavaScript, Scala, Rust and Python**.

The shape is the same everywhere: after minting the declaration's node, walk its initializer through `visitFunctionBody` with that node pushed on the scope stack, so the calls inside it are attributed to the symbol that owns them instead of leaking to the file node — or, in most of these languages, vanishing outright.

Every language moves in **both arms together** — `src/extraction/` and the matching `codegraph-kernel/src/` walker — so `kernel-*-parity` stays byte-identical; that is asserted per language, and each torture fixture gains the new shapes.

---

## Per language

**Kotlin** — three commits. The `property_declaration` hook consumed the whole subtree, so the dispatcher only fn-ref-scanned it.

- The initializer (`= …`) and a `property_delegate` (`by lazy { … }`) are now walked under the property.
- So is an accessor, on **either** line: a same-line `val x: Int get() = f()` nests under the declaration, an own-line one parses as a following **sibling**. The property claims its own-line accessors and the accessor branch skips what the property claimed — the skip re-derives the property's kind instead of remembering it across nodes, so a destructured or local declaration (which mints nothing and therefore claims nothing) keeps falling through. No cross-node state in either arm.
- `init { val cfg = load() }` and `val (a, b) = makePair()` mint no symbol of their own, so they are walked at the **enclosing** scope — an init block's calls belong to the class, like its other statements. Both used to disappear completely.
- The scope/kind decision moved into one function (`kotlinPropertyKind` / `property_kind`) now that two branches need it.

**Java** — `extractField` minted the node and stopped. The declarator's `value` is now walked under the field. The anonymous-class form lost more than edges: `new LocationListener() { … }` in a field initializer was never extracted at all, so the class and its overrides did not exist as symbols. `extractField` is shared, but the walk is keyed on the `value` **field**, which only Java's `variable_declarator` carries — C# (bare child), VB.NET (`initializer`) and PHP (`default_value`, separate branch) are untouched.

**TypeScript / JavaScript** — two defects in `extractVariable`'s JS-family branch.

1. The walk ran with only the FILE on `nodeStack`, so `const cfg = loadConfig()` recorded the *file* as loadConfig's caller. (This is the part #693's closing note assumed JS/TS didn't have; it holds only for `const f = () => …`, where the arrow value delegates to `extractFunction`.)
2. Object literals were excluded from the walk outright, on the grounds that their function-valued members are extracted individually below — but that only happens for **exported** consts. `const obj = { handler: () => target() }` therefore contributed nothing at all.

Both fixed by walking the initializer scoped to the declared symbol and skipping only the shapes whose members really are extracted one-by-one (exported object-of-functions, RTK endpoints, Pinia setup, Vue store collections) — walking those too would double-count each member arrow's calls.

**Scala** — the `val`/`var` hook returned `true` without walking; the grammar exposes `value` directly, so this is the closest analogue to the Go fix. Scala puts almost everything in a `val`, which is why it shows the largest per-file gain here.

**Rust** — `const_item`/`static_item` ride `extractVariable`'s generic fallback. That fallback is shared, so the walk is gated to Rust; the declared symbol is identified by the `name` **field**, which leaves the fallback's phantom node for a bare-identifier initializer (`const MAX: u32 = OTHER;` → `MAX` plus a spurious `OTHER`) exactly as it is. Separate defect, deliberately not folded in — noted in #1510.

**Python** — the `assignment` branch minted the node and stopped. `right` is now walked under the assigned name; a tuple target mints no symbol, so its RHS is walked at the enclosing scope rather than lost. Gated to Python — Ruby shares the branch. Class attributes are untouched: they never reach this branch (the dispatcher's class-scope gate excludes them), so their calls keep riding the class node; giving them symbols of their own is a separate question, since there are no nodes for them today at all.

---

## Two follow-on repairs the change surfaced

Both are in this PR because the initializer walk is what exposed them.

**CFML.** `<cfscript>` bodies are delegated to a separate extractor as if they were a whole module, so a `var x = helper()` local inside a `<cffunction>` minted a top-level variable node and the walk attributed `helper` to it. Those snippet-top-level non-callables are locals of the enclosing function; their refs now redirect to it, exactly like the snippet's file-attributed refs already did.

**`codegraph_explore` dynamic-dispatch links.** The section stopped listing a synthesized edge once the same pair also had a static call edge: it asked `getCallers`/`getCallees`, which return one row per **neighbour** (the #1086 de-dup), so the static edge hid the heuristic one. It now asks the edges directly. This is the RTK thunk case — `dispatch(innerThunk(n))` inside a thunk initializer is now a real static edge, which is the point, but the synthesized hop must still show up in the summary.

---

## Validation

Each language was measured by diffing one extractor arm against upstream `main` over real trees, classifying every ref change as *lost* / *re-attributed* (same file+kind+name+line, different owner) / *new*:

| language | corpus | lost | re-attributed | new | kernel parity |
|---|---|---|---|---|---|
| Kotlin | 113-file Android app | 0 | 0 | **+368** | 99/113 (14 grammar-deferred — same set as upstream) |
| Java | 409-file Android SDK sample tree | 0 | 0 | **+651**, +21 nodes | 409/409 |
| TS/JS | 499 files, three projects | 0 | **780** (file → constant) | +228 | 324/326 (2 deferred) |
| Scala | 32-file SpinalHDL project | 0 | 0 | **+812** | 32/32 |
| Rust | 282 files, three projects | 0 | 0 | +86 | 226/230 (4 deferred) |
| Python | 799 files, three projects | 0 | 0 | **+3057** | 6221/6221 sweep |

Node and edge counts are unchanged everywhere except Java (+21 nodes — the anonymous listener and `Parcelable.Creator` classes that were previously invisible).

End-to-end on a mixed 110-Kotlin / 100-JS / 37-Go project, full `codegraph init` both sides: **4311 → 4318 nodes, 12106 → 12254 edges, 0 edges lost.** 13 edges changed identity — 9 JS initializer calls moving off the file node onto their declaring constant, and 4 Kotlin `run(…)` call sites that used to name-match a JavaScript test helper in `desktop/test/` and now match a Kotlin `run` override.

The headline case that started this: an MSDK `ICameraStreamManager.CameraFrameListener` field now appears in `callers` of the method it invokes.

`npm test` — **2930 passing**, on top of upstream `main` as of a7db24d.

## Test coverage added

- `__tests__/extraction.test.ts` — a regression test per language (Kotlin ×3 shapes, Java, TS/JS, Scala, Rust, Python), each asserting the exact set of owners for the calls in question, including the cases that deliberately still ride the enclosing scope.
- Kernel-parity torture fixtures extended for all six languages, so the two arms are pinned on the new shapes (incl. CRLF variants).
- `__tests__/function-ref.test.ts` — one expectation updated: a function-as-value in an initializer (`REGISTRY = {"org": Serializer}`) now produces a reference from **both** the assigned name and the file node, because the dispatcher's own scan runs either way. That is the shape #744 already ships — verified against the Go arm on the same input — so the two languages stay consistent.

## Deliberately not in this PR

- **C#** — needs its own shape (the initializer is a bare child, no `value` field) and, more importantly, `extractProperty` walks neither the accessor bodies nor `=> expr` nor `{ get; } = …`. Bigger than a field initializer; described in #1510.
- **Rust's phantom node** for a bare-identifier initializer.
- **Kotlin's own-line accessor asymmetry is now closed**, but the analogous Kotlin `object` → `INSTANCE` gap reported under #750 is not touched here.
- Ruby, VB.NET, PHP, Solidity — they share the touched branches but spell their initializer differently, so each is a separate change with its own validation.


bompus:
Verified on a real TypeScript repo (Chrome MV3 extension, 582 files, TS/JS/Vue/markdown, Windows 11, tree-sitter wasm walker, kernel off). Branch: this PR merged onto current `main` (b9ca4b7); control build indexed the same tree without the PR.

Merges clean onto main. On our tree: `calls` edges whose source is a `variable`/`constant`/`property` node 0 → 311, edges 33,927 → 34,133, nodes +4, no change to self-edge or import metrics. Samples look right: `MESSAGE_HANDLERS -> getRecLock` (handler map in a service worker), `draftStateApplyLane -> createIdleLane`, `urgencyList -> ttdPriorityPctOf` in a `.vue` `<script setup>`. PR test files: 652/652 pass.

danusha2345:
Merged current `main` (`3adf067`, post-#1770) into this branch: head `752ddef`. The only conflict was the CHANGELOG entry; no source conflict. `tsc` clean, the branch's tests pass.

danusha2345:
Merged current `main` (`43271f3`, post-#1771) into this branch: head `fa69cbb`. Conflicts were the CHANGELOG and the kernel-parity torture.js fixture (both fixture blocks kept). Kernel rebuilt; `tsc` clean; the branch's tests and the kernel parity suites pass.

danusha2345:
Merged current `main` (`85550eb`) into this branch: head `f188330`. The only conflict was the CHANGELOG entry; no source conflict. `tsc` clean, the branch's tests pass.

danusha2345:
Merged current `main` (`8733c28`) into this branch: head `ff12a1f`. The only conflict was the CHANGELOG entry; no source conflict. `tsc` clean, the branch's tests pass.

danusha2345:
Merged current `main` (`8c04734`) into this branch: head `d7b657d`. The only conflict was the CHANGELOG entry; no source conflict. `tsc` clean, the branch's tests pass.

danusha2345:
Merged current `main` (`ee83636`) into this branch: head `2997899`. The only conflict was the CHANGELOG entry; no source conflict. `tsc` clean, the branch's tests pass.

danusha2345:
Merged current `main` (`e720f6c`) into this branch: head `aaf9d4e`. The only conflict was the CHANGELOG entry; no source conflict. `tsc` clean, the branch's tests pass.

danusha2345:
Merged current `main` (`de5adba`) into this branch: head `056f396`. The only conflict was the CHANGELOG entry; no source conflict. `tsc` clean, the branch's tests pass.

danusha2345:
Merged current `main` (`cece072`) into this branch: head `af098d9`. The only conflict was the CHANGELOG entry; no source conflict. `tsc` clean, the branch's tests pass.

danusha2345:
Merged current `main` (`040ba38`) into this branch: head `b26f888`. The only conflict was the CHANGELOG entry; no source conflict. `tsc` clean, the branch's tests pass.

danusha2345:
Merged current `main` (`374b3b4`) into this branch: head `c236fda`. The only conflict was the CHANGELOG entry; no source conflict. `tsc` clean, the branch's tests pass.

danusha2345:
Merged current `main` (`71d049c`) into this branch: head `d282f9e`. The only conflict was the CHANGELOG entry; no source conflict. `tsc` clean, the branch's tests pass.

colbymchenry:
Superseded by Forge land https://github.com/colbymchenry/codegraph/pull/1802, which squash-lands this branch (`d282f9e8`) onto current main as `9664bb7a` (CHANGELOG condensed/credited; implementation, tests, fixtures and follow-ups retained).

Thanks @danusha2345 — Linux fail to pass verified across Kotlin, Java, TS/JS, Scala, Rust and Python (wasm + rebuilt native kernel) for #1510. Closing this PR as superseded.

colbymchenry:
Superseded by #1802

Issue #1093, opened 2026-07-01
C++ forward declarations (`class Foo;`) index as phantom class nodes, masking the real definition


luoyxy:
### What happens
In a large C++/Unreal-Engine codebase, `codegraph_explore` for a heavily used
class (e.g. `APXCharacter`) returns mostly forward-declaration sites and even
picks one as the blast-radius representative, while the real definition (with
members and callers) is crowded out.

### Root cause
A `class Foo;` forward declaration parses as a bodiless `class_specifier`.
`extractStruct` (#831) and `extractEnum` skip their bodiless forms, but
`extractClass` does not — so every forward decl (repeated across dozens of
headers) mints a phantom bodiless `class` node that competes with the single
real definition.

### Expected
Bodiless class specifiers in C/C++ should be skipped, exactly like bodiless
structs/enums, so only the real definition is indexed. Languages where a
bodiless class is a complete definition (Kotlin `class Empty`, Scala) must be
unaffected.

### Repro
Index any C++ project where a class is forward-declared in many headers, then
`codegraph_explore` that class name — the results are dominated by
`class Foo;` sites instead of the definition.

colbymchenry:
Fixed in #1095 (merged to `main`) — it'll ship in the next release.

Your root-cause analysis was exactly right: a `class Foo;` forward declaration parses as a bodiless `class_specifier`, and while `extractStruct` (#831) and `extractEnum` already skipped their bodiless forms, `extractClass` didn't — so each forward decl across your headers minted a phantom `class` node that competed with, and could even be picked as the blast-radius representative over, the single real definition. A probe reproduced it exactly: a class forward-declared twice then defined produced 3 `class` nodes instead of 1.

The fix mirrors the struct/enum skip, but scoped to C/C++ via an opt-in extractor flag so that languages where a bodiless class *is* a complete definition — Kotlin `class Empty`, Scala `case object` / `trait` — stay indexed. Post-fix, `APXCharacter` collapses to just its real definition (members and callers intact), with regression tests covering all of the above.

Thank you for the detailed report and the precise diagnosis — this is exactly the kind of feedback that makes CodeGraph better with every update. 🙏

Issue #831, opened 2026-06-12
C# Records are not Indexed


b564518128:
After running `codegraph index` with a C# repo, Records are not getting indexed into class kinds. In fact, running `codegraph query` will not return records (except file if the record name is the same as the filename).

Reproduction Steps:
1. Indexing a C# repo that has record definition
2. Run `codegraph query` on the record name

colbymchenry:
Confirmed — and already fixed on `main`, shipping in the next release (which is imminent).

You're exactly right about the symptom: on the released **0.9.9**, C# `record` declarations produce no symbol at all, so `codegraph query` only finds the *file* when its name happens to match. Record support landed in #708 (merged June 6, after 0.9.9 was cut): `record` / `record class` / `record struct` declarations now index as type symbols, so MediatR/CQRS-style codebases — where commands, queries, and DTOs are nearly all records — resolve properly.

Verified against a fixture covering eleven record forms (positional one-liners, bodied records, `record class`, `record struct`, `readonly record struct`, derived, generic, attributed, partial, nested-in-class, and records following C#12 syntax): 0.9.9 indexes none of them; current `main` indexes all, and each is returned by `codegraph query`.

After upgrading, run `codegraph index -f` once so the new extraction applies to your existing index.

PR #1514, opened 2026-08-06
fix(rust): index unit structs — bodiless is a definition, not a forward declaration (#1513)

### LINKED ISSUE #1513: rust: unit structs (`struct Unit;`) are not indexed, taking their `impl Trait for` edges with them
`codegraph` extracts **no symbol at all** for a Rust unit struct (`struct Unit;`),
so the type is absent from the graph — and so is everything attached to it. The
most visible loss is `impl SomeTrait for Unit`: the edge never forms, because its
source endpoint does not exist.

Brace structs (`struct S { .. }`) and tuple structs (`struct S(u32);`) are
extracted normally. Only the unit form is affected.

The grammar parses the file correctly, so this is in the symbol-extraction layer,
not the grammar.

## Environment

- codegraph: 1.5.0 (npm), reproduced again on `main` @ `d6d1728`
- Backend: node:sqlite (WAL)
- Platform: Linux x64
- Both extraction paths affected: the native kernel and the wasm/TS walker
  (`CODEGRAPH_KERNEL=0`)

## Repro

`src/lib.rs`:

```rust
pub struct UnitStruct;
pub struct TupleStruct(pub u32);
pub struct BraceStruct { pub x: u32 }

pub trait Greet { fn hi(&self) -> String; }

impl Greet for UnitStruct  { fn hi(&self) -> String { "unit".into()  } }
impl Greet for TupleStruct { fn hi(&self) -> String { "tuple".into() } }
impl Greet for BraceStruct { fn hi(&self) -> String { "brace".into() } }
```

```
codegraph init .
sqlite3 .codegraph/codegraph.db \
  "SELECT kind, name FROM nodes WHERE kind IN ('struct','trait') ORDER BY name;"
```

## Expected

Three `struct` nodes and three `implements` edges.

## Actual

Two of each — `UnitStruct` is missing, and with it the `Greet <- UnitStruct`
realization.

```
struct  BraceStruct
struct  TupleStruct
trait   Greet
```

```
Greet <|.. TupleStruct
Greet <|.. BraceStruct
```

## Impact

Unit structs are the idiom for zero-sized markers, test doubles and stub
implementations, so the loss concentrates in test and adapter layers rather than
being spread evenly.

On a 2,669-node Rust/TypeScript repository, re-indexing after a local fix
recovered **13 structs and 9 `implements` edges** — every one of them a unit
struct such as `SystemClock`, `StubTransport`, `MockEncoder`, `StubSettings`.
Asking for implementors of a trait, or the impact radius of one, silently omitted
these.

## Cause

`extractStruct` treats a missing body field as a forward declaration and returns
before minting the node. That is correct for C/C++ (`struct Foo;` really is a
forward declaration) but **Rust has no forward declarations** — a bodiless struct
is always a complete definition.

This looks like the same case two earlier patches already recognised:

- **#831** carved out C# positional records (`record struct M(decimal Amount);`)
  as "complete definitions with no body block".
- **#1093** added `skipBodilessClass` so a bodiless *class* is kept unless a
  language opts into skipping it.

Struct kept the unconditional skip, so Rust was never covered.

The kernel's Rust walker mirrors the behavior deliberately — it is listed in
`rustlang.rs`'s bug-for-bug quirk header and in the R7b port checklist — so both
walkers need the same change to stay at parity.

ctype-lab:
Fixes #1513.

## Summary

Rust unit structs (`struct Unit;`) never enter the graph, and every
`impl SomeTrait for Unit` loses its edge with them.

This is not a new class of bug — it is the **one case two earlier patches
left uncovered**. Both established that "no body" does not mean "not a
definition"; struct just never got the same treatment.

| | what it settled | left uncovered |
|---|---|---|
| **#831** | C# positional records (`record struct M(decimal Amount);`) are "complete definitions with no body block" — hardcoded carve-out in `extractStruct` | only that one node type |
| **#1093** | `skipBodilessClass`: a bodiless **class** is kept unless a language opts into skipping (C/C++ forward decls) | struct kept its unconditional skip |

The rule itself is correct and stays: in C/C++, `struct Foo;` *is* a forward
declaration and must be skipped. **Rust has no forward declarations** — a
bodiless struct is always a complete definition. The C-shaped rule was simply
applied to a language it does not describe.

## The fix

`allowBodilessStruct` — the struct-side counterpart to `skipBodilessClass`,
opt-in, set for Rust only. No other language changes behavior.

```ts
// src/extraction/languages/rust.ts
allowBodilessStruct: true,
```

The polarity is inverted relative to `skipBodilessClass` because the two
defaults differ (bodiless class → kept by default; bodiless struct → skipped
by default). That is documented on both flags so the asymmetry is not a
surprise later.

The kernel's Rust walker carried the same guard, placed before `create_node`,
and its header listed the behavior in the mirrored bug-for-bug quirk list.
Node creation now happens first and the body is walked only when present —
the shape `extract_interface` in that same file already had. Kernel↔wasm
parity is preserved.

Brace and tuple structs were never affected: both carry a body field
(`field_declaration_list` / `ordered_field_declaration_list`). Only the unit
form fell through.

### The whole change

Four lines of logic across four source files:

```diff
  src/extraction/tree-sitter.ts
- if (!body && node.type !== 'record_declaration') return;
+ if (!body && node.type !== 'record_declaration' && !this.extractor.allowBodilessStruct)
+   return;

  src/extraction/tree-sitter-types.ts
+ allowBodilessStruct?: boolean;

  src/extraction/languages/rust.ts
+ allowBodilessStruct: true,

  codegraph-kernel/src/rustlang.rs
- let Some(body) = node.child_by_field_name("body") else { return };   // before create_node
+ let Some(body) = node.child_by_field_name("body") else { return };   // after create_node
```

The kernel line is moved, not changed — the early return simply happens after
the node is created instead of before it.

Everything else is comments, the two regression tests, and the docs that
described the old behavior as intended.

| | files | lines |
|---|---|---|
| source logic | 4 | **4** |
| source comments | (same 4) | 32 |
| regression tests | 1 | +29 |
| docs describing the old behavior | 3 | ±7 |
| CHANGELOG | 1 | +2 |

The comment weight is deliberate: `allowBodilessStruct` and
`skipBodilessClass` have **opposite polarity**, and without the reason
written on both flags the next reader will assume one of them is backwards.

Three places documented the skip as intended behavior and are updated with
the code:

- the kernel header's mirrored bug-for-bug quirk list (the entry is gone —
  it is no longer mirrored),
- `kernel-rustlang-parity`'s docstring, which listed "unit-struct skip"
  among the quirks the torture fixture pins,
- `docs/design/rust-lang-kernel-port-checklist.md`, which states it "remains
  the quirk reference for the walker" — its `struct_item` row is marked
  superseded rather than deleted, so the port survey stays readable as
  history.

## Reproduction

```rust
pub struct UnitStruct;
pub struct TupleStruct(pub u32);
pub struct BraceStruct { pub x: u32 }

pub trait Greet { fn hi(&self) -> String; }

impl Greet for UnitStruct  { fn hi(&self) -> String { "unit".into()  } }
impl Greet for TupleStruct { fn hi(&self) -> String { "tuple".into() } }
impl Greet for BraceStruct { fn hi(&self) -> String { "brace".into() } }
```

| | before | after |
|---|---|---|
| `struct` nodes | 2 — `UnitStruct` missing | 3 |
| `implements` edges | 2 | 3 |

On a real Rust/TypeScript repository (2,669 nodes): **+13 structs, +9
`implements` edges**. The recovered types are exactly the codebase's
zero-sized markers, test doubles and stub implementations — `SystemClock`,
`StubTransport`, `MockEncoder`, `StubSettings`, … — which is the idiom unit
structs exist for, so the loss concentrated in test and adapter layers.

## Tests

Two regression cases in `__tests__/extraction.test.ts`: one asserting all
three struct forms extract, one asserting `impl Trait for <unit struct>`
survives.

Verified they fail without the fix and pass with it, on **both** extraction
paths — the wasm path via `CODEGRAPH_KERNEL=0`, and the kernel path with a
staged build. `kernel-rustlang-parity` passes (it compares the two walkers,
so a one-sided fix would have failed it).

Full suite green: **170 files, 2,922 tests**.

## Note for the maintainer

`skipBodilessClass` (#1093) settled the class side as *keep by default*.
Struct is still *skip by default*, which is why this patch needs an
inverted-polarity flag rather than reusing the existing one.

Flipping struct's default to match would remove the asymmetry, but it touches
all 27 languages that declare `structTypes`, so I kept this PR to the
provable Rust case. Happy to do that follow-up if you would rather converge
the two.

Two other languages already work around the same guard with their own
extraction code (`solidity.ts`, `erlang.ts` — both comment on it), so a
general fix would likely simplify those too.


colbymchenry:
Superseded by Forge land https://github.com/colbymchenry/codegraph/pull/1800, which cherry-picks `a94c9dc9` onto current `main` (conflicts resolved for CHANGELOG / rust.ts / rustlang.rs / parity docstring while keeping the `allowBodilessStruct` design).

Thanks @ctype-lab — Linux fail→pass verified on both kernel and wasm paths for #1513. Closing this PR as superseded.

colbymchenry:
Superseded by #1800 (Forge cherry-pick of this fix onto current main).

PR #1516, opened 2026-08-06
fix(c,cpp,objc,rust): index union declarations

### LINKED ISSUE #1515: Union declarations are not indexed in C, C++, Objective-C, or Rust
In CodeGraph's C, C++, Objective-C, and Rust extractors, a `union` declaration
produces **no symbol at all**. The type is absent from the graph, and so is
everything attached to it.

`union_specifier` (C/C++/ObjC) and `union_item` (Rust) appear **nowhere** in the
extraction layer — no `<x>Types` list on the TS side, no dispatch branch in
either kernel walker:

```
$ grep -rn 'union_specifier\|union_item' src/ codegraph-kernel/src/
(no output)
```

Structs and enums beside them extract normally. The grammars parse unions fine,
so this is in symbol extraction, not the grammar.

## Environment

- codegraph: 1.5.0 (npm), reproduced again on `main` @ `d6d1728`
- Backend: node:sqlite (WAL)
- Platform: Linux x64
- C, C++, and Rust reproduce on both extraction paths: the native kernel and
  the wasm/TS walker (`CODEGRAPH_KERNEL=0`). Objective-C continues to use its
  existing wasm/TS extraction path.

## Repro

`src/lib.rs`:

```rust
pub union Reg { pub raw: u32, pub bits: [u8; 4] }
pub struct Ctl { pub n: u32 }

pub trait Describe { fn describe(&self) -> String; }

impl Describe for Reg { fn describe(&self) -> String { "reg".into() } }
impl Describe for Ctl { fn describe(&self) -> String { "ctl".into() } }

pub fn dump(r: &Reg) -> String { r.describe() }
```

`src/mem.c`:

```c
union Word { int i; char b[4]; };
struct Ctl  { int x; };
```

```
codegraph init .
sqlite3 .codegraph/codegraph.db "SELECT kind, name FROM nodes WHERE kind='struct';"
```

## Expected

`Reg`, `Ctl`, `Word`, `Ctl` — and two `implements` edges.

## Actual

The two unions are missing, and with them `Reg`'s realization of `Describe`:

```
struct  Ctl        (lib.rs)
struct  Ctl        (mem.c)
```

```
implements:  Ctl -> Describe          # Reg -> Describe never forms
```

## Impact

Three separate losses, and the second is the one that hurts.

**1. The type is gone.** It cannot be searched for, does not appear in
`codegraph_explore`, and any function taking `union Reg *` has no type to point
at.

**2. Its methods become orphans that point at a type which does not exist.**
This is worse than a plain omission, because the graph ends up internally
inconsistent rather than merely incomplete:

```
Describe::describe   <-  contains from  trait:Describe            ✅
Ctl::describe        <-  contains from  file:lib.rs + struct:Ctl  ✅
Reg::describe        <-  contains from  file:lib.rs               ❌   its qualifiedName says "Reg"
```

Asking what `Reg::describe` belongs to gets the file. The `Reg` in its own
qualified name resolves to nothing.

**3. Trait-dispatch synthesis silently under-reports.** The synthesized
(`provenance='heuristic'`) fan-out is `Describe::describe -> Ctl::describe` and
nothing else. A union implementor is not a candidate, because it is not a node —
so "who implements this trait" is quietly wrong rather than visibly incomplete.

In C the loss splits by form:

| form | today |
|---|---|
| `union U { … };` | **no node at all** |
| `typedef union { … } U;` | survives as `type_alias:U` — name kept, kind wrong, no members |

Unions are load-bearing in C: tagged variants, register overlays, protocol
frames, and — relevant to this repo — function-pointer dispatch tables, which
`c-fnptr-synthesizer.ts` finds by scanning nodes with `kind === 'struct'`. A
union-shaped dispatch table is invisible to it.

## Cause

`extractStruct` is only reached for node types registered in a language's
`structTypes`, and no language lists its union node type. Same in both kernel
walkers, where the dispatch is a hardcoded `kind == "struct_specifier"` /
`kind == "struct_item"` branch.

ctype-lab:
Fixes #1515.

## Summary

On upstream main (revision d6d1728, 2026-08-06), C, C++, Objective-C, and
Rust union declarations do not enter the graph at all. Tree-sitter gives
struct and union declarations different AST node types: in C-family languages,
struct declarations are named "struct_specifier" and unions are named
"union_specifier"; in Rust, they are "struct_item" and "union_item". The
extractor registrations and native C/C++ and Rust walkers recognize only the
struct forms, so extraction never starts for a union. Consequently a union
cannot be searched, referenced, or own members, and Rust impl Trait for Union
cannot attach to a type node.

The first three commits in this PR were an **unmerged intermediate repair**:
they made unions extract by routing them through `structTypes` and emitting
`kind: 'struct'`. That representation has never been merged or released.
Before review, the later commits replace it in the same PR with the correct
public model: `kind: 'union'`. Review this PR as its final head, not as two
separately shipped fixes.

| State | Union declaration in the graph |
| --- | --- |
| Upstream `main` | absent |
| Early, unmerged commits in this PR | present as `struct` |
| This PR head | present as first-class `union` |

The final change is therefore not a follow-up to a merged fix. It resolves
#1515 directly while preserving the semantic distinction that the temporary
representation would have lost.

## What Changes

`union` is appended to the TypeScript and native-kernel node-kind tables. The
append-only placement leaves all existing wire-code values unchanged. C/C++,
Objective-C, and Rust register union syntax separately from `structTypes`; the
shared extractors walk both aggregate forms through the same logic while
creating the explicit kind supplied by the registration.

```ts
structTypes: ['struct_specifier'],
unionTypes: ['union_specifier'],
```

The native C/C++ and Rust walkers mirror this dispatch. Union members, C++
member methods, Rust `impl` attachment, containment, and inheritance handling
continue to use the existing aggregate walk, but the node and its scope retain
`kind: 'union'`.

`typedef union { ... } word_t` keeps the typedef name and produces one named
union node, rather than an anonymous duplicate. A bodiless C/C++/Objective-C
declaration such as `union U;` remains a forward declaration and is skipped;
this PR changes node identity, not the forward-declaration rule.

## Downstream Audit

A separate node kind must be recognized wherever CodeGraph treats a node as a
type or class-like container. The final PR head updates:

- class-like scopes and method-owner lookup
- import/type resolution and search filtering
- context and MCP definition-kind lists
- name-match ranking for type instantiation
- C function-pointer dispatch-table synthesis
- native-kernel and wasm parity tables
- resolver promotion from `calls` to `instantiates`

The resolver coverage closes a C++ case the structural extraction tests do not
exercise: `Packet()` value-initializes a union. Extraction initially records a
call-shaped reference; after `Packet` resolves to a union, that edge must be
`instantiates`, as it already is for classes and structs. An `instantiates`
reference also prefers a union definition over a same-named function.

## Downstream Follow-up

Additional regression coverage verifies consumers that depend on unions being
a distinct container kind:

- C function-pointer dispatch through a named union field
- inline union dispatch tables with macro-built entries
- union dispatch tables declared through an object-macro type alias
- default context retrieval of a union definition
- an imported C++ union static-member call (`Ops::run()`)

The function-pointer cases cover both the TypeScript and native-kernel
scanners, so union-shaped dispatch tables remain visible after extraction
rather than being limited to struct-shaped tables.

## Reproduction

```rust
pub union Reg { pub raw: u32, pub bits: [u8; 4] }
pub struct Ctl { pub n: u32 }
pub trait Describe { fn describe(&self) -> String; }
impl Describe for Reg { fn describe(&self) -> String { "reg".into() } }
impl Describe for Ctl { fn describe(&self) -> String { "ctl".into() } }
```

```cpp
union Packet { unsigned int raw; };
void initialize() { Packet(); }
```

| Query/result | Upstream `main` | Final PR head |
| --- | --- | --- |
| Rust `Reg` node | absent | `union:Reg` |
| `impl Describe for Reg` attachment | absent | attached to `union:Reg` |
| C++ `Packet()` edge | no union target exists | `instantiates -> union:Packet` |
| Union kind available to graph consumers | no | yes |

## Compatibility

The wire table change is append-only, so codes for existing node kinds do not
move. Existing indexes must nevertheless be rebuilt after upgrade: any union
node produced by this PR is stored as `union`, while upstream `main` has no
union node to migrate.

## Tests

Regression coverage includes:

- Rust union extraction, `impl Trait for Union`, and method containment
- named C unions and skipped bodiless forward declarations
- `typedef union { ... } name_t` without an anonymous duplicate
- C++ union member ownership
- native-kernel to wasm parity fixtures for C/C++ and Rust
- C++ union value initialization promoted to `instantiates`
- union-over-function selection for `instantiates` references
- named, inline macro-built, and object-macro-alias C union dispatch tables
- default context retrieval of a union definition
- imported C++ union static-member resolution

Validated on the final PR head:

```text
npm run build
npm run build:kernel
npm test

170 test files, 2,932 passed, 6 platform/environment-gated skips
```


colbymchenry:
Thanks for this — well-built PR, and the diagnosis in #1515 was unusually precise.

Verified end-to-end before merging: unions extract identically on both the wasm and native-kernel paths, the orphan-method and `typedef union` cases are fixed, `union:Reg -> trait:Describe` forms, and a union-shaped C dispatch table now synthesizes its function-pointer edges (on `main` it produced no node and no edge at all). Modeling unions as a distinct kind rather than aliasing them onto `struct` was the right call, and appending to `NODE_KINDS` instead of reordering keeps the kernel wire contract intact.

I pushed two commits to your branch rather than round-tripping — hope that's alright:

**5b0c4b8** — Impact #3 from your issue was still open. `interfaceOverrideEdges` (`src/resolution/callback-synthesizer.ts:1064`) enumerates its concrete side as `['class', 'struct']`, so a union implementor stayed out of the trait-dispatch fan-out even once it had a real node and a real `implements` edge — `Describe::describe -> Reg::describe` never formed, which is exactly the "quietly wrong rather than visibly incomplete" case you described. Added `'union'` there plus a regression test; it fails on the union assertion alone before the change, with the struct implementor as the control proving the synthesizer ran. Worth knowing: that file has a stray NUL byte at line 3135, so plain `grep` reports **no** matches anywhere in it — likely why it didn't surface when you swept for `'struct'` consumers.

**d289bf8** — merge of `main` to clear a CHANGELOG conflict (`main` had landed 15 changelog-touching commits since your branch point). Everything else auto-merged, including `src/mcp/tools.ts`, which `main` reworked substantially for the explore allocation work while this branch was adding `union` to its container sets — confirmed all the union additions survived.

Two smaller edits: rewrote the changelog entry (the original said re-indexing would "replace the earlier struct-shaped union nodes," but those never existed — unions produced no node at all, or a `type_alias` in the typedef case), and skipped the `EXTRACTION_VERSION` bump. That one isn't needed here: `main` is already at 25 against v1.5.0's 24, so existing indexes get flagged stale for the next release regardless, and over-bumping is what turns the re-index hint into noise.

Validated on the merged tree with the native kernel built: **3070 passed, 9 skipped, 0 failed**. One heads-up for future kernel work — the `kernel-*-parity` suites silently *skip* unless `npm run build:kernel` has run, and that includes the only assertion pinning `buffers.rs`'s `NODE_KINDS` to `src/types.ts`. A PR adding a node kind is exactly the one whose contract check vanishes from a default `npm test`.

Thanks again — nice work.

colbymchenry (approved):
Verified end-to-end: extraction parity across wasm and native kernel, downstream consumers updated, 3070 passed / 0 failed on the merged tree with the kernel built. Trait-dispatch gap closed in 5b0c4b8.

PR #1517, opened 2026-08-06
fix(resolution): resolve line-wrapped Python parenthesized import lists

JacquesBLR:
## Summary
`extractPythonImports` re-parses raw source text with a regex, separate from the tree-sitter AST. The regex `[^#\n]+` stopped at the first newline, so a PEP 8 line-wrapped `from pkg import (a,\n    b)` list lost every name after the statement's first physical line — not just aliased ones, the `ImportMapping` never existed at all for those names.

Found while running CodeGraph on a real project whose imports wrap this way; reproduces with as few as two names split across two lines.

## Fix
`fromImportRegex` now matches either a parenthesized list (`[\s\S]*?` up to the closing paren, so it spans newlines) or the existing single-line form, and the capture group used depends on which branch matched.

## Test plan
- [x] Added a regression test with two unaliased names split across a wrapped `from pkg import (...)` list, asserting both resolve (not just the first or last name)
- [x] Full test suite green: 2752 passed, 175 skipped, 0 failed

### DIFF
--- a/__tests__/resolution.test.ts
+++ b/__tests__/resolution.test.ts
@@ -1243,6 +1243,59 @@ def external_caller():
       expect(externalCalls).toHaveLength(0);
     });
 
+    it('resolves Python imports inside a line-wrapped `from pkg import (...)` list', async () => {
+      // extractPythonImports re-parses raw source with a regex, separately from
+      // the tree-sitter AST. That regex was `from\s+([\w.]+)\s+import\s+([^#\n]+)`
+      // — the `[^#\n]+` capture stops at the first newline. PEP 8 wraps a long
+      // parenthesized import list across multiple physical lines, so every name
+      // after line one of the statement fell outside the match and got no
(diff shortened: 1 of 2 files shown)


inth3shadows:
FYI, #1921 also covers this case. It's a broader rewrite of Python import resolution split out of #1704, and its description credits this PR.

I ran this PR's own test (`resolves Python imports inside a line-wrapped \`from pkg import (...)\` list`) as a standalone test against both:
- **Current `main` (`ba3c21e`):** fails. `etudes` gets no `calls` edge.
- **#1921 (`833fec8`):** passes. Both `rentabilite.compute()` and `echeancier.upcoming()` resolve to their module files.

#1921 reads parenthesised multi-line lists, backslash continuations and `import a, b`, and ignores docstrings and comments. That's a superset of the regex fix here. If the maintainer would rather take the smaller change first, this PR is still the more focused one.

PR #1518, opened 2026-08-06
fix(resolution): resolve direct calls through aliased Python function imports

JacquesBLR:
## Summary
Two compounding gaps in resolving `from mod import fn as alias` followed by a bare `alias()` call:

1. `resolveViaImport`'s generic "reference name matches an import" loop resolved `imp.source` through `resolveImportPath`, which for Python only maps *relative* dotted paths (`.mod`, `..pkg.mod`). An *absolute* dotted source (e.g. `scripts.lire_fec` from `from scripts.lire_fec import ...`) returned null there, and unlike `resolvePythonModuleMember` / `resolveModuleImportToFile` (#578), this loop had no `findPythonModuleFile` fallback — so it silently produced no edge for any direct (non-member) call through an absolute-module import, aliased or not.
2. Even once the file resolved, `findExportedSymbol` filters candidates on `n.isExported` — and the Python extractor never implemented `isExported`, so it was `undefined`/falsy for every Python symbol. This normally went unnoticed because an *unaliased* call (`from mod import summarize; summarize()`) still resolves via unrelated same-name fuzzy matching when the name is globally unique — but an *aliased* direct call has no name to fuzzy-match on, so this generic import path was its only route, and it always failed.

Found on a real project (`from scripts.lire_fec import summarize as fec_summary`, called as `fec_summary()`).

## Fix
- Added the same `findPythonModuleFile` fallback already used elsewhere (#578) to `resolveViaImport`'s generic loop.
- Implemented `isExported` for the Python extractor: every module- and class-level `def`/`class` is exported (Python has no export syntax — a leading underscore is only convention, and `__all__` only restricts `import *`); only names nested inside a function body (closures) are not.

## Test plan
- [x] Added a regression test for a direct call through an aliased function import from an absolute dotted module path
- [x] Full test suite green: 2752 passed, 175 skipped, 0 failed

### DIFF
--- a/__tests__/resolution.test.ts
+++ b/__tests__/resolution.test.ts
@@ -1243,6 +1243,62 @@ def external_caller():
       expect(externalCalls).toHaveLength(0);
     });
 
+    it('resolves a direct call through an aliased Python function import (`from mod import fn as alias; alias()`)', async () => {
+      // Two compounding gaps, both surfaced on a real project via a
+      // `from scripts.lire_fec import summarize as fec_summary` import,
+      // then a bare `fec_summary()` call:
+      //
+      // 1. resolveViaImport's generic "reference name matches an import"
+      //    loop resolved `imp.source` through resolveImportPath, which for
+      //    Python only maps RELATIVE dotted paths (`.mod`, `..pkg.mod`).
+      //    An ABSOLUTE dotted source (`scripts.lire_fec`, from `from
+      //    scripts.lire_fec import ...`) returned null there, and unlike
+      //    resolvePythonModuleMember / resolveModuleImportToFile (#578),
+      //    this loop had no findPythonModuleFile fallback — so it silently
+      //    produced no edge for ANY direct (non-member) call through an
+      //    absolute-module import, aliased or not.
+      //
+      // 2. Even once the file resolved, findExportedSymbol filters
+      //    candidates on `n.isExported` — and the Python extractor never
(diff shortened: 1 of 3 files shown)


danusha2345:
Took this into a local integration build of current `main`; the resolver half works, but the `src/extraction/languages/python.ts` change does not reach runtime once the native kernel is present — Python extraction then runs in `codegraph-kernel/src/python.rs`, and the PR's own test (`from scripts.lire_fec import summarize as fec_summary; fec_summary()`) fails with the kernel built and passes only with `CODEGRAPH_KERNEL=0`. The published bundles ship the kernel, so the extractor part needs mirroring in `python.rs` (and a line in `__tests__/fixtures/kernel-parity/torture.py` so parity pins it). #1517 has no such half and merged cleanly.

inth3shadows:
FYI, #1921 also covers this case. It's a broader rewrite of Python import resolution split out of #1704, and its description credits this PR.

I ran this PR's own test (`resolves a direct call through an aliased Python function import`) as a standalone test against both:
- **Current `main` (`ba3c21e`):** fails. `build_context` gets no `calls` edge.
- **#1921 (`833fec8`):** passes. `fec_summary()` resolves to `scripts/lire_fec.py:summarize`.

#1921 fixes gap 1 from your description at the root: absolute dotted specifiers now go through one resolver shared by every Python path, so there's no longer a fallback that one call site can miss. If the maintainer would rather take the smaller change first, this PR is still the more focused one.

Issue #1276, opened 2026-07-13
Go: external receiver calls resolve to unrelated local interface methods


ignazio-ingenito:
CodeGraph 1.4.1 on linux-x64 attributes calls on a concrete external receiver (database/sql.DB) to same-named methods declared by an unrelated local interface.

Minimal fixture:

    type InternalStore interface {
        Exec(string, ...any) (sql.Result, error)
        QueryRow(string, ...any) *sql.Row
    }

    type Target struct { conn *sql.DB }

    func (target *Target) Write() error {
        _, err := target.conn.Exec("insert")
        return err
    }

    func (target *Target) Read() *sql.Row {
        return target.conn.QueryRow("select")
    }

After codegraph init, codegraph callees Write --json reports Exec at the InternalStore declaration, and codegraph callees Read --json reports QueryRow at the InternalStore declaration. Target neither imports an internal implementation nor declares that it implements InternalStore.

Expected: calls resolve to database/sql.DB when supported, or remain external/unlinked. They should not create an internal dependency on InternalStore.

Actual JSON excerpts:

    {"name":"Exec","kind":"method","filePath":"flow/flow.go","startLine":7}
    {"name":"QueryRow","kind":"method","filePath":"flow/flow.go","startLine":8}

The positive fixture chain StartImport -> loadManifest -> PersistLineage is resolved correctly, so the failure is specific to generic receiver-method precision.

colbymchenry:
Fixed in #1316 (merged to main, ships in the next release), following exactly the policy you proposed: external receivers stay unlinked rather than creating internal dependencies.

Root cause: for a chained receiver (`target.conn.Exec`), extraction dropped the chain and emitted a bare `Exec`, which exact-name matching bound to the only `Exec` in the project — your unrelated `InternalStore`. Chained Go field calls now resolve through two validated inference hops (receiver's type from the method signature, field's declared type from the struct definition) and **only** that way:

- `conn *sql.DB` → external → no edge. Your fixture is a pinned regression test asserting `Write`/`Read` never appear as callers of the interface's declarations.
- In-project field types — including unexported ones — now gain *correct* edges that never existed before: on go-chi/chi this newly resolves `mx.tree.FindRoute` / `InsertRoute` / `routes` through the unexported `tree *node` field, all validated against the target type.
- Package-qualified types are followed only when the package is in-module, so `handler http.Handler` can't bind a same-named local type either.

[...]

PR #1521, opened 2026-08-06
fix(go): resolve cross-module calls in multi-module layouts

ericquan8:
A Go monorepo with several independent modules under one root (and no root
go.mod) resolved cross-module calls at ~4% recall / ~7% precision: the
single-module reader only checked projectRoot/go.mod, so every in-repo
import was classified third-party and resolution fell through to global
name-matching, wiring most cross-module calls to the wrong target. A wrong
edge is worse than a missing one for impact analysis.

This is the follow-up #388's own code comment asked for.

Multi-module import resolution:
- loadGoModules(): index EVERY go.mod under projectRoot (skip-list,
  depth/count caps, entries sorted by modulePath length DESCENDING so the
  longest prefix wins, deterministic traversal, shortest-relDir wins on a
  duplicated module path). packageDir() reproduces the old single-module
  algorithm byte-for-byte when relDir=''; it can legally return '', so
  callers test === null. loadGoModule()/GoModule untouched.
- ResolutionContext.getGoModules?() added (optional); wired as a lazy,
  memoised provider alongside getGoModule(), matching the existing
  path-aliases / workspace-packages convention. Like those, it is resolver
  metadata only — it produces no nodes and no edges.
- isExternalImport (Go branch), resolveGoCrossPackageReference and the
  name-matcher Go field-type guard prefer the index and fall back to the
  single-module path verbatim. The exact-parent match (fileDir === pkgDir)
  and the #1276 anti-fabrication guard are preserved.

Two extraction-layer defects with the same symptom (both reproduce in a
single-module repo too):
- Go const/var/method isExported was never set, so the resolver's
  `if (!node.isExported) continue` dropped them. extractMethod never called
  the hook; the Go const/var branch omitted it AND would have fed the hook
  the declaration node, which has no `name` field — the identifier is on the
  const_spec/var_spec child. Computed per-spec inside the Go branch; the
  go.ts hook and the shared declaration-level isExported are untouched.
- Grouped `var (...)` produced zero nodes. tree-sitter-go wraps a grouped
  var's specs in a `var_spec_list` node while a grouped const's specs are
  direct children, so the direct-child filter found nothing. Flattening the
  wrapper also restores calls made inside package-level grouped-var
  initializers, which previously had no source node at all.

Verification (reproducible) — 14 tests build real multi-module layouts in
temp dirs and index them for real, no mocks:

    npx vitest run __tests__/resolution.test.ts -t "Go multi-module"
    npx vitest run __tests__/extraction.test.ts -t "Go const/var extraction"

covering cross-module resolution, same-name symbol in a non-imported module
must not be picked, longest-prefix precedence, single-root-module
no-regress, no-go.mod no-op, sub-package exclusion, scan-depth cap, grouped
var/const extraction, isExported across all four const/var forms, and
TS/Python/Rust no-regress. Full suite green.

Scale check (private repo, NOT reproducible here; reported for magnitude).
11.5k Go files, 61 side-by-side modules. Ground truth is derived from the
source with no manual labelling — an import alias binds to one module path,
which maps to one local directory; a target defined exactly once in that
directory is unambiguous. Anything ambiguous is discarded, so the numbers
are conservative.

    cross-module call recall        4.03%  ->  99.67%
    cross-module target precision   6.83%  ->  100%
    file coverage                  99.94%  ->  99.94%
    call-site line precision       99.20%  ->  99.73%
    node count                    302,900  -> 311,347  (+2.8%, grouped vars)

Known gap, out of scope: cross-module references to package-level const/var
still do not resolve. Value-position `alias.Symbol` references are never
EXTRACTED as references (flushValueRefs is same-file only by design), so
they never reach the resolver regardless of isExported. The isExported fix
above is still correct, but it removes a guard nothing currently reaches.
The real fix is an extraction-layer change, tracked separately.

Design notes: docs/design/go-multi-module-resolution.md

Refs #388.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixes
+
+- In a Go project with several modules placed side by side under one directory (a common monorepo layout), a call from one module into another now resolves to the correct definition instead of being dropped or wired to the wrong symbol (#388).
+- Exported Go methods are now correctly marked as exported, so cross-package method calls resolve.
+- In a Go project, variables declared in a grouped `var (...)` block are now indexed, and exported package-level constants and variables are now correctly marked as exported.
+
 
 ## [1.5.0] - 2026-07-21
 
--- a/__tests__/extraction.test.ts
+++ b/__tests__/extraction.test.ts
@@ -347,6 +347,73 @@ in
   });
 });
 
+describe('Go const/var extraction (grouped var + isExported)', () => {
+  it('extracts every entry in a grouped var (…) block', () => {
+    // Defect B: a grouped `var ( … )` wraps its specs in a `var_spec_list`
+    // node (tree-sitter-go asymmetry vs grouped const), so they were skipped.
+    const code = `package main
+
+var (
+  A = 1
+  b = 2
+)`;
+    const result = extractFromSource('main.go', code);
+    const names = result.nodes.filter((n) => n.kind === 'variable').map((n) => n.name);
+    expect(names).toContain('A');
+    expect(names).toContain('b');
+  });
+
+  it('does not regress grouped const (…) extraction', () => {
+    const code = `package main
+
+const (
+  C = 1
+  D = 2
+)`;
+    const result = extractFromSource('main.go', code);
+    const names = result.nodes.filter((n) => n.kind === 'constant').map((n) => n.name).sort();
+    expect(names).toEqual(['C', 'D']);
+  });
+
+  it('sets isExported by leading-case for all Go var/const forms', () => {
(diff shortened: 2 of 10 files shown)


danusha2345:
Tried to take this into a local integration build of current `main` and it merged (CHANGELOG / test conflicts only), but three of its own tests then failed — `isExported` on Go methods, the same-file constant reference, the grouped-var call — and pass again only with `CODEGRAPH_KERNEL=0`. The reason: since the native kernel landed, Go extraction runs in `codegraph-kernel/src/go.rs` whenever the `.node` is present (every published bundle ships it), so a change made only in `src/extraction/tree-sitter.ts` has no effect at runtime and the two arms diverge. The resolver half (`go-module.ts`, the import-resolver / name-matcher changes) is unaffected. To land now the extraction part needs the same change in `go.rs`, plus a shape in `__tests__/fixtures/kernel-parity/torture.go` so `kernel-grammar-parity` pins it.

PR #1000, opened 2026-06-26
feat: add support for the Odin programming language

RainerXE:
This PR adds support for indexing and extracting code structure from the Odin programming language (`.odin`).

### Summary of Changes

1. **Odin Language Extractor:**
   - Added support for extracting procedures, structs, unions, bit-fields, enums, enum members, variables/constants, struct fields, and import declarations.
   - Built and integrated the `tree-sitter-odin.wasm` grammar binary.

2. **Core Struct Field Bug Fix:**
   - Modified `extractStruct` in `src/extraction/tree-sitter.ts` to leverage the custom `resolveBody` hook (similar to enums).
   - This ensures that struct fields are successfully resolved even for languages whose tree-sitter grammars do not define a named `body` field on the struct declaration nodes.

### Verification

Tested successfully against:
- Small test projects.
- The `tarn` repository.
- A large, complex Odin codebase (`RuiShin`, indexing 11,970 Odin nodes and 21,294 Odin relations in under 40 seconds).

PR #1523, opened 2026-08-06
feat(extraction): add Odin language support

ivandrenjanin:
Adds [Odin](https://odin-lang.org) support: a vendored `tree-sitter-odin` grammar, an extractor, the registration points, and tests.

## It works

I indexed three public Odin repos with this branch. A `main` build finds nothing in any of them, since `.odin` is not a known extension yet.

| repo | commit | files | nodes | edges | on `main` |
| --- | --- | --- | --- | --- | --- |
| [DanielGavin/ols](https://github.com/DanielGavin/ols) | `e62a371` | 138 | **4,640** | 10,646 | 0 / 0 |
| [greenya/SpaceLib](https://github.com/greenya/SpaceLib) | `95bf115` | 166 | **3,893** | 8,223 | 0 / 0 |
| [laytan/odin-http](https://github.com/laytan/odin-http) | `65f57ca` | 35 | **1,627** | 2,968 | 0 / 0 |

An Odin package is a directory, so a call into another package has to carry its qualifier. Every cross-directory call edge across the three resolves by qualified name, and none point at a builtin like `len` or `append`.

## About #1000

@RainerXE proposed Odin back in June in #1000 and I built on his work. He picked the grammar and worked out the node types an extractor needs. I checked those against the parser and they are right.

I did not cherry-pick his code. The extractor is new, and the overlap is boilerplate the `LanguageExtractor` interface forces. His `extractStruct` fix is a real bug that is still on `main`, and I left it for him.

He got there first and I would rather he got the credit. Take anything useful here, it is MIT. I opened separately because #1000 has conflict markers committed and its registration edits got reverted by upstream merges, so no `.odin` file would actually be opened. That is not carelessness on his part: his head branch is his fork's `main`, so every sync merges into the PR. Up to you which one you take.

## Not done

**The agent A/B in `CLAUDE.md` step 3.** It needs a logged-in `claude` CLI and tmux, and the harness targets macOS/Linux. Steps 1 and 2 ran fine. So I have no retrieval-quality number against a without-codegraph arm. I will run it if you can point me at an environment.

**CRLF checkouts lose symbols.** The grammar chokes on a trailing-backslash line continuation when the file has CRs, which is what `odinfmt` emits on a wrapped line. Roughly twice as many files fail to parse on a Windows checkout as on a POSIX one, and a few lose their declarations outright. Two smaller grammar gaps sit in the doc. I worked around none of them, and the README and CHANGELOG rows say so.

**Windows only.** I did not test macOS or Linux. The two things that break on a case-sensitive filesystem are fine: the wasm filename and the `./odin` import are both lowercase.

**Some tests already fail on Windows before this PR**, all `EPERM` from `fs.rmSync` in cleanup. Same names on `main`. My branch adds passing tests and nothing else moves.

## Grammar

| | |
| --- | --- |
| source | [tree-sitter-grammars/tree-sitter-odin](https://github.com/tree-sitter-grammars/tree-sitter-odin) |
| tag | **`v1.3.0`** |
| license | MIT, Copyright (c) 2023 Amaan Qureshi |
| sha256 | `99fd3217b82b1ba4f438120591ec2ed24108f782ab693d69b330d753188632fb` |
| built from | the tag's checked-in `src/parser.c`, no regenerate, no patch |

I pinned the tag rather than master because master's one unreleased commit breaks multi-value returns, turning `f :: proc() -> (string, bool)` into an error tree. Both binaries load cleanly, so you only catch it by diffing a corpus. Provenance and a rebuild recipe are in `docs/grammars/tree-sitter-odin.md`.

Wasm only. The kernel has no Odin, so `codegraph-kernel` is untouched.

## Testing

`tsc --noEmit` and `npm run build` are clean, and the wasm reaches `dist/` with its hash intact. The Odin tests pass. The fixture covers what those three repos do not reach on their own: calling conventions, `when` blocks, unions, bit fields, and procedure groups.

---

Twelve files in two commits. The wasm sits alone in the first one so the source diff stays readable. Nothing outside Odin, and no version bump.

Odin did surface a bug in `codegraph affected`, but the fix changes Go, Rust and Python too, so it is not this PR's to make. I can send it separately.

Refs #648, #1000.


RainerXE:
@ivandrenjanin - I'm happy if you could take this over and work on it, as t the moment i have very limited time for this. Thank you Rainer PS: My code was created and run on macosx.

Issue #1520, opened 2026-08-06
FTS retrieval quality: camelCase sub-word recall, column-weighted bm25, path filtering, dedupe


LordXyTh:
Ran codegraph (1.5.0) on a ~13k-file multi-repo workspace (TS/Python/PHP/Liquid) and measured the FTS layer against the live DB. Five findings, each with a small, sourced fix. Happy to split into separate issues if you prefer.

## 1. camelCase identifiers are unreachable by sub-word query (biggest one)

`nodes_fts` uses the default `unicode61` tokenizer. `_` is a separator (Unicode `Pc`), so snake_case works — but there's no case-boundary splitting, so `getShippingMethodIdFromCheckout` indexes as one opaque token.

Measured on my index: **53.5% of symbols are camelCase/PascalCase, and 93.4% of them can't be found by any of their constituent words** (e.g. query `checkout` reaches 6 of 76 camelCase symbols containing it; `session` 77/1,383; `product` 110/1,473).

Fix: a custom FTS5 tokenizer emitting the whole identifier as the primary token plus each sub-word with `FTS5_TOKEN_COLOCATED` at the same position — whole-identifier exact match and sub-word match from one index, prefix queries intact. Documented mechanism: https://www.sqlite.org/fts5.html#custom_tokenizers (§7.1, the synonym example transposes directly). Cheaper alternative: pre-split identifiers into a denormalized search column at insert time.

## 2. No column weighting in ranking

Default `rank` weights `name`, `qualified_name`, `docstring`, `signature` equally, so a docstring mention scores like a symbol-name hit. For code search the name should dominate — Zoekt weights symbol matches 14x over content words (`scoreSymbol 7000` vs `scoreWordMatch 500`, https://github.com/sourcegraph/zoekt/blob/main/index/contentprovider.go).

One-line fix, persisted in the table: `INSERT INTO nodes_fts(nodes_fts, rank) VALUES('rank', 'bm25(0.0, 10.0, 5.0, 1.0, 2.0)');`

## 3. `id` is in the FTS index

Opaque node IDs are never a useful query target — pure index bloat and a source of accidental matches. Suggest dropping the column from the FTS table (or `UNINDEXED`).

## 4. No way to scope a query to a path/repo

[...]

PR #1530, opened 2026-08-08
feat(search): surface camelCase symbols via segment vocab (#1520)

terrabeckhub:
## Summary

Addresses the largest finding in #1520: camelCase / PascalCase identifiers are unreachable by sub-word FTS queries because `nodes_fts` uses the default `unicode61` tokenizer (no case-boundary splits).

`name_segment_vocab` already materializes those sub-words at index time for the prompt hook. This PR merges that vocabulary into `searchNodes` candidate generation so queries like `checkout` surface `getShippingMethodIdFromCheckout` without a custom FTS5 tokenizer or schema migration.

### Changes

- After FTS, call `supplementWithSegmentMatches()` to pull symbol names from `name_segment_vocab` for each query term and join them into the candidate set (still subject to kind/language filters and the existing rescoring path).
- Add a regression test that indexes a camelCase helper and asserts `searchNodes('checkout')` finds it.

Other #1520 items (persisted bm25 rank weights, dropping `id` from FTS, `--path` CLI, GROUP BY dedupe) are left for follow-ups; search already applies column-weighted `bm25(...)` in SQL and supports `path:` query filters.

### Test plan

- [x] `npx vitest run __tests__/search-segment-supplement.test.ts`
- [ ] CI

### DIFF
--- a/__tests__/search-segment-supplement.test.ts
+++ b/__tests__/search-segment-supplement.test.ts
@@ -1,0 +1,41 @@
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import { afterEach, describe, expect, it } from 'vitest';
+import { CodeGraph } from '../src/index';
+
+/**
+ * #1520: FTS unicode61 keeps camelCase as one token. searchNodes must still
+ * surface sub-word hits via name_segment_vocab.
+ */
+describe('searchNodes camelCase segment supplement (#1520)', () => {
+  let tmpDir: string | undefined;
+
+  afterEach(() => {
+    if (tmpDir) fs.rmSync(tmpDir, { recursive: true, force: true });
+    tmpDir = undefined;
+  });
+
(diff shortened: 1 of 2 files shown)


danusha2345:
Merged this into a local integration build of current `main` and bisected a regression to it: with the segment-vocab candidates folded into `searchNodes`, three cases in `__tests__/explore-allocation-e2e.test.ts` (`CG-21 — a reservation under the file size still buys the file › the reservation is spent`: "delivers the target WHOLE rather than as a cluster stub", "spends more than the reservation", "leaves no rendered file both under its reservation and short of content") fail with `expected 'clusters' to be 'whole'`; they pass on the merge-base and again once the merge is reverted. The extra sub-word candidates change explore's candidate set enough that the reserved file is rendered as a cluster stub. Worth running that suite against the branch — the recall gain for `checkout → getShipping…` is real, but it needs to keep the allocation contract.

PR #1564, opened 2026-08-17
fix(resolution): remove catastrophic backtracking in the Vapor route regex

### LINKED ISSUE #1544: [Bug] Vapor route detector regex has catastrophic backtracking, hangs index/sync/MCP on generated Swift files
**Repo:** colbymchenry/codegraph
**Type:** Bug (performance / hang)
**Component:** `src/resolution/frameworks/swift.ts` — `vaporResolver.extract`

## Summary

The Vapor route detector uses a regex with a nested repetition group that exhibits
**catastrophic backtracking** on Swift files containing a `.get(...)` / `.post(...)` /
`.put(...)` call with a long list of comma-separated labelled arguments and **no**
`use:` label. This shape is common in generated API request code.

On such a file a single regex match takes minutes to never return, which makes
`codegraph index`, `codegraph sync`, and the MCP server (it runs a connect-time
catch-up sync) hang indefinitely. In our repo this effectively bricked the MCP
integration: every `tools/call` timed out.

## Steps to reproduce

```swift
// generated request-builder style file (one of many in our repo)
let requests: [String: VKRequest] = [:]
requests["friend_list"] = FriendsRequestFactory.get(
    count: 10 as NSNumber,
    extended: nil,
    fields: "online_info,is_friend,friend_status,is_closed,screen_name,photo_200",
    filters: "adult,no_access,blacklists",
    listId: nil,
    nameCase: nil,
    offset: nil,
    order: "smart",
    ref: "profile_friends_block",
    userId: ownerId
)
```

Run `codegraph sync` (or `codegraph index`) on a repo containing such a file, or run
the attached `repro-catastrophic-backtracking.js`:

```
kwargs=10  BROKEN:       0.3ms   FIXED:    0.1ms
kwargs=16  BROKEN:      17.5ms   FIXED:    0.1ms
kwargs=20  BROKEN:     278.7ms   FIXED:    0.0ms
# real generated .swift file:
  BROKEN: HANG (>8s)   FIXED:    0.3ms
```

Growth is exponential in the number of comma-separated arguments (see repro output).
Two real files in our repo that trigger it:

- `VKClientCourier/.../CompoundRequests/UserProfileModernCompound.swift`
- `CommunityPage/.../VKCommunityPageModel.swift`

(Both were also reported with the misleading message "Parse timed out after 30000ms"
in `.codegraph/errors.log`.)

## Root cause

```js
const routeRegex = /\b(\w+)\.(get|post|put|patch|delete|head|options)\s*\(\s*((?:[^,()]+,\s*)*)use:\s*([A-Za-z_][\w.]*)/g;
```

The captured group `(?:[^,()]+,\s*)*` is a `+` nested inside `*`. When the segment
list is long and `use:` never appears, the engine tries every possible way to split
the list at comma boundaries while scanning for `use:` — O(2^N) in the number of
segments. Worst case: `.METHOD(` followed by many labelled args where the final arg
has no trailing comma, so the group can never fully consume up to `use:`.

## Proposed fix

Replace the backtracking-prone segment group with a linear one:

```diff
--- a/src/resolution/frameworks/swift.ts
+++ b/src/resolution/frameworks/swift.ts
@@ -316,7 +316,7 @@ export const vaporResolver = {
         // args before `use:` — the label keeps only the string parts. `use:`
         // discriminates a real route from Environment.get("X")/req.parameters.get("X").
-        const routeRegex = /\b(\w+)\.(get|post|put|patch|delete|head|options)\s*\(\s*((?:[^,()]+,\s*)*)use:\s*([A-Za-z_][\w.]*)/g;
+        const routeRegex = /\b(\w+)\.(get|post|put|patch|delete|head|options)\s*\(\s*((?:[^,()]+,)*[^,()]*?)use:\s*([A-Za-z_][\w.]*)/g;
         let match;
```

- **Broken:** `(?:[^,()]+,\s*)*` — `+` inside `*`, can re-partition the whole segment
  list exponentially.
- **Fixed:** `(?:[^,()]+,)*[^,()]*?` — comma-terminated segments consumed once, plus
  a final comma-free chunk. Same accepted language, linear time.

### Semantics preserved

Verified identical results on real Vapor routes before/after:

```
app.get("users", use: UserController.index)        -> 1 match
todos.post(use: create)                            -> 1 match
app.delete("users", ":id", use: delete)            -> 1 match
router.get(use: handler)                           -> 1 match
app.get("a", "b", "c", use: foo.bar)               -> 1 match
```

The group content fed to `segJoin` is unchanged in shape, so route paths are
extracted identically.

## Impact without the fix

- `codegraph sync` / `index` never complete on repos with such generated files.
- MCP connect-time catch-up sync wedges; every tool call times out.
- Misleading "Parse timed out" errors in `.codegraph/errors.log`.

## Files

- `src/resolution/frameworks/swift.ts` — one-line regex fix (see patch).
- `repro-catastrophic-backtracking.js` — self-contained repro (Node 20+/22+;
  terminates because the broken regex is run in a child process with a hard budget).

[repro-catastrophic-backtracking.js](https://github.com/user-attachments/files/30990076/repro-catastrophic-backtracking.js)
[codegraph-vapor-regex-catastrophic-backtracking.patch](https://github.com/user-attachments/files/30990095/codegraph-vapor-regex-catastrophic-backtracking.patch)

pucedoteth:
Fixes #1544.

## The bug

`src/resolution/frameworks/swift.ts` matched Vapor routes with:

```js
/\b(\w+)\.(get|post|...)\s*\(\s*((?:[^,()]+,\s*)*)use:\s*([A-Za-z_][\w.]*)/g
```

`[^,()]+` also matches whitespace, so each space after a comma can be consumed **either** by `\s*` on the current iteration **or** by `[^,()]+` on the next one. When the call never reaches `use:` the match fails, and the engine explores every possible split.

Measured on `app.get(arg0, arg1, … x)` with no `use:` label:

| arguments | ms |
|---|---|
| 16 | 1.7 |
| 20 | 23.8 |
| 24 | 379.7 |
| 28 | 6100.1 |

~4× per additional argument. At 40 arguments it does not finish — I aborted a run after 10 minutes. Any generated Swift file with a long argument list on a `.get`/`.post`/… call stalls index/sync/MCP indefinitely.

## The fix

Keep the repeated group to `<run>,` and match the gap before `use:` once, outside the loop:

```js
((?:[^,()]+,)*)\s*use:
```

`,` is excluded from the class, so each iteration is forced to span exactly one comma-delimited run — there is only one way to split the input, and the failure is linear. **2000 arguments now match in 0.2 ms**; the 40-argument case is 0.5 ms.

## Behaviour is unchanged

The capture group is only ever consumed by `segJoin`, which pulls `"quoted"` literals out of it via `/"([^"]*)"/g`. Moving whitespace outside the group therefore cannot change a route path. I diffed old vs new across the real call shapes and the receiver/method/path/handler are identical on every one:

```
same  app.get("users", use: list)                          -> GET /users -> list
same  app.get(use: index)                                  -> GET / -> index
same  app.get("users", ":id", use: UserController.show)    -> GET /users/:id -> UserController.show
same  routes.post("a","b", use: h)                         -> POST /a/b -> h
same  app.get( "x" ,  "y" ,  use:  self.handler )          -> GET /x/y -> self.handler
same  app.get("users", BlogUser.parameter, use: show)      -> GET /users -> show
same  req.parameters.get("id")                             -> no match
same  Environment.get("X")                                 -> no match
```

The last two matter: `use:` is what discriminates a route from `Environment.get(…)` / `req.parameters.get(…)`, and both still correctly fail to match.

## Tests

Added `does not backtrack exponentially on a call that never reaches use: (#1544)` to `__tests__/frameworks.test.ts` — 40 arguments, no `use:`, asserts it completes under 1s and extracts no routes. It passes in 0.5 ms with the fix; with the old regex restored it does not terminate, so the assertion fails.

`npm run build && npx vitest run` is green: **2902 passed, 178 skipped**, up from 2901 by exactly the new test.

Note for anyone reproducing locally: the suite needs `npm run build` first — 54 tests in 15 files fail without `dist/` because the CLI tests spawn `dist/bin/codegraph.js`.

The three sibling regexes in this file (`(?:\w+\s*,\s*)*` in the SwiftUI/UIKit patterns) are **not** affected — `\w` and `\s` are disjoint, so there is no ambiguity to backtrack on. I left them alone.

### DIFF
--- a/__tests__/frameworks.test.ts
+++ b/__tests__/frameworks.test.ts
@@ -1436,6 +1436,19 @@ describe('vaporResolver.extract', () => {
     expect(references[0].referenceName).toBe('listUsers');
   });
 
+  it('does not backtrack exponentially on a call that never reaches use: (#1544)', () => {
+    // `app.get(a, b, c, ... )` with no `use:` label: the route regex must fail
+    // fast. The previous pattern had `\s*` inside the repeated group, which
+    // overlapped `[^,()]+`, so each extra argument roughly quadrupled the work
+    // and a generated file could hang indexing for minutes.
+    const args = Array.from({ length: 40 }, (_, i) => `arg${i}, `).join('');
+    const src = `app.get(${args}x)\n`;
+    const started = Date.now();
+    const { nodes } = vaporResolver.extract!('routes.swift', src);
+    expect(Date.now() - started).toBeLessThan(1000);
+    expect(nodes).toEqual([]);
+  });
+
   it('extracts grouped RouteCollection routes with the group prefix and no path arg', () => {
     const src = `
 func boot(routes: RoutesBuilder) throws {
--- a/src/resolution/frameworks/swift.ts
+++ b/src/resolution/frameworks/swift.ts
@@ -367,7 +367,14 @@ export const vaporResolver: FrameworkResolver = {
     // (`BlogUser.parameter`, `:id`, a path constant) so accept any comma-separated
     // args before `use:` — the label keeps only the string parts. `use:`
     // discriminates a real route from Environment.get("X")/req.parameters.get("X").
-    const routeRegex = /\b(\w+)\.(get|post|put|patch|delete|head|options)\s*\(\s*((?:[^,()]+,\s*)*)use:\s*([A-Za-z_][\w.]*)/g;
+    // The `\s*` that used to sit inside the repeated group overlapped
+    // `[^,()]+` (which also matches whitespace), so every space after a comma
+    // could be consumed by either branch. On a call whose argument list never
+    // reaches `use:` that ambiguity backtracks exponentially — ~4x per extra
+    // argument, seconds by 28 — hanging index/sync/MCP on generated files.
+    // Keeping the group purely `<run>,` and matching the gap before `use:`
+    // once, outside the loop, leaves exactly one way to split the input.
+    const routeRegex = /\b(\w+)\.(get|post|put|patch|delete|head|options)\s*\(\s*((?:[^,()]+,)*)\s*use:\s*([A-Za-z_][\w.]*)/g;
     let match: RegExpExecArray | null;
     while ((match = routeRegex.exec(safe)) !== null) {
       const [, receiver, method, segsStr, handlerExpr] = match;


pucedoteth:
Closing: superseded by #1547, which landed the same fix on 2026-08-22.

That PR de-ambiguates the same repetition in the same regex (`\s*` folded into the group rather than left outside it — both make the split unique and the match linear) and ships the same long-arg-list regression test in `__tests__/frameworks.test.ts`. Nothing here is left to add.

PR #1098, opened 2026-07-01
Add GDScript and Godot scene graph support

i-snyder:
Hi! I wanted to help move #364 forward since the original Godot/GDScript branch had drifted pretty far from `main`.

This PR rebases @KirisamaMarisa’s original work onto current `main`, resolves the conflicts, and keeps the original extractor design intact. I also fixed the small amount of rebase fallout needed for current CodeGraph behavior, especially the language-family gate that was otherwise dropping valid `.gd` to `.tscn` references.

I tested this with `npm install && npm run build`, the focused Godot/GDScript extraction and resolution tests, and built-CLI smoke tests on both a small multi-file Godot 4 fixture and a real Godot 4 production codebase. The testing verified that Godot files index successfully, GDScript symbols are searchable with signatures, cross-file GDScript callers resolve, `res://` references resolve, node paths resolve into `.tscn` scene nodes, and instanced scene callers show up correctly.

This is genuinely useful for Godot projects, where CodeGraph currently has no `.gd` / `.tscn` / `.tres` coverage. I’d love to see @KirisamaMarisa’s work land, and I’m happy to adjust anything that would make this easier to review.

PR #1579, opened 2026-08-21
feat(extraction): add GDScript language support

Plumvery:
## Summary

Adds **GDScript** (`.gd`, Godot Engine) as a supported language on the wasm extraction path, following the shape of previous language additions (Nix #1190, ArkTS #648).

- **Extractor** (`src/extraction/languages/gdscript.ts`, ~100 lines): functions with typed signatures, `_init` constructors, inner classes with methods, the full `var`/`const`/`@export`/`@onready` variable family, signals, enums with members, `static func` detection, and call edges through `call` / `attribute_call` / `base_call`.
- **Grammar**: [PrestonKnopp/tree-sitter-gdscript](https://github.com/PrestonKnopp/tree-sitter-gdscript) v6.1.0 (MIT), vendored as an **ABI-15 wasm rebuilt from upstream source** (`tree-sitter-cli 0.25.10 generate` + `build --wasm`, emscripten/emsdk:4.0.4), external scanner (indentation) included — the same approach as the Nix wasm.

## Grammar shapes that needed care

- The var/const family names its target via a **`name`-typed child, not `identifier`**, so the core's generic variable fallback can't read them. A `visitNode` hook creates variable/constant nodes itself, then walks initializers so calls inside them (`preload(...)`, `Foo.new()`) are captured.
- `func _init(...)` (constructor_definition) has **no name field** — `resolveName` supplies the conventional `_init`. `enumerator` names its identifier via `left`.
- `static` on functions is a **named `static_keyword` child with no field** (the var statements carry it via a field) — a child scan covers both.
- `signal foo(a, b)` extracts as a **property** carrying the parameter list, so connect()-heavy scripts expose their signal surface in the graph.
- `obj.method(args)` parses as `attribute(identifier, attribute_call(...))` — `attribute_call` joins callTypes and the core's `namedChild(0)` callee fallback yields the bare method name (name-match resolution, same as self/this receivers elsewhere).

## Validation

`scripts/add-lang/check-grammar.mjs`: **ABI 15, 20/20 clean parses, heap-safe under multi-grammar reuse** (the Lua ABI-13 corruption scenario this script exists for does not occur).

`scripts/add-lang/verify-extraction.mjs` criteria on two corpora (all critical and soft checks pass):

| Corpus | .gd files | nodes | edges | index time |
|---|---|---|---|---|
| [godotengine/godot-demo-projects](https://github.com/godotengine/godot-demo-projects) (public) | 461 | 5,216 | 6,783 | 3.7s |
| Private Godot 4 game | 138 | 2,996 | 11,484 | 1.8s |

Kind spread on the public corpus: 2,027 functions / 184 methods / 45 classes / 141 signals→properties / 39 enums + 130 members / 1,570 variables / 422 constants — symbol density well above the 1-per-file soft bar.

Tests: 4 new extraction describe-blocks + detection/support assertions in `__tests__/extraction.test.ts`; **full vitest suite passes locally** (all pre-existing tests unaffected).

## Not included (deliberately small)

- No kernel routing (stays on the wasm path, like the other recent language additions).
- No `preload`/`load` **import-edge** resolution to `res://` paths (calls are captured; resolving them to file nodes would follow the Luau instance-path `require` pattern and can come as a follow-up).
- No README language-grid SVG icon (asset pipeline is maintainer-side).

---

**Rebased onto `main` (post-1.6.0).** The only conflict was `CHANGELOG.md`: the
GDScript entry now sits under `## [Unreleased]` rather than inside the released
1.6.0 section. `src/` merged without conflicts — 1.6.0's `grammars.ts` changes
(the C++ base-clause detection) don't touch the language-registration sites.
`tsc` clean, `__tests__/extraction.test.ts` passes in full (627 tests) on the
rebased tree.


tiantian520tt:
it's wonderful. we need it, pls. thanks a lot!

wakaru44:
How does this relate to the already ongoing GDScript PR's? https://github.com/colbymchenry/codegraph/pull/1098

Plumvery:
Good question — short answer: #1098 is broader, this one is deliberately narrow.

- **#1098** (@i-snyder) — GDScript *and* Godot scene-graph (`.tscn`): 17 files, +1657.
- **#1499** (@cosmicnag) — closest in scope: 5 files, +147.
- **This PR** — extraction only: 8 files, +226. The "Not included (deliberately
  small)" section above is the scope line: no kernel routing, no `preload`/`load`
  import-edge resolution, no scene parsing.

So against #1098 it's overlap rather than duplication — scene-graph support is real
work this PR doesn't attempt, and it could sit on top of whichever extractor lands.

Two things here that are independent of which one wins: the wasm is rebuilt from
upstream grammar source as ABI 15 and clears `scripts/add-lang/check-grammar.mjs`
(20/20 parses, heap-safe under multi-grammar reuse), and the `var`/`const`/`signal`
family needs a `visitNode` hook because those name their target through a
`name`-typed child the generic variable fallback can't read.

No attachment to this one landing over the others — happy to close it if #1098 or
#1499 goes in first. @colbymchenry's call.

PR #1580, opened 2026-08-21
fix: avoid quadratic synthesis line lookups

danusha2345:
## Summary

- replace the remaining per-match `source.slice(0, index).split('\n')` line calculations in dynamic-dispatch synthesis with the existing lazy newline index
- cover ArkUI, Gin, Redux thunk, object registries, Pinia, Vuex, Celery, Spring, MediatR, Sidekiq, Erlang, and Laravel passes
- pin both synthesized edge line metadata and the absence of the quadratic allocation pattern

## Root cause and impact

Several regex-driven resolution passes recalculated a match's line by copying and splitting the entire source prefix. On a match-dense file, doing that once per match makes the final resolution pass quadratic on the main thread; reducing parse workers cannot affect it.

The existing `makeLineAt` helper builds one newline-offset array lazily per scanned source and resolves each match with a binary search. This keeps edge attribution unchanged while removing repeated prefix allocations.

This addresses one concrete main-thread amplifier relevant to the full-resolution OOM investigation in #1553. It does not claim that every reported 4–6.5 GB case has the same sole cause.

## Validation

- `npm run build:kernel`
- `npm run build`
- `npx vitest run --reporter=dot` — 185 files passed; 3120 tests passed, 9 skipped
- affected synthesizer suites — 12 files, 28 tests passed
- controlled full-index A/B on a 20,000-match TypeScript fixture (three alternating runs): median 3.81s → 0.75s; median peak RSS 398,356 KB → 376,848 KB
- semantic graph hashes are identical before and after for nodes, edges, and all 20,001 unresolved references

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixes
 
+- Dynamic-dispatch analysis no longer repeatedly copies every source prefix while scanning match-dense files, avoiding quadratic work and excessive peak memory during the final resolution pass.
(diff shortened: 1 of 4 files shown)


danusha2345:
Folded this exact commit into #1583 as 180c6ea so the quadratic resolution-pass work and the related memory fixes can be reviewed and tested together. Closing this draft as superseded; no code was dropped.

PR #1587, opened 2026-08-22
fix(resolution): Rust field receivers resolve by field type, not by name

### LINKED ISSUE #1585: Rust: a method call on a typed field resolves to a same-named method on the wrong type
In Rust, a method call whose receiver is a typed field can resolve to a method of the same name on an unrelated type — including the calling method itself, producing a self-recursive edge that does not exist in the source. It also happens when the receiver's type is not a local one at all (a std container, an external crate, a generic parameter): the call is attached to a same-named local method.

The same structure in C++ resolves correctly (`Outer::run -> Inner::run`), so this looks specific to Rust.

### Minimal reproduction

```toml
# Cargo.toml
[package]
name = "repro"
version = "0.1.0"
edition = "2021"
```

```rust
// src/lib.rs
pub mod inner;
pub mod outer;
```

```rust
// src/inner.rs
pub struct Inner {
    pub n: usize,
}

impl Inner {
    pub fn run(&mut self) {
        self.n += 1;
    }
}
```

```rust
// src/outer.rs
use crate::inner::Inner;

pub struct Outer {
    pub inner: Inner,
}

impl Outer {
    pub fn run(&mut self) {
        self.inner.run();
    }
}
```

```
codegraph init && codegraph index .
```

```sql
SELECT s.qualified_name || ' -> ' || t.qualified_name, e.line
FROM edges e JOIN nodes s ON e.source = s.id JOIN nodes t ON e.target = t.id
WHERE e.kind = 'calls';
```

Actual:

```
Outer::run  ->  Outer::run     line 9
```

`Outer::run` does not call itself. `self.inner` has its type declared on the struct, so `Outer::run -> Inner::run` is what the source says. Resolving it correctly would be ideal; leaving the ref in `unresolved_refs` would at least be safe — as it stands `unresolved_refs` holds no entry for this call, so a consumer has no way to tell the edge was a guess.

### Impact

The shape is common in idiomatic Rust: a wrapper or delegating struct holds a field of another type and forwards a method of the same name, or the field is a std container / external crate type while a local method happens to share its name.

Three real examples from ripgrep — in all three the receiver is not `self`, yet each is recorded as a self-recursive edge:

```rust
// crates/ignore/src/walk.rs:824      ig_builder: IgnoreBuilder
self.ig_builder.add_custom_ignore_filename(file_name);

// crates/ignore/src/walk.rs:1195     its: std::vec::IntoIter<...>
match self.its.next() {

// crates/globset/src/lib.rs:983      matcher: Regex
self.matcher.is_match(candidate.path.as_bytes())
```

There are a few hundred such non-existent self-edges in ripgrep and more in tokio (`SELECT count(*) FROM edges WHERE kind = 'calls' AND source = target`, then checking the receiver at each call site). Self-edges are only the subset that is easy to falsify — a wrong pick that lands on some other method cannot be spotted this way, so the affected set is larger.

A few self-edges are genuine recursion (e.g. ripgrep's `Error::is_partial` unwrapping a nested error), and are not counted above.

### Environment

Reproduced both on the v1.5.0 release (npm `@colbymchenry/codegraph@1.5.0`) and on a local build of current `main` (`81e1f4a`). macOS arm64, Node v24.7.0.

Dshuishui:
Fixes #1585.

A Rust method call on a field — `self.inner.run()` — resolves to whatever same-named method sits nearest, because the receiver is dropped and the call becomes a bare `run`. Rust has no implicit `self`, so every call on a field is written this way, and it is worst when the field's type is external (`Vec`, `Arc<…>`, another crate) — there the nearest project method is never the right answer.

This follows the shape of the Go field-chain fix (#1276): the receiver is kept, the field's declared type is read from the enclosing type's own declaration, and the method is validated on it. **A field whose type cannot be established stays unresolved rather than being guessed.**

Field types: `Box`/`Rc`/`Arc` resolve through to what they wrap; `Option`/`Mutex`/`Vec` keep their own methods; a generic type answers with its constructor; `dyn Trait` resolves to the trait; tuple structs index by position.

Verified by indexing ripgrep and tokio: self-recursive edges the source never had are roughly halved, and several hundred more edges move from the wrong type to the right one. Sampling the call sites that lost their edge and checking them by hand, the large majority were wrong before.

Deeper chains (`self.a.b.method()`), receivers rooted at a local, and method chains keep their existing behavior — each was implemented and measured, and each removed fewer wrong edges than it cost correct ones.

Coverage is also limited by generic impls qualifying their methods with the trait rather than the type (`impl<T> Stream for Empty<T>` → `Stream::poll_next`). That is a separate bug, filed as #1588.

Other languages are unaffected — indexing TypeScript, Python and Erlang corpora before and after produces byte-identical nodes, edges and unresolved refs, with no change in indexing time.

The suite passes; the three new tests fail on `main` and pass here.


Dshuishui:
Closing in favour of #1599 — it covers this case and goes further than this PR did: the native kernel path (`codegraph-kernel/src/rustlang.rs`) and its parity fixture, which this PR left untouched, so byte-parity would not have held here.

Thanks for the quick turnaround on both reports.

Issue #1567, opened 2026-08-18
codegraph init walks into node_modules correctly excluded by a NESTED (non-root) .gitignore — ~63x file blowup (509 -> 32,260 files) on v1.5.0


milos85vasic:
## Summary

On `codegraph init . --force` (v1.5.0), a re-index of an existing, previously-clean project walked into two subdirectories whose `node_modules` trees are correctly excluded for `git` itself by **nested** (non-root) `.gitignore` files — producing a ~63x file-count blowup (509 files -> 32,260 files, 8,906 nodes -> 514,456 nodes, 724,013 edges, 1.8 GB DB) before the run was manually aborted (host-resource discipline; the run was still growing).

This is a real-world regression report from an active project (`boba`, a torrent-search proxy stack), not a synthetic construction. I (an independent contributor, not affiliated with the project that found it) verified the root-cause evidence first-hand and spent significant effort trying to reproduce it in isolation — full details below, including a negative result that may itself help triage.

## Root cause evidence (git side — confirmed first-hand)

```
$ git check-ignore -v frontend/node_modules
frontend/.gitignore:10:/node_modules	frontend/node_modules

$ git check-ignore -v extension/node_modules
extension/.gitignore:2:node_modules/	extension/node_modules

$ git ls-files frontend/node_modules extension/node_modules | wc -l
0
```

Both directories are correctly and unambiguously excluded for `git` by their own **nested** `.gitignore` (not the project's root `.gitignore`, which does not mention `extension/node_modules` at all — only `frontend/node_modules/` is separately listed there as a belt-and-suspenders entry). Yet `codegraph init . --force` on v1.5.0 walked into both anyway (365 MB + 236 MB, tens of thousands of files).

The project's own `codegraph` version had moved from `0.9.9` (last validated 2026-06-06, 509 files / 8,906 nodes, `7 PASS / 0 FAIL`) to `1.5.0` with no re-verification in between — `frontend/` and `extension/` (each with their own `node_modules` and nested `.gitignore`) were both added to the project **after** the 0.9.9 baseline, so this exclusion path had never been exercised against the newer CodeGraph version before this incident.

## What I could NOT reproduce (important negative result)

I read `src/extraction/index.ts` on `main` (current HEAD, still `1.5.0` per `package.json`) and traced both scanning implementations:

[...]

PR #1604, opened 2026-08-23
fix(extraction): keep the git-aware scan working on git older than 2.36 (#1549)

### LINKED ISSUE #1549: includeIgnored silently bypassed on older git (<2.34): git ls-files -s --recurse-submodules fails, falls back to filesystem walk that ignores the opt-in
### Summary
On systems with older git (verified failing on git 2.25.1 / Ubuntu 20.04; working on git 2.34.1+ / Ubuntu 22.04/24.04/26.04), codegraph index silently drops files that should be pulled in via codegraph.json 's includeIgnored . The user gets no error — just a smaller index. Same project, same codegraph.json , same codegraph version (1.5.0) indexes correctly on newer git.

### Environment
- codegraph: @colbymchenry/codegraph 1.5.0
- OS: Ubuntu 20.04 LTS
- git: 2.25.1 (fails) — works on 2.34.1+ (Ubuntu 22.04/24.04/26.04)
- Node: 20.x
- engines in package.json only constrains Node ( >=20.0.0 <25.0.0 ); no minimum git version is declared.
### Reproduction
Minimal project ( t1/ ):

On Ubuntu 20.04 / git 2.25.1:

On Ubuntu 22.04 / git 2.34.1 (same repo, same codegraph 1.5.0):

### Root cause
Both call sites run git ls-files -z -s --recurse-submodules . On older git, --recurse-submodules only supports -c/--cached , not -s/--stage :
 Call site 1 — collectGitFiles() (the indexing path, fatal )
No try/catch here → execFileSync throws (exit 128) → exception escapes → getGitVisibleFiles() catches it and return null → scanDirectory() falls back to scanDirectoryWalk() (the non-git filesystem walk). scanDirectoryWalk() only parses .gitignore and does not implement includeIgnored — that opt-in lives exclusively in the git path ( findIgnoredEmbeddedRepos() , called from collectGitFiles() ), which never runs. dir_b is gitignored → skipped → only a.c is indexed.

Net effect: on older git the entire includeIgnored mechanism is silently disabled, and because the failure is swallowed into a "non-git project" fallback, the user gets no diagnostic — just missing files.
 Call site 2 — discoverEmbeddedRepoRoots() (the watcher scope, non-fatal but inconsistent )
This one is already wrapped in try/catch, so on older git the gitlink discovery pass is silently skipped (unexpanded gitlinks → not found). Non-fatal, but it means the watcher scope diverges from the indexer scope on older git: the indexer (once site 1 is fixed) recurses into gitlinks via collectGitFiles() , while the watcher's discoverEmbeddedRepoRoots() misses them.

### Impact
Any project relying on includeIgnored to index embedded/gitignored repos is silently mis-indexed on older git. The super-repo-of-clones layout (the very use case includeIgnored exists for, per issues #622/#699/#514) is affected — and in particular embedded repos tracked as gitlinks (mode 160000, without .gitmodules ) are missed by the watcher on old git even after site 1 is fixed. No warning or error is surfaced in either case.

### Suggested fix
Wrap both call sites in a try/catch and fall back to git ls-files -z -s (drop --recurse-submodules ) when the combined form is unsupported.

Site 1 — collectGitFiles() :

Site 2 — discoverEmbeddedRepoRoots() :
 Why the -s fallback is safe
- -s keeps the <mode> <object> <stage>\t<path> format, so the existing 160000 gitlink detection still works.
- Without --recurse-submodules , active submodules surface as 160000 gitlinks and are recursed into by the existing gitlink pass at the end of collectGitFiles() ( classifyGitDir(...) === 'embedded' covers submodule checkouts). So active submodules are still indexed.
- Embedded independent repos (the dir_b case — untracked, not a gitlink) are still discovered by the untracked trailing-slash pass and findIgnoredEmbeddedRepos() , so includeIgnored works again. Verification
Applied to the packaged lib/dist/extraction/index.js on Ubuntu 20.04 / git 2.25.1:

```
$ codegraph index .
◆  Indexed 2 files        # a.c + dir_b/b.c  ✓
$ codegraph files
├── dir_b
│   └── b.c (c, 2 symbols)
└── a.c (c, 2 symbols)
```
Behavior on newer git is unchanged (the try branch still runs the original command).

### Alternatives considered
- Declare a minimum git version (e.g. ≥2.34) and error out clearly on older git. This avoids the silent-misindex UX but excludes Ubuntu 20.04 LTS users entirely; the graceful fallback above is strictly better for compatibility.
- Fall back to git ls-files -z -c --recurse-submodules (which old git supports). Rejected: -c omits mode bits, breaking 160000 gitlink detection, so the gitlink pass would stop finding unexpanded submodules. The -s fallback keeps mode bits.

maxmilian:
Fixes #1549.

`git ls-files -s --recurse-submodules` is rejected outright before git 2.36. `builtin/ls-files.c` listed `show_stage` among the modes that trigger `die("ls-files --recurse-submodules unsupported mode")`, and it was only dropped in 2.36:

| tag | guard |
|---|---|
| v2.34.1 (`:720`) | `recurse_submodules && (show_stage \|\| show_deleted \|\| …)` → dies |
| v2.35.0 (`:728`) | `recurse_submodules && (show_stage \|\| show_deleted \|\| …)` → dies |
| **v2.36.0 (`:728`)** | `show_stage` removed → passes |

The `die` is unconditional — it never checks whether the repo actually has submodules — so on those versions *every* call fails. That is Ubuntu 22.04 LTS (git 2.34.1) and Debian 11 (2.30.2), not just the 20.04 the report mentions.

## Why it cost more than submodule expansion

The call at `collectGitFiles` had no `try`/`catch`, so the throw escaped to `getGitVisibleFiles`' `catch { return null }`, and `scanDirectory` fell through to `scanDirectoryWalk`. That walk has no notion of `includeIgnored` — all 39 references sit above the fallback — and the same `try` also covers gitlink recursion and `collectIncludedFilesForRoot`, so the `codegraph.json` `include` allowlist went with it.

The net effect on those distros: CodeGraph indexes a different file set than the configuration asks for, with no error, no warning, and no symptom other than files quietly missing from query results.

## The fix

Retry as `git ls-files -z -s`, dropping `--recurse-submodules` rather than `-s`. Keeping `-s` is the important half — gitlink detection reads the mode bits to spot `160000` entries — and embedded repos are reached through that gitlink recursion anyway, so the fallback loses very little on old git and nothing at all on new.

Both call sites now go through one helper. The second one (`discoverEmbeddedRepoRoots`) already had a `try`/`catch` so it degraded less noisily, but it was silently losing the same expansion.

## Testing

`__tests__/extraction.test.ts` gains a regression test that installs a PATH shim: a `git` that exits 128 with the real `fatal: ls-files --recurse-submodules unsupported mode` whenever it sees `-s` together with `--recurse-submodules`, and delegates to the real binary otherwise. That makes this reproducible on any git version rather than only on a machine old enough to fail naturally.

Fixture is the one from the issue — root repo, `dir_b/` in `.gitignore`, `codegraph.json` with `{"includeIgnored": ["dir_b/"]}`, `dir_b` carrying its own `.git`:

```
before, under the shim →  ["a.ts"]                     # dir_b/b.ts silently dropped
after,  under the shim →  ["a.ts", "dir_b/b.ts"]
```

The test asserts the real-git baseline first, so it fails loudly if the shim ever stops shimming rather than passing vacuously.

`__tests__/extraction.test.ts` is 613/613. The full suite is unchanged against `main` — same 57 pre-existing environment failures on both, with the one extra pass being this test.

## Credit

@newshowardz777 wrote the root-cause analysis and the patch shape in the issue; the commit carries a `Co-authored-by` trailer for them. I only widened the version boundary and added the shim-based coverage.

One non-claim, since it looks tempting from the outside: #1567 (nested `.gitignore`, `node_modules` walked into) has the shape of this same fallback, but `scanDirectoryWalk` does apply `buildDefaultIgnore`, which covers `node_modules` — so I can't call it the same root cause and haven't touched it here.

Worth noting #1568 edits this exact `catch` block to add a `logDebug`. It doesn't change behaviour, but it would have made this failure visible, so the two are complementary — happy to rebase around whichever lands first.


colbymchenry:
Thank you @maxmilian for the patch, the accurate Git <2.36 boundary, and the regression test, and @newshowardz777 for the root-cause analysis.

Your commit `27c149524d968485164fb43bdec994fdb9323c75` is now carried forward in Forge PR #1793: https://github.com/colbymchenry/codegraph/pull/1793. It uses your `lsFilesStaged()` helper at both call sites and keeps `__tests__/extraction-old-git.test.ts` byte-for-byte unchanged. The landing preserves current main's extraction/watcher changes, resolves the changelog conflict, and credits you both as co-authors on the Colby-authored commit `3e9ccb24bf3e4310566214acfab0b0b0d266941a`.

Linux verification reproduced the missing `dir_b/b.ts` against current main using your old-Git PATH-shim approach, then confirmed the same fixture passes with the fix. Tracked-gitlink watcher discovery also changes from `[]` to `["lib/"]`. All 93 selected tests and `npx tsc --noEmit` passed.

Closing #1604 as superseded by #1793, which remains open awaiting merge.

Issue #769, opened 2026-06-10
feat: add MCP tool to list all initialized CodeGraph projects


woan-lufei:
## Problem

Currently, CodeGraph has no way to discover which projects are already initialized/indexed. Each project stores its `.codegraph/` directory independently, and all MCP tools require the caller to already know the exact project path via the `projectPath` parameter.

For AI agents working across multiple repositories, this means every cross-repo query requires the agent to be told every project path — it cannot discover what is available.

## Current capability

All MCP tools (`codegraph_explore`, `codegraph_search`, `codegraph_callers`, `codegraph_files`, `codegraph_status` etc.) accept an optional `projectPath` string parameter to query a different project. But there is no tool to enumerate which projects exist.

## Proposed solution

Add a new MCP tool (e.g. `codegraph_projects`) that returns a list of all initialized CodeGraph projects, with basic metadata per project:

- **Project path** — the root directory
- **Index timestamp** — when it was last indexed
- **Stats** — file/node/edge counts
- **Status** — whether the index is up-to-date with the working tree

## Design considerations

1. **Discovery mechanism** — how to find projects? Options:
   a. A registry file (e.g. `~/.config/codegraph/projects.json`) that `codegraph init` writes to and `codegraph uninit` removes from
   b. Scan well-known directories
   c. Both: registry as primary, with a `CODEGRAPH_PROJECTS_PATH` env var for additional roots

2. **Minimum viable** — start with the registry approach: `codegraph init` registers the project, and the new tool reads the registry

3. **MCP-only or also CLI?** — the tool should work both as `codegraph_projects` via MCP and `codegraph projects` via CLI

/cc @woan-lufei

PR #1605, opened 2026-08-23
fix(cli): report unsupported-language projects instead of finishing silently (#1502)

### LINKED ISSUE #1502: init silently reports success on an unsupported-language project (0 files, index_state: complete)
If a project is entirely in a language codegraph doesn't support, `codegraph init` scans it, finds nothing it can parse, and reports success anyway: `index_state: complete`, 0 files, no warning. From an MCP client (or a human) that looks identical to "indexed fine, your query just didn't match anything", since `codegraph_explore` comes back empty either way.

Repro:
```
mkdir demo && cd demo
printf 'module demo::vault { public fun f() {} }\n' > a.move
codegraph init          # "Indexed 0 files", exits 0, index_state: complete
codegraph explore "f"   # empty, no hint that .move isn't supported
```
(Any unsupported language does it, Move's just my case.)

Why it stings: an agent wired up over MCP burns a few calls concluding "no matches" before it works out, usually by poking at the SQLite db, that codegraph never parsed the files at all. There's no way to tell "codegraph is inactive for this tree" apart from "codegraph is active and found nothing."

Would be great if a 0-files-but-non-empty-workspace init surfaced that, maybe a one-liner on init and in `codegraph_explore`:
> No supported source files were indexed (42 files present, no matching grammar). codegraph is inactive for this workspace.

Listing the extensions it saw but skipped would be a nice bonus. Nothing needs to change on the happy path.

codegraph 1.5.0, linux-x64.

maxmilian:
Fixes #1502.

A project written in a language CodeGraph has no grammar for is currently indistinguishable from an empty one. Unsupported extensions are filtered out during discovery, so `filesDiscovered` is 0; the reconciliation in `src/index.ts:704-724` computes `discovered - accounted = 0`, finds no shortfall, and records `index_state: complete`; and the CLI prints the same `No files found to index` it prints for a genuinely empty repo. Exit 0 throughout.

That reconciliation already carries the right intent in its own comment — *"don't let the index pass as complete"* — it just can't see this case, because its denominator is files it recognised.

The silence is worst over MCP, where an empty result is byte-identical to "no match" and the agent has been told to trust the graph rather than grep. @netbrah's follow-up is the sharp version: 24k Perl files, every one classified `unknown`, and nothing anywhere saying so.

## What changed

The scan already walks every file, so the tally of what it declined to index rides along on the walk it was doing anyway — no second pass, no extra I/O. That matters at 24k files.

`init`/`index` now ends with:

```
▲  No supported source files found — 3 file(s) present, none in a language CodeGraph indexes: .move (2), .pl (1)
●  CodeGraph is inactive for this workspace — searches will return nothing. Use your own file tools here.
```

instead of `No files found to index`.

## What I deliberately did not change

**`index_state` keeps its existing values.** Adding an `inactive` or `empty` state would change the `getIndexState()` union and the `status --json` contract, and would ripple into the CLI's warn branches. That's a call for you to make, not something to slip into a bug fix — so this reports the condition without redefining the state.

**MCP is left for a follow-up.** `src/mcp/tools.ts:3354` returns the same `No relevant code found for "<query>"` string whether the index is empty-because-unsupported or simply has no match, and `src/mcp/session.ts:251` picks instructions on whether `.codegraph/` exists, so a 0-file index still receives the full "trust codegraph, don't grep" playbook. The vocabulary for this already exists — `SERVER_INSTRUCTIONS_NO_ROOT_INDEX` from #769 says almost exactly the right thing — so the fix is likely to point that judgement at "indexed but empty" as well as "not indexed". I'd rather land the CLI half and hear how you want the MCP half worded than guess at instruction text, since that file is the SSOT for it.

## Testing

Three cases in `__tests__/extraction.test.ts`, covering both discovery paths, because they filter in different places: the git path (`git ls-files` then extension filter) and the filesystem-walk fallback. The third asserts the tally stays empty when everything was indexable, so the counter can't silently fire on healthy projects.

Verified end to end on a real run rather than only in unit tests — the block above is the actual output of `codegraph init` in a directory holding `a.move`, `b.move`, `c.pl`.

Full suite is 3011 passed / 0 failed on Node 22.

One small correction to the report, since it's the kind of thing that wastes a reviewer's time: the CLI was not entirely silent — it did print `No files found to index`. The substantive part of the complaint stands, though: that message is the empty-repo message, `index_state` still said `complete`, the exit code was still 0, and MCP had nothing at all.


colbymchenry:
Superseded by Forge land #1806 (cherry-pick of this PR's tip `2c208927` → `aff436be` on current main, CHANGELOG conflict only). Closing in favor of #1806. Thanks @maxmilian and @netbrah!

colbymchenry:
Superseded by #1806.

PR #1615, opened 2026-08-25
fix(erlang): give same-name different-arity functions separate arity-qualified nodes (#1610)

### LINKED ISSUE #1358: Erlang: erlangArityAt doesn't skip <<binary>> literals (contrary to its docstring), so binary-arg dispatch sites get an inflated arity and the behaviour-callback edge is dropped or mis-linked — verified
## Summary

`erlangArityAt` (the arity counter for the Erlang behaviour-dispatch synthesizer) does **not** skip `<<binary>>` literals, despite its docstring claiming it does. Commas inside a binary literal are counted as top-level argument separators, so a dispatch site like `Mod:decode(<<1,2,3>>, Opts)` is computed as arity **4** instead of **2**. The behaviour-dispatch lookup then keys on the wrong `name/arity`, so the edge to the real `-callback` is silently dropped (missed edge) — or collides with an unrelated behaviour that happens to declare that inflated arity (wrong edge). Verified against the real compiled function.

## Root cause

```ts
// src/resolution/callback-synthesizer.ts (erlangArityAt)
/** ... Skips nested (), [], {}, <<>> content, "strings", 'atoms', and $c char literals ... */
function erlangArityAt(src: string, openIdx: number): number {
  let depth = 1; let commas = 0; ...
  for (let i = openIdx + 1; i < limit; i++) {
    const ch = src[i]!;
    if (ch === '"' || ch === "'") { /* skip string/atom */ }
    if (ch === '$') { /* skip char literal */ }
    if (ch === '(' || ch === '[' || ch === '{') { depth++; ... }
    if (ch === ')' || ch === ']' || ch === '}') { depth--; ... }
    if (ch === ',' && depth === 1) { commas++; continue; }   // <- binary's inner commas land here
    ...
  }
}
```

The scan special-cases `()`/`[]`/`{}` but never `<<`/`>>`, so the `,` inside `<<1,2,3>>` is seen at `depth === 1` and counted. The docstring's own worked example (`init(fun((a, b) -> ok), #{k => v})` → 2) passes only because `()` and `{}` balance; the `<<>>` claim in the same sentence is unimplemented.

## Repro (executed against the real compiled function)

| dispatch site | real arity | `erlangArityAt` |
|---|---|---|
| `Mod:decode(<<1,2,3>>, Opts)` | 2 | **4** |
| `Mod:encode(<<"pre", Rest/binary>>, Ctx)` | 2 | **3** |
| `Mod:handle(Msg, State)` (control, no binary) | 2 | 2 |
| `init(fun((a, b) -> ok), #{k => v})` (docstring example) | 2 | 2 |

Binary literals with commas are ordinary Erlang (`<<1,2,3>>`, `<<"prefix", Rest/binary>>`), so any `Var:fn(<<...>>, ...)` dispatch site through a behaviour hits this.

## Suggested direction

Track `<<`/`>>` depth alongside the existing bracket depth: on a two-char `<<` increment `depth` and advance past both chars, on `>>` decrement (guarding the two-char match so a lone `<`/`>` comparison isn't misread). Then the docstring's `<<>>` claim holds and the arity is correct.

## Environment

`main` @ `5955d04`, package `1.4.1`, Node 22.23.1 (Linux). Function extracted from the built `dist/` and exercised directly.

### LINKED ISSUE #1610: Erlang: same-name different-arity functions collapse into one node
### Problem

In Erlang, arity is part of a function's identity: `f/1` and `f/2` are two independent top-level definitions that can be exported separately and are unrelated at the language level. The canonical spelling is `module:function/arity` — `-export`, `fun M:F/A`, stack traces and the docs all carry it.

The comment on `handleFunDecl` explains the merge is deliberate ("deliberately grouped under one node, the way overloads are elsewhere"), and the reasoning makes sense to us — on C++/Java the analogy holds, overloads really are versions of one function. Erlang differs, though: `f/1` and `f/2` are separate definitions, either one can be exported alone, and they may do entirely unrelated things.

Merging clauses that share a name **and** arity into one node is correct — semantically they are one function. The issue is that the merge condition compares only the name, so the outcome also depends on whether the two definitions happen to be adjacent:

```erlang
f(X)    -> X + 1.
f(X, Y) -> X + Y.
```
```
function  gap::f    span 4-5          <- one node holding two functions
```

```erlang
f(X)    -> X + 1.
g()     -> ok.
f(X, Y) -> X + Y.
```
```
function  gap::f    span 4-4
function  gap::g    span 6-6
function  gap::f    span 8-8          <- two nodes with identical qualified_name
```

Neither shape distinguishes arity.

### Impact

`cowboy_req.erl`:

```erlang
420  header(Name, Req) ->
421      header(Name, Req, undefined).
422
423  -spec header(binary(), req(), Default) -> binary() | Default when Default::any().
424  header(Name, #{headers := Headers}, Default) ->
425      maps:get(Name, Headers, Default).
```

**Inside a single node:**

```
signature:  -spec header(binary(), req()) -> binary() | undefined.     <- the spec for header/2
span:       420 - 425                                                  <- covers both /2 and /3
```

`signature` comes from the first definition while `span` covers both functions; the `-spec` for `/3` on line 423 falls inside the span but is not used. Fetching the source for this node returns two functions plus a type declaration — not the thing `signature` describes.

**Call edges:**

```
header -> header          line 421      <- the real header/2 -> header/3 call becomes a self-loop
parse_cookies -> header                 <- no way to tell whether /2 or /3 was called
filter_cookies -> header
```

Line 421 is a genuine function call.

### Scope

Counting `-export([name/arity, ...])` across cowboy, ranch, rebar3, jsx and fast_xml, same-name different-arity accounts for roughly 8–17% of the exported API.

Erlang has neither default arguments nor loop constructs: the former means a "with a default" API has to be written as `f/N` delegating to `f/N+1`; the latter makes "public `f/1` calling private `f/3` with an accumulator" the standard way to iterate. In the cowboy sample about 86% of the same-name groups have a real call between their members — so what this mostly affects is not unrelated functions that happen to share a name, but pairs with a genuine call edge that turns into a self-loop.

### Elsewhere

With no arity on the node, other layers re-derive it or drop it:

- the behaviour dispatch in `resolution/callback-synthesizer.ts` needs to look up by `name/arity`; since the node doesn't carry it, `erlangArityAt` recounts arguments from source text (#1358 is a bug in that text scan)
- the explore query normalization in `mcp/tools.ts` strips `/3` off `mod:fn/3` before searching; the comment there records a measurement on cowboy where an agent asked for `cowboy_stream_h:request_process/3` twice, got nothing back, and fell back to Read

### Environment

Reproduced both on the v1.5.0 release (npm `@colbymchenry/codegraph@1.5.0`) and on a local build of current `main` (`44e1812`). macOS arm64, Node v24.7.0.

colbymchenry:
Fixes #1610. Also fixes #1358 (the `<<binary>>` arity miscount in behaviour dispatch, reported separately and hit by the same code path).

## Problem

Arity is part of an Erlang function's identity — `f/1` and `f/2` are unrelated top-level definitions — but the extractor merged consecutive same-name `fun_decl`s regardless of arity. Reproduced on main exactly as reported:

- adjacent `f(X) -> …. f(X, Y) -> ….` → **one** node spanning both, with the first definition's signature;
- interleaved `f/1, g/0, f/2` → two nodes with **identical** `qualified_name`;
- `cowboy_req`'s `header(Name, Req) -> header(Name, Req, undefined).` → a **self-loop** `header → header`, with the `-spec` for `/3` swallowed by the merged span;
- `-export([f/1])` marked every arity exported.

## Fix

- **One node per (name, arity).** Clauses of the same name+arity still merge (that part of the old behavior was correct); a different arity starts a new node. `qualifiedName` carries the canonical spelling — `mod::f/1` — while the node **name stays bare** so search and bare-name matching are unchanged.
- **`-export` and `-spec` are per-arity.** `-export([f/1])` exports exactly `f/1`; a spec sitting between two arities attaches to the arity its signature names.
- **Refs carry the call-site arity** wherever it's statically known: local `f/1`, remote `mod::f/2`, `fun f/1` / `fun mod:f/1` values, `gen_server` dispatch (`handle_call/3`, `handle_cast/2`), and spawn/apply MFA lists (`spawn_link(?MODULE, work, [A, B])` → `work/2`).
- **The matcher resolves only to the named arity** — same file first (a local call targets its own module) — and when no definition of that arity exists it resolves to **nothing** rather than a sibling arity: silent beats wrong. An arity-less dynamic-MFA ref resolves only when the module defines exactly one arity of that name.
- **Behaviour dispatch** selects the implementer node of the site's arity, and the arity counter now skips `<<1,2,3>>` binary-literal commas per its own docstring (#1358) — `Mod:decode(<<1,2,3>>, Opts)` counts 2, not 4.
- **`codegraph_explore` / `codegraph_node`** accept the written `mod:fn/3` spelling against the new arity-qualified names (the issue's measured `cowboy_stream_h:request_process/3` shape).

## Validation

Minimal fixtures (all three reported shapes) now index as `gap::f/1` + `gap::f/2`, distinct `inter::f/1`/`inter::f/2`, and a real `deleg::header/2 → deleg::header/3` edge with no self-loop.

Cowboy (fresh `--depth 1` clone, this build vs unmodified main build):

| | main | this PR |
|---|---|---|
| nodes | 3,668 | 3,748 (+80 — the arity splits; no explosion) |
| erlang function nodes | 2,850 | 2,930 |
| behaviour dispatch edges | 38 | **44** |
| `cowboy_req::header` | one node, span 420–425, /3's spec lost | `header/2` (420–421, its own spec) + `header/3` (424–425, its spec) |
| delegation | self-loop | `header/2 → header/3` |

`calls` edges drop 6,059 → 5,656: a sample of every removed pair shows the false-positive class the issue predicted — out-of-repo/BIF calls (`length/1`, `error/1`, `quicer:*`) that previously name-matched onto unrelated same-named in-repo functions now stay unresolved.

Tests: new arity coverage in extraction + a new arity-resolution integration suite + a #1358 binary-literal behaviour test; updated existing Erlang expectations to the arity-carrying spellings. Full suite: **3,018 passed, 0 failed**.

No migration: an existing Erlang index picks the new shape up on its next re-index (`codegraph sync` / re-`init`).

Erlang is wasm-only (not in the native kernel), so there is no kernel-parity surface.

https://claude.ai/code/session_01LxZj6W6Y1SHXwvpT3uwJpK


danusha2345:
I reviewed and reproduced two remaining correctness gaps in this patch:

1. A bare Erlang `f/N` with no same-module definition still falls back to a project-wide candidate. In a minimal repro, the BIF call `length/1` incorrectly resolved to an unrelated in-project `other::length/1`.
2. `-import(Module, [f/N])` is not represented in `extractImportMappings`, so when multiple modules define the same `f/N`, path proximity can select the wrong module. My repro with `-import(imported, [pick/1])` resolved to a nearby `wrong::pick/1`.

The node/edge shape also needs an `EXTRACTION_VERSION` bump; otherwise existing indexes are not marked stale and unchanged Erlang files are skipped by `codegraph sync`.

I pushed a focused follow-up commit that fixes all three points and adds permanent end-to-end regressions:

- commit: https://github.com/danusha2345/codegraph/commit/2d54da884185bcb61571f0104684b3944a4ac256
- branch: https://github.com/danusha2345/codegraph/tree/fix/1610-erlang-import-arity
- cherry-pick: `git cherry-pick 2d54da884185bcb61571f0104684b3944a4ac256`

Validation after the final change: `npm run build:kernel`, `npm run build`, and the full suite — 190 test files, 3,190 passed, 9 skipped. The local index was then force-rebuilt and reports extraction version 26 with zero pending changes/refs.

Dshuishui:
Verified this branch end-to-end — the three shapes are fixed, and the `calls` edges that disappear on cowboy are all genuine false positives (`quicer:*` is out of repo, `error/1` and `length/1` are BIFs).

One small thing: arity is counted with `namedChildCount`, and tree-sitter counts `comment` as a named child, so a comment at the top level of an argument list inflates it:

```erlang
f(A, % why
  B)        % read as f/3 — the edge is silently dropped
```

Only rebar3 is affected in the projects I checked (2 real edges). I put a patch and a regression test for it in #1619, targeting this branch.

PR #1619, opened 2026-08-26
fix(erlang): don't count comments as arguments when computing arity

Dshuishui:
Follow-up to #1615, rebased onto `main` now that it has landed.

Arity is computed with `namedChildCount`, but tree-sitter reports `comment` as a named child, so a comment at the top level of a parameter or argument list inflates it:

```erlang
plain( % a comment
    1, 2)          % counted as plain/3 — the edge is dropped

withdef(X, % a comment
        Y) -> ...  % indexed as withdef/3
```

Same for the MFA list in `spawn(?MODULE, f, [A, % note\n B])`. A comment nested one level deeper (inside `#state{...}`, a tuple, a list) is unaffected — only the top level of the list counts.

This surfaced with #1615: before it, arity took no part in resolution, so the miscount was harmless. Now arity is identity, so one extra unit silently drops the edge.

Small blast radius. Across cowboy, ranch, jsx, rebar3 and fast_xml only rebar3 is affected, losing two real edges (`rebar_compiler.erl:403`, `rebar_release_SUITE.erl:247`); re-indexing those five with this patch changes nothing else — no nodes added or removed, no edges removed, exactly those two edges restored.

The three counts now go through one `namedArgCount` helper so the invariant is stated once.

The new test in `erlang-arity-resolution.test.ts` covers the four shapes above. I checked it is not vacuous: on `main` without this patch it fails, and it passes with it. The full suite passes (3,053).

No CHANGELOG entry — #1615's entry already describes the user-visible behaviour, and the trigger here is rare; this just keeps that rare case from silently dropping edges.

### DIFF
--- a/__tests__/erlang-arity-resolution.test.ts
+++ b/__tests__/erlang-arity-resolution.test.ts
@@ -120,6 +120,39 @@ go(Args) ->
     expect(edges.some((e) => e.sq === 'spawner::go/1' && e.tq.startsWith('multi::job'))).toBe(false);
   });
 
+  it('counts arity past comments written inside a parameter or argument list', async () => {
+    fs.mkdirSync(path.join(dir, 'src'), { recursive: true });
+    fs.writeFileSync(
+      path.join(dir, 'src', 'noted.erl'),
+      `-module(noted).
+-export([run/0, plain/2, spaced/2]).
+
+plain(X, Y) -> X + Y.
+
+spaced(X, % why this one is special
+       Y) ->
+    X * Y.
+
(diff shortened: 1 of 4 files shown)


Dshuishui:
Rebased onto `main` now that #1615 has merged, and retargeted this PR from the (now-merged) topic branch to `main`. Same single commit, unchanged.

Re-verified on current `main`: the miscount is still there (all three `namedChildCount` sites), the new test fails without this patch and passes with it, and the full suite is green.

PR #1621, opened 2026-08-26
fix(mcp): keep explore's already-sent tracking per calling context

### LINKED ISSUE #1620: codegraph_explore tells a fresh subagent that source was "already sent" to a different context
### Describe the bug

`codegraph_explore`'s cross-call dedup replaces already-served source with an "Already sent earlier in this conversation … Use it from your context; do NOT Read this file" pointer. The already-sent record is kept **per MCP connection**, but one connection is not always one agent context:

1. **Subagents (Claude Code):** subagents dispatch tool calls over the parent's MCP connection. A fresh subagent exploring a file the parent already received gets the pointer instead of the source — wrong on both counts: the source is not in its context, and it's told not to Read the file. There is no in-tool recovery; re-querying hits the same suppression, so the subagent falls back to Read/grep (or worse, answers without the source).
2. **Compaction:** after the host compacts the conversation, the context no longer holds the already-sent source, but the server's record still does — subsequent explores return pointers to text that is gone until the connection restarts.

### Reproduction

1. In Claude Code, run any `codegraph_explore` that returns a file's source.
2. Spawn a subagent (Task tool) and have it explore the same symbols.
3. The subagent's very first explore returns `> **Already sent earlier in this conversation:** … Do NOT Read this file.`

Compaction variant: explore, `/compact`, explore the same symbols again — same pointer, source no longer in context.

### Why it matters

The pointer is designed to be trusted ("that copy is still exact"). When it points at source the current context never had, the agent either reads the file anyway (defeating dedup and teaching it to distrust codegraph) or proceeds without source it believes it has.

Codex is unaffected on the subagent side (each thread gets its own MCP server process), but the compaction case applies to every host whose connection outlives a compact.

AtkinsChang:
Fixes #1620

### Problem

The explore already-sent record (the source of "Already sent earlier in this conversation" back-references) is scoped to the MCP connection. Harnesses like Claude Code route a subagent's tool calls over the same connection as the main agent, and a compaction empties the context while the connection lives on — in both cases the agent is handed pointers to source its current context never held.

### What this does (4 commits)

1. **`fix(mcp)` — per-context bucketing.** The session record is bucketed by an optional, host-injected `sessionId` on `codegraph_explore`. No id → the shared default bucket, byte-for-byte today's behavior (goldens unchanged). Ids are opaque Map keys (never logged/echoed/pathed); over-long ids are sha256-hashed rather than truncated so distinct ids can never collide into one bucket. The schema doesn't advertise the parameter by default; `CODEGRAPH_MCP_EXPLORE_SESSION_PARAM=1` declares it for hosts that validate args (hook-injected fields reach the server either way — verified on Claude Code 2.1.246).
2. **`feat(mcp)` — the reset path.** A one-shot control line on the daemon socket (`{"codegraph_control":1,"op":"clear-session-record","sessionId":…}`) clears one caller's buckets across live sessions; it can never clear the shared default bucket. Two always-exit-0 hook entry points drive it: `codegraph hooks pre-tool-use` stamps `sessionId` (`<session_id>` for the main agent, `<session_id>:<agent_id>` for a subagent) onto explore calls via the host's PreToolUse `updatedInput`; `codegraph hooks post-compact` reads the compact hook payload and sends the control line. `--agent codex` additionally pairs `permissionDecision:"allow"` with the rewrite — codex requires the pair (and treats the decision as approval-inert) while Claude Code would act on it, so it is emitted only where the wiring explicitly declares codex.
3. **`feat(installer)` — Claude Code wiring.** `codegraph install` adds two hooks to Claude settings: `PreToolUse` (matcher `mcp__.*__codegraph_explore`) → `hooks pre-tool-use --agent claude`, and `PostCompact` → `hooks post-compact`. Surgical merge, idempotent, uninstall reverses; user hooks untouched.
4. **`feat(installer)` — Codex wiring.** Same two entries into `~/.codex/hooks.json` (global) / `./.codex/hooks.json` (local), with the same `PostCompact` event (codex fires it only on successful compaction, and every failure path in all four compaction implementations returns before the history rewrite commits — a failed compaction leaves the agent's context intact, so skipping the reset there is correct). Entries are appended, never reordered — codex's hook-trust key includes the group index, so reordering would silently un-trust existing user hooks. `[hooks.state]` is never written: codex asks the user to review & trust the new entries on next run.

### Compatibility

- Hosts with no hooks wired: identical behavior (default bucket, schema unchanged).
- A malformed/absent id can never clear or pollute the shared bucket; eviction and hashing both fail toward re-serving.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New Features
 
-- Two new hook entry points let a host agent tell CodeGraph which context is calling: `codegraph hooks pre-tool-use` tags each `codegraph_explore` call with the agent that made it, and `codegraph hooks post-compact` clears that context's already-sent record after a compaction. `codegraph install` wires them into Claude Code automatically, so already-sent tracking follows each subagent and each compacted session instead of the connection they share; `codegraph uninstall` removes them, and hooks you added yourself are left alone either way. Other agents are unaffected — nothing changes for hosts that don't run them.
+- Two new hook entry points let a host agent tell CodeGraph which context is calling: `codegraph hooks pre-tool-use` tags each `codegraph_explore` call with the agent that made it, and `codegraph hooks post-compact` clears that context's already-sent record after a compaction. `codegraph install` wires them into Claude Code and Codex automatically — on Codex they are recorded as trusted as they are written, so it doesn't stop to ask you to review them — and already-sent tracking then follows each subagent and each compacted session instead of the connection they share. `codegraph uninstall` removes them, and hooks you added yourself are left alone either way. Other agents are unaffected — nothing changes for hosts that don't run them.
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### New Features
+
+- Two new hook entry points let a host agent tell CodeGraph which context is calling: `codegraph hooks pre-tool-use` tags each `codegraph_explore` call with the agent that made it, and `codegraph hooks post-compact` clears that context's already-sent record after a compaction. Wire them into your agent's hook settings and already-sent tracking follows each subagent and each compacted session instead of the connection they share. Nothing changes for hosts that don't run them.
+
 ### Fixes
 
 - `codegraph_explore` no longer tells a subagent that source was "already sent" when it went to a different agent. Harnesses like Claude Code route a subagent's tool calls over the same connection to CodeGraph as the main agent, so a fresh subagent could be handed a pointer to code its own context never received — and it would go read the file instead. Already-sent tracking is now kept per calling context.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New Features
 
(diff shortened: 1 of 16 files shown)


AtkinsChang:
@colbymchenry  Before I polish this further: are you okay with the general approach here?


Since an MCP connection can span multiple contexts (main/subagents, and across compaction), this PR:

* buckets already-sent state by host session_id / agent_id
* uses the host compact hook to reset that bucket

So CodeGraph relies on the harness to define context lifecycle rather than trying to infer it.

Is this ownership model acceptable, or would you prefer a different approach?

bompus:

Merges onto main with a CHANGELOG-only conflict. No index change (identical node/edge counts), as expected.

PR test files: 347 pass, 2 fail, both in `installer-targets.test.ts` (`codex: records ONLY our two entries`, `codex: re-trusting an upgraded launcher keeps the user's enabled flag`), and both pass on main. Cause is Windows paths in the test, not the code: the writer emits `hooks.state.${JSON.stringify(key)}`, which correctly escapes backslashes for TOML (`"C:\\Users\\..."`), while the test builds its expected header and its `replace()` target from the raw path with single backslashes, so `toContain` misses and the `enabled = false` line is never inserted. Escaping the key the same way in the test should fix both.

Not exercised here: the live Claude Code subagent path (needs the session hooks installed into our user settings, which I did not want to do on a verification build). #1624 verified on the same tree: 26/26.

colbymchenry:
Superseded by Forge landing PR https://github.com/colbymchenry/codegraph/pull/1788, which fixes #1620 via the simpler #1624 opt-in default (re-serve unless `CODEGRAPH_EXPLORE_DEDUP=1`).

This draft's per-context `sessionId` buckets + PreToolUse/PostCompact hooks are a strong longer-term path to keep dedup *on* safely when hosts inject lifecycle, but it had no clean merge-base with current `main` and wasn't needed to correct the subagent/compaction failure mode. Closing as superseded — thanks @AtkinsChang.

Issue #1532, opened 2026-08-09
[Bug] v1.5.0 requires FTS5 which is missing in official Node.js builds


xsdxq-null:
> This issue was generated by an AI assistant (Codex/OpenAI).

## Description

codegraph v1.5.0 uses Node.js built-in node:sqlite and requires FTS5, but official Node.js binaries do **not** compile FTS5 by default. This causes codegraph init and codegraph index to fail on any standard Node.js installation.

## Environment

- OS: Windows 11 (also affects all platforms)
- Node.js: v22.14.0 (official build from nodejs.org)
- codegraph: v1.5.0
- SQLite version: 3.47.2 (no ENABLE_FTS5 in compile_options)

## Error

[ERR] Failed to index: no such module: fts5

Direct reproduction:`javascript
const { DatabaseSync } = require('node:sqlite');
const db = new DatabaseSync(':memory:');
db.exec('CREATE VIRTUAL TABLE t USING fts5(content)');
// Error: no such module: fts5
``n
## Root Cause

Node.js node:sqlite module uses SQLite compiled WITHOUT FTS5 extension. codegraph v1.5.0 relies on FTS5 for full-text search (CREATE VIRTUAL TABLE ... USING fts5).

## Suggestions

1. Detect FTS5 at runtime, fallback to LIKE-based or FTS4 search
2. Document FTS5 as a requirement
3. Make FTS5 optional for basic indexing

## Workaround

npm install -g @colbymchenry/codegraph@1.0.1

colbymchenry:
## Fixed on Linux (Forge)

Landed upstream #1625 (@aniruddhaadak80) onto current main as Forge PR #1810, with a small schema-boundary follow-up so required post-FTS tables still create when FTS5 is missing.

**PR:** https://github.com/colbymchenry/codegraph/pull/1810
**Commits:** 4c9b9452 (cherry-pick of upstream) + cf4f5650 (boundary + tests + CHANGELOG)

### Linux verify (Node v22.19.0)
- This box has real FTS5; missing-FTS5 path simulated by intercepting FTS virtual-table creation.
- Before: schema exec fails with no such module: fts5.
- After: init/index succeed with warning; search uses LIKE + fuzzy; reopen + bulk helpers no-op safely.
- Present path: real FTS5 search + bulk rebuild still green.
- Tests: 34/34 focused (7 new fts5-fallback tests); build PASS.

Closing as completed. Upstream #1625 will be closed as superseded by #1810.

PR #1625, opened 2026-08-26
fix(db): graceful FTS5 fallback when Node.js build lacks FTS5 support (#1532)

aniruddhaadak80:
## What Problem This Solves

codegraph v1.5.0 uses Node.js built-in `node:sqlite` and requires FTS5, but official Node.js binaries do **not** compile FTS5 by default. This causes `codegraph init` and `codegraph index` to fail on any standard Node.js installation with `Error: no such module: fts5`.

## Why This Change Was Made

The schema at `src/db/schema.sql:117` creates an FTS5 virtual table unconditionally. When FTS5 is not available in the SQLite build bundled with Node.js, the entire schema execution fails, preventing codegraph from functioning at all.

## User Impact

codegraph now works on Node.js builds without FTS5. When FTS5 is unavailable:
- Schema creation succeeds (FTS5 parts are skipped with a console warning)
- Search falls back to LIKE + fuzzy matching (already implemented in `searchNodes`)
- Bulk-load operations skip FTS5 trigger management
- Full-text search performance is reduced but functional

## Evidence

- Issue: https://github.com/colbymchenry/codegraph/issues/1532
- Root cause: `node:sqlite` uses SQLite compiled without FTS5
- Fix: split schema execution, detect FTS5 at init and open, gate FTS5 operations on `fts5Available` flag
- Files changed: `src/db/index.ts` (schema split + flag), `src/db/queries.ts` (search fallback)
- 2 files changed, 60 insertions, 7 deletions

### DIFF
--- a/src/db/index.ts
+++ b/src/db/index.ts
@@ -83,10 +83,17 @@ export class DatabaseConnection {
    */
   private openedInode: string | null;
 
-  private constructor(db: SqliteDatabase, dbPath: string, backend: SqliteBackend) {
+  /**
+   * Whether FTS5 is available in this Node.js build. When false, search
+   * falls back to LIKE + fuzzy matching (#1532).
+   */
+  readonly fts5Available: boolean;
+
+  private constructor(db: SqliteDatabase, dbPath: string, backend: SqliteBackend, fts5Available: boolean) {
     this.db = db;
     this.dbPath = dbPath;
     this.backend = backend;
+    this.fts5Available = fts5Available;
     this.openedInode = statInode(dbPath);
   }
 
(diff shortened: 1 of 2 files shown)


colbymchenry:
Superseded by Forge PR https://github.com/colbymchenry/codegraph/pull/1810 which cherry-picks this fix onto current main, bounds the optional FTS schema section so post-FTS required tables still create, and adds Linux regression coverage for missing and present FTS5.

Thanks @aniruddhaadak80 — credited in CHANGELOG and preserved as Author on the cherry-pick commit.

PR #1632, opened 2026-08-28
fix(resolution): keep the existence probe inside the project root (#1631)

### LINKED ISSUE #1631: Resolution: fileExists probes paths outside the project root (uncontained path.join fallback)
`fileExists` in `src/resolution/index.ts` falls back to the filesystem without containing the path:

```js
fileExists: (filePath: string) => {
  if (this.knownFiles) { /* … index hit, O(1) … */ }
  // Fall back to filesystem for files not yet indexed
  const fullPath = path.join(this.projectRoot, filePath);
  try { return fs.existsSync(fullPath); } catch { … }
}
```

`resolveRelativeImport` computes `path.relative(projectRoot, basePath)`, which can carry `../` segments, and `path.join` does not clamp — `path.join('/root', '../../etc/passwd')` is `/etc/passwd`. So a relative import in an indexed file probes existence outside the project root.

## Reproduction

Nine lines, no OpenSCAD involved — this affects every language with relative imports:

```bash
mkdir -p esc/proj/src esc/outside
echo 'export function outsideSecret() { return 42; }' > esc/outside/secret.js
printf "import { outsideSecret } from '../../../outside/secret.js';\nexport function useIt() { return outsideSecret(); }\n" > esc/proj/src/a.js
cd esc/proj && git init -q . && git add -A && git -c user.email=t@t -c user.name=t commit -qm init
strace -f -e trace=newfstatat,stat,lstat,access -o /tmp/tr.txt codegraph init -i .
grep -c "outside/secret.js" /tmp/tr.txt
```

Result: **9 `access()` calls** against paths two levels above the project root.

```
access(".../esc/outside/secret.js.js",   F_OK) = -1 ENOENT
access(".../esc/outside/secret.js.jsx",  F_OK) = -1 ENOENT
access(".../esc/outside/secret.js.mjs",  F_OK) = -1 ENOENT
…
```

Only `access` — no `openat` for that path, so nothing outside is read.

## Severity, stated plainly

Low, and I would rather understate it than dress it up:

- **No read.** The content-serving sinks are guarded separately by the `validatePathWithinRoot` work in #527, so this is not the file-disclosure path that fixed.
- **No wrong edge.** The target is not in the index, so resolution declines and no edge is produced.

What remains is an **existence oracle driven by repository content**: a `.scad`/`.js`/`.py` file in an indexed repo can make the indexer probe whether an arbitrary absolute path exists, and timing or subsequent behaviour could in principle expose that. It is also just surprising — every other layer contains paths, and this one quietly does not.

## Suggested fix

The `knownFiles` set is consulted *first*, so any indexed file resolves regardless of where it lives. Only the filesystem fallback needs bounding, and a path outside the root can never be an indexed project file — so returning `false` there is the correct answer rather than a new restriction:

```js
const fullPath = path.join(this.projectRoot, filePath);
if (!validatePathWithinRoot(this.projectRoot, fullPath, { allowSymlinkEscape: true })) {
  return false;
}
```

`allowSymlinkEscape` matches the indexing tier established in #935, so an in-root symlink to a vendored library outside the repo keeps working; only the lexical `../` escape is refused.

I have deliberately **not** included this in #1630 (OpenSCAD support). Fixing it there would give one language a guarantee the other forty lack, and the change belongs to whoever owns this code path rather than to a language PR.

Found while validating OpenSCAD support; confirmed on `6a056ec` (v1.6.0).

maxmilian:
Fixes #1631.

`fileExists` fell back to `path.join(this.projectRoot, filePath)`, which does not clamp, while relative-import resolution hands it paths carrying `../` segments. A crafted relative import in an indexed file therefore made the resolver `stat` arbitrary absolute paths. Nothing outside the root is read (the content sinks are guarded separately by `validatePathWithinRoot`) and no edge is produced, but the probe is an existence oracle driven by repository content — and every other layer contains paths, so this one quietly not doing it is surprising on its own.

The `knownFiles` set is consulted first, so any indexed file resolves regardless of where it lives; only the filesystem fallback needs bounding. A path outside the root can never be an indexed project file, so returning `false` is the correct answer rather than a new restriction.

## Why lexical containment, and not `validatePathWithinRoot`

The issue suggests `validatePathWithinRoot(root, p, { allowSymlinkEscape: true })`. That is right about the intent but wrong about the cost here: `fileExists` runs per import candidate, and `validatePathWithinRoot` does two `realpathSync` calls per invocation. Measured over 20,000 calls:

| | existing file | missing file |
|---|---|---|
| before (unguarded) | 1.54 µs/call | 1.38 µs/call |
| `validatePathWithinRoot` | **105.77 µs/call** | **153.24 µs/call** |
| lexical only (this PR) | 1.86 µs/call | 1.91 µs/call |

The missing-file column is worse because the realpath pair goes through the `ENOENT` throw path, and building the exception dominates.

It is also the wrong check semantically. The symlink half exists to stop an in-root symlink whose real target escapes the root (#527) — but indexing deliberately follows exactly those symlinks (#935), which is why `allowSymlinkEscape` exists in the first place. On this path only the lexical `../` escape should be refused.

So this splits the lexical half out as `lexicalPathWithinRoot` and has `validatePathWithinRoot` call it, rather than duplicating the containment rule. Its doc comment says plainly that it is not a substitute for the full check on any path whose contents get served.

## Tests

`__tests__/resolution-fileexists-containment.test.ts`, three cases:

- a path escaping the root is refused **even though the target really exists** (the test asserts the target's existence first, so `false` can only come from the guard) — this is the one that fails before the change
- files inside the root still resolve, present and absent alike
- **an in-root symlink pointing outside the root still resolves** — green before and after, guarding the #935 behaviour against being closed off by this change

Full suite: 179 files / 3055 tests passing (3052 on `main` plus these three). `npm run build` clean.

Reported and diagnosed by @ErQrYfkrju, who deliberately kept it out of #1630 so it wouldn't land as a single-language guarantee.


colbymchenry (approved):
Rebased onto current `main` (includes #1749). Verified on Linux:

- fail→pass: with the lexical guard removed, `refuses to probe a path that escapes the root` fails (`fileExists` returns true for `../outside/secret.js`); with the guard restored, all 3 tests in `__tests__/resolution-fileexists-containment.test.ts` pass.
- Related path/symlink filter tests green.

Also fixed a small JSDoc ordering issue so `validatePathWithinRoot`'s docs stay attached to that function after `lexicalPathWithinRoot` was split out.

Looks good to merge — thanks @maxmilian.

Issue #1531, opened 2026-08-09
Codex CLI now supports project-local config (.codex/config.toml) — CodexTarget.supportsLocation('local') is stale


shortFisherman:
## Summary

`src/installer/targets/codex.ts` still claims Codex CLI has no project-local config concept, but OpenAI Codex has since added a first-class **project config layer** (`.codex/config.toml` in the repo root). This makes the `global only` label stale and prevents per-project installs for Codex users.

## Evidence

The `supportsLocation('local') => false` logic traces back to this comment in `codex.ts`:

> "Codex CLI **as of 2026-05** has no project-local config concept — everything lives under `~/.codex/`."

That is no longer accurate:

1. **Official layering model** — `codex-rs/config/src/loader/README.md` in [openai/codex](https://github.com/openai/codex) documents the config layer stack, where **`Project` config (`.codex/config.toml`) is layer 4**, above the user config layer (`~/.codex/config.toml`, layer 6). Layers merge recursively top-over-bottom.

2. **`mcp_servers` is read from the merged effective config** — `codex-rs/core/src/mcp.rs` builds `McpConfig` via `runtime_config(&Config)` where `Config` is the merged layer stack output, and `configured_mcp_servers` / `effective_mcp_servers` consume that catalog. So a `[mcp_servers.codegraph]` table in a project's `.codex/config.toml` is honored, project-scoped.

3. **Verified on Codex CLI 0.147.0** (2026-08): an MCP server declared in `<project>/.codex/config.toml` shows up in `codex mcp list` inside that project and nowhere else.

## Suggested change

- `CodexTarget.supportsLocation('local')` → return `true` (optionally version-gated: only for Codex builds that ship the project layer, with a graceful fallback to `global` for older ones).
- `local` install should write `[mcp_servers.codegraph]` to `<project>/.codex/config.toml` and the marker-fenced CodeGraph section to the project-root `AGENTS.md` (mirroring the opencode target's local behavior — Codex already reads project-root `AGENTS.md`).
- `uninstall --location=local` should roll those back, and `detect(loc)` should check `<project>/.codex/config.toml` for an existing `[mcp_servers.codegraph]` table.

## Why it matters

CodeGraph's whole install story is "wire up only the agents you use, per project when possible". Codex is currently the one agent that forces a global install, which leaks the MCP server into every Codex session on the machine — exactly what the `local` location exists to avoid.

PR #1029, opened 2026-06-27
installer: honor CLAUDE_CONFIG_DIR for global Claude Code installs

borfast:
## Problem

The Claude Code installer target hardcodes `~/.claude` and `~/.claude.json`, so `codegraph install --location global` always writes to the **default** profile — even when the user runs Claude Code against a custom profile via the [`CLAUDE_CONFIG_DIR`](https://docs.claude.com/en/docs/claude-code/settings) env var (e.g. `~/.claude-gc`).

The result: the MCP server entry, permissions, and `CLAUDE.md` instructions all land in a directory Claude Code never reads, so codegraph silently never loads for users on a non-default profile.

## Fix

Resolve the global config dir from `CLAUDE_CONFIG_DIR` when set, falling back to `~/.claude`. The `.claude.json` MCP file follows Claude Code's own placement rule:

- **default profile** → `~/.claude.json` (in `$HOME`, *beside* `~/.claude`)
- **`CLAUDE_CONFIG_DIR` set** → `$CLAUDE_CONFIG_DIR/.claude.json` (*inside* the profile dir)

`settings.json` and `CLAUDE.md` move with the profile dir. Project-local installs are unaffected.

## Tests

- Added a test asserting that with `CLAUDE_CONFIG_DIR` set, a global install writes `.claude.json`, `settings.json`, and `CLAUDE.md` into the custom profile and leaves `~/.claude` untouched. Verified it fails on `main` and passes with this change.
- Hardened the `installer-targets` harness to clear `CLAUDE_CONFIG_DIR` in `setHome`, so a custom profile in the real environment can't leak into the existing default-profile tests.

All 152 tests in `installer-targets.test.ts` pass; `tsc --noEmit` is clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1633, opened 2026-08-28
fix(installer): honor CODEX_HOME for the Codex global install (#1627)

### LINKED ISSUE #1627: codegraph install ignores CLAUDE_CONFIG_DIR and CODEX_HOME environment variables
## Description
When running codegraph install, the installation process ignores custom path configuration set via the CLAUDE_CONFIG_DIR and CODEX_HOME environment variables. The tool consistently falls back to installing and writing files directly to the user's home directory (~/).

## Steps to Reproduce

- Set custom environment variables:

```bash
export CLAUDE_CONFIG_DIR="/custom/path/to/claude_config"
export CODEX_HOME="/custom/path/to/codex_home"
```

- Run the installation command:

```bash
codegraph install
```

- Check the target directories.

## Expected Behavior
codegraph install should read and respect CLAUDE_CONFIG_DIR and CODEX_HOME to determine where to place configuration files and runtime artifacts.

## Actual Behavior
The installation hardcodes or defaults to ~/ (e.g., ~/.claude or ~/.codex), completely ignoring the exported environment variables.

## Environment
OS: Linux

maxmilian:
Addresses the `CODEX_HOME` half of #1627.

Codex resolves its user layer from `CODEX_HOME` and only falls back to `~/.codex`. The Codex target hardcoded the fallback, so a user running Codex against a custom profile got a *correct* install into a directory Codex never reads — the `[mcp_servers.codegraph]` entry, the `AGENTS.md` block, and `detect()` all pointed at the default profile. Nothing errors; CodeGraph simply never loads.

Resolution now mirrors what the copilot-cli target already does for `COPILOT_HOME`:

```ts
const override = process.env.CODEX_HOME;
if (override && override.trim().length > 0) return override;
return path.join(os.homedir(), '.codex');
```

Only the user layer moves. The project layer from #1531 stays anchored to the project — `.codex/config.toml` beside the repo is not a profile artifact — and there is a test pinning that.

## Scope: this is deliberately half of #1627

The issue names two variables. The `CLAUDE_CONFIG_DIR` half is already handled by **#1029** (open since June, rebased onto current `main` on Aug 25, waiting on review) — it was opened before this issue existed, so the two aren't linked in either direction. This PR does not touch `targets/claude.ts`, so the two don't collide. Landing both closes #1627.

## Tests

Four cases in a new `describe`, following the shape of the existing `opencode XDG config path (#535)` block:

- global install writes to `$CODEX_HOME` and **nothing lands in `~/.codex`** — fails before the change
- `detect()` and `uninstall()` follow `$CODEX_HOME` too — fails before the change
- unset **or blank** `CODEX_HOME` still falls back to `~/.codex` — green before and after, guarding the fallback
- local install is unaffected by `CODEX_HOME` (#1531) — green before and after

One extra fix in the harness: `setHome()` clears `HERMES_HOME` and `COPILOT_HOME` but not `CODEX_HOME`. Now that the target reads it, the existing codex tests would fail on any machine that has the variable exported, so it is cleared and restored alongside the other two.

Full suite: 178 files / 3056 tests passing (3052 on `main` plus these four). `npm run build` clean.

Reported by @seanchann.


colbymchenry:
Superseded by Forge land https://github.com/colbymchenry/codegraph/pull/1783 — your `CODEX_HOME` commit was cherry-picked intact (author preserved), then extended with the `CLAUDE_CONFIG_DIR` half from #1029 so #1627 is fully covered. Closing this PR as superseded; thanks @maxmilian!

PR #1634, opened 2026-08-28
fix(extraction): preload the Objective-C grammar for C-family headers (#1628)

### LINKED ISSUE #1628: Objective-C .h files fail to parse unless a .m file preloads the grammar
## Summary

An Objective-C header can fail with `Failed to get parser for language: objc` when a project contains no `.m`/`.mm` file.

CodeGraph initially classifies `.h` files as C and preloads the C and C++ grammars. After the worker pool has been initialized, content-aware detection correctly reclassifies the header as Objective-C, but the worker has no Objective-C grammar loaded.

This was originally observed in OpenCV's `opencv2/videoio/cap_ios.h`, but the reproducer below has no dependencies or imports.

## Environment

- CodeGraph 1.6.0
- Linux 5.15 x86_64

## Minimal reproduction

Create an otherwise empty directory containing exactly this file:

### `repro.h`

```objc
@interface CGRepro
- (void)run;
@end
```

Run:

```console
$ codegraph init --yes
```

Actual result:

```text
Indexing failed — all 1 files had errors

1 parser initialization failures
```

`.codegraph/errors.log`:

```text
1 files with errors

repro.h: Failed to get parser for language: objc
```

`codegraph status --json` reports `fileCount: 1`, `nodeCount: 0`, `languages: ["objc"]`, and `state: "failed"`.

The same failure reproduces with a subsequent `codegraph index`.

## Expected result

`repro.h` should be parsed as Objective-C and its class/method should be indexed, without requiring a sibling `.m` or `.mm` file.

## Control experiment

Adding this second file causes both files to index successfully:

### `grammar_seed.m`

```objc
void codegraph_objc_grammar_seed(void) {}
```

With that file present, `codegraph index` reports:

```text
Indexed 2 files
4 nodes, 2 edges
```

Removing `grammar_seed.m` and rebuilding restores the parser initialization failure.

## Diagnosis

The failure is deterministic:

```text
path-only detection:       repro.h -> c
preloaded grammars:        c, cpp
content-aware detection:  repro.h -> objc
getParser("objc"):        null
```

Loading the Objective-C grammar explicitly changes `getParser("objc")` from null to a valid parser, so the shipped `tree-sitter-objc.wasm` is present and loadable.

The preload logic currently derives `neededLanguages` without file contents and adds `cpp` when `c` is present, but not `objc`:

https://github.com/colbymchenry/codegraph/blob/main/src/extraction/index.ts#L1665-L1669

Later, parsing performs content-aware detection and sends the resulting `objc` language to the already initialized worker.

## Possible fixes

- When `.h`/C is present, preload both `cpp` and `objc`; or
- lazily load the final content-detected grammar in the worker.

Either approach should preserve the current content-aware `.h` detection while ensuring the selected parser is available.

maxmilian:
Fixes #1628.

A `.h` file is classified from its path alone as C, so the preload set handed to the parse workers contains `c` — plus `cpp`, added explicitly because a header might turn out to be C++. Parse-time detection then reads the source and can also return `objc` (`detectLanguage`'s `.h` branch checks `looksLikeCpp` *and* `looksLikeObjc`), and no worker holds that grammar. The file fails outright with `Failed to get parser for language: objc` and nothing in it reaches the index.

That is why the failure looked arbitrary: any `.m` file in the project pulls `objc` into the set by path, so the very same header indexes fine. @Juddd's control experiment — adding a one-line `grammar_seed.m` and watching both files index — pins it exactly.

So `objc` is preloaded alongside `cpp` whenever `c` is present: both are readings that content-aware `.h` detection can produce, and both have to be available before the workers start.

## Verified against the reporter's reproduction

A directory containing only:

```objc
@interface CGRepro
- (void)run;
@end
```

**Before** (`codegraph init --yes` on current `main`):

```
1 parser initialization failures
# .codegraph/errors.log
repro.h: Failed to get parser for language: objc
```

**After:**

```
Indexed 1 files
2 nodes, 1 edges in 315ms
```

## One refactor, and why it's in scope

The three lines that build this set existed **twice** — once for the full index, once for the changed-file reindex — and fixing only the copy the issue points at leaves the other one carrying the same bug for incremental runs. Both now call `preloadLanguagesForFiles()`, which also makes the rule unit-testable instead of buried inside a 200-line method.

## Tests

`__tests__/preload-languages.test.ts`, four cases: both ambiguous readings of a `.h` are covered; a project with no C-family header gets neither grammar added; a language the files already need isn't duplicated; extension overrides still drive the base set.

Confirmed red by reverting just the `objc` half of the change (`expected [ 'c', 'cpp' ] to include 'objc'`), so the test pins this fix rather than restating the implementation.

Full suite: 179 files / 3056 tests passing (3052 on `main` plus these four). `tsc --noEmit` clean.

Reported and diagnosed by @Juddd.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -242,6 +242,8 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 - Fixed a long-running `codegraph ui` session serving a symbol that a sync had already deleted. The viewer keeps one connection to your index open, and its in-memory lookup didn't notice when another process — your agent's sync, or `codegraph sync` — rewrote the file underneath it, so a symbol screen could keep showing a body with no callers while search correctly reported it had moved. Because a symbol's identity includes the line it starts on, this happened after almost any edit above it.
 
+- Objective-C headers now index in a project that has no `.m` file. A `.h` file is read as C from its name alone, and only later — once its contents are read — recognized as Objective-C; the grammar for that was never loaded up front, so the file failed with a parser error and nothing in it reached the index. Adding any `.m` file used to make the same header work, which is what made this look arbitrary. Thanks @Juddd. (#1628)
+
 ## [1.6.0] - 2026-08-26
 
 ### Highlights
--- a/__tests__/preload-languages.test.ts
+++ b/__tests__/preload-languages.test.ts
@@ -1,0 +1,41 @@
+/**
+ * Grammar preload set for a file list (#1628).
+ *
+ * Path-only detection calls every `.h` file C, but parse-time detection reads
+ * the source and can reclassify it as C++ or Objective-C. Workers only ever
+ * receive the grammars named by this set, so a header that turns out to be
+ * Objective-C in a project with no `.m` file had no parser to go to and the
+ * file failed outright with `Failed to get parser for language: objc`.
+ */
+
+import { describe, it, expect } from 'vitest';
+import { preloadLanguagesForFiles } from '../src/extraction';
+
+describe('grammar preload set (#1628)', () => {
+  it('covers both ambiguous readings of a .h file, C++ and Objective-C', () => {
+    const langs = preloadLanguagesForFiles(['repro.h']);
+    // Path-only detection says C…
+    expect(langs).toContain('c');
+    // …and parse-time detection may say either of these instead.
+    expect(langs).toContain('cpp');
+    expect(langs).toContain('objc');
+  });
+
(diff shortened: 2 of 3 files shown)


colbymchenry:
Superseded by #1781 — same fix (preload objc alongside cpp for C-family headers via preloadLanguagesForFiles) cherry-picked onto current main with CHANGELOG under Unreleased/Fixes/Symbols and Linux fail→pass verified. Thanks @maxmilian!

colbymchenry:
Closing as superseded by #1781.

Issue #1626, opened 2026-08-26
Python: aliased module imports (`import x as y`, `from p import x as y`) still produce zero call edges — remaining gap after #578 / #715


JoeyNPP:
## Summary

#578 (module-attribute call sites after `from pkg import module`) was fixed by #715 and the plain form now resolves correctly. But the **aliased** forms still produce no `calls` edges on 1.6.0 (latest, npm). The call lands in `unresolved_refs` with `status='failed'` and the caller is silently absent from `codegraph_callers` and from the MCP `codegraph_explore` blast-radius summary.

## Repro (1.6.0, win32-x64, Python project)

```python
# services/kit.py
def sync_subscribers(get_db):
    ...

# routes/email.py
from services import kit as kit_service        # <-- aliased from-import

def handler():
    kit_service.sync_subscribers(get_db)       # no calls edge
```

```python
# hook.py
import build_api_contract as builder           # <-- aliased plain import

builder.source_fingerprint()                   # no calls edge
```

Observed on a real repo (side-by-side in the same index, same session):

| Import form | `calls` edge created? |
|---|---|
| `from services import email_send` -> `email_send.log_sent()` | yes (fixed by #715) |
| `from services import leads as leads_svc` -> `leads_svc.convert_lead_to_buyer()` | no |
| `import build_api_contract as builder` -> `builder.source_fingerprint()` | no |

Module-level vs function-scoped import makes no difference — both aliased forms miss.

`unresolved_refs` rows for the failing sites look like:

```
('builder.source_fingerprint', 'calls', 'failed')
('kit_service.sync_subscribers', 'calls', 'failed')
```

## Why it bites

Same consequence as #578: `codegraph_callers` reports a function as having one caller (a script) when a live route handler also calls it through the alias — "is this dead code?" gets the wrong answer. The MCP `codegraph_explore` output is also self-inconsistent in this case: the blast-radius header says "1 caller in scripts/kit_sync.py" while the verbatim source it prints below shows the aliased call site.

## Suggested fix

Wherever #715 taught the Python resolver that a bare imported-module name qualifies attribute calls, the alias table needs to participate: `import x as y` / `from p import x as y` should register `y` as a binding for module `x` in the same scope map.

maxmilian:
[...]

PR #1635, opened 2026-08-28
fix(resolution): resolve Python module members through an aliased from-import (#1626)

maxmilian:
Fixes the `from … import … as …` half of #1626.

`resolvePythonModuleMember` rebuilds the submodule's dotted path by joining the import source with the **local** name. Under `from pkg import mod as alias` that produces `pkg.alias` — a module that doesn't exist — so the file lookup finds nothing, the member never resolves, and the call lands in `unresolved_refs` with `status='failed'`. `codegraph_callers` then reports the target as having fewer callers than it does, which is exactly the wrong "is this dead code?" answer that #578 produced for the unaliased form.

Joining with the exported name fixes it. For an unaliased import the two names are identical, so nothing changes there; the namespace form (`exportedName === '*'`) keeps binding at the local name, which is what it already did.

## Scope: the report has two halves and only one reproduces

The issue also lists `import build_api_contract as builder` → `builder.source_fingerprint()` as producing no edge. **That form resolves on current `main`.** It's a namespace import that binds at `source` — the real module name — so it never goes through the local-name join this PR fixes.

I checked rather than assumed: with the resolver change reverted, a probe on the plain-aliased call site in the test fixture still reports one `calls` edge. So if @JoeyNPP is still seeing that case fail on a real project, there's a second cause and it isn't this one — worth keeping #1626 open on that half, or splitting it out.

## Tests

One case added beside the existing #578 test in `resolution.test.ts` (same integration shape — real `CodeGraph.init` with `index: true`), asserting **both** halves:

- `from pkg import module as mod_alias` → `mod_alias.func()` resolves to `pkg/module.py` — fails before the change (`expected [] to have a length of 1`)
- `import top_level as tl` → `tl.top_func()` resolves — green before and after, pinned so fixing one half can't break the other

Full suite: 178 files / 3053 tests passing (3052 on `main` plus this one). `tsc --noEmit` clean.

Reported by @JoeyNPP.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -242,6 +242,8 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 - Fixed a long-running `codegraph ui` session serving a symbol that a sync had already deleted. The viewer keeps one connection to your index open, and its in-memory lookup didn't notice when another process — your agent's sync, or `codegraph sync` — rewrote the file underneath it, so a symbol screen could keep showing a body with no callers while search correctly reported it had moved. Because a symbol's identity includes the line it starts on, this happened after almost any edit above it.
 
+- Python calls made through a renamed import are no longer missing from the graph. After `from package import module as alias`, a call like `alias.func()` produced no call edge, so `codegraph_callers` could report a function as uncalled while a live caller reached it under the alias — the same wrong answer to "is this dead code?" that the unaliased form used to give. Thanks @JoeyNPP. (#1626)
+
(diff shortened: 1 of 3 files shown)


colbymchenry:
Superseded by Forge PR https://github.com/colbymchenry/codegraph/pull/1785, which cherry-picks this change (`f7a8940e` → `57d56859`) and extends the same exported-name join to `resolveModuleImportToFile` so the aliased from-import also restores the file→file `imports` edge. Thanks @maxmilian — closing this as superseded.

colbymchenry:
Superseded by #1785.

Issue #1545, opened 2026-08-12
PHP: static call through an import alias resolves to the wrong same-named method


justin81527:
# PHP: static call through an import alias resolves to the wrong same-named method

**Version:** 1.5.0 · **Language:** PHP (Laravel)

## Summary

For a static call made through an import alias — `Alias::method()` where
`use Some\Fqn\Klass as Alias;` — the receiver is not used to disambiguate.
Resolution falls back to matching the *method name* alone, so when two classes
define the same method name the call is attributed to the **wrong** class,
silently.

`extractPHPImports` already captures the alias correctly, so this looks like the
mapping simply isn't consulted when resolving `scoped_call_expression`.

## Reproduction

```php
// app/Services/SettleService.php
namespace App\Services;
class SettleService {
    public static function getSettlesToExcel($stores, $selected, $dayRange, $id) { /* … */ }
}

// app/Repositories/SettleRepository.php
namespace App\Repositories;
class SettleRepository {
    public static function getSettlesToExcel($storeIds, $dayRange) { /* … */ }
}

// app/Http/Controllers/Backend/SettleController.php
namespace App\Http\Controllers\Backend;
use App\Services\SettleService as Settle;          // ← alias

class SettleController extends Controller {
    public function excel($stores, $startDay = null, $endDay = null, $id = null) {
        return Settle::getSettlesToExcel($stores, request()->_SELECTED, [$startDay, $endDay], $id);
    }                //  ↑ the SERVICE
}
```

### Expected

`SettleController::excel` is a caller of **`App\Services\SettleService::getSettlesToExcel`**.

### Actual

```console
$ codegraph impact getSettlesToExcel

Impact: "App\Services::SettleService::getSettlesToExcel (app/Services/SettleService.php:46)" affects 1 symbols
  app/Services/SettleService.php:  getSettlesToExcel:46            ← no callers found

Impact: "App\Repositories::SettleRepository::getSettlesToExcel (app/Repositories/SettleRepository.php:105)" affects 3 symbols
  app/Repositories/SettleRepository.php:  getSettlesToExcel:105
  app/Http/Controllers/Backend/SettleController.php:  excel:52     ← WRONG CLASS
  routes/web.php:  GET /excel/{startDay?}/{endDay?}/{id?}:149
```

```console
$ codegraph callers getSettlesToExcel --file app/Services/SettleService.php
No callers found for "getSettlesToExcel"
```

```console
[...]

Issue #1496, opened 2026-08-01
TypeScript: a call through `this.<field>` resolves to the ENCLOSING method when the two share a name — silent self-edge, 0% recall on that shape


assafkamil:
## Summary

`this.someField.doThing()` resolves to the **calling method** instead of the
field's method, whenever the calling method happens to be named `doThing` too.
The result is a stored `calls` self-edge — `A::doThing → A::doThing` — so
`callers`, `callees`, `impact`, and `trace` are wrong, silently and plausibly.

The identical call resolves **correctly** if the calling method has any other
name. Name collision is the only variable; no framework, no decorators, no DI
involved.

Version `1.5.0` (npm, `darwin-arm64` bundle). Language: TypeScript.

## Reproduction

Three files, no dependencies.

```ts
// src/mailer.ts
export class Mailer {
  send(msg: string): string { return msg; }
}
```

```ts
// src/outbox.ts — calling method SHARES the callee's name
import { Mailer } from './mailer';
export class Outbox {
  private mailer = new Mailer();
  send(msg: string): string { return this.mailer.send(msg); }
}
```

```ts
// src/relay.ts — identical call, calling method has a DIFFERENT name
import { Mailer } from './mailer';
export class Relay {
  private mailer = new Mailer();
  forward(msg: string): string { return this.mailer.send(msg); }
}
```

```console
$ codegraph init && codegraph callers Mailer.send
Callers of "Mailer.send" (1):

method      forward
  src/relay.ts:6
```

`Outbox::send` is missing. The stored edges show where it went:

```console
$ sqlite3 .codegraph/codegraph.db "SELECT n1.qualified_name, n2.qualified_name
    FROM edges e JOIN nodes n1 ON n1.id=e.source JOIN nodes n2 ON n2.id=e.target
    WHERE e.kind='calls';"

Outbox::send    ->  Outbox::send      ❌ self-edge
Relay::forward  ->  Mailer::send      ✅
```

## Expected

Both should resolve to `Mailer::send`. `Outbox::send` does not call itself.

## The field style does not matter

Same fixture, varying only how `mailer` is declared. Only the name collision
changes the outcome:

| field declaration | calling method | result |
| --- | --- | --- |
| `private mailer = new Mailer()` | `forward` | ✅ `Mailer::send` |
| `private mailer: Mailer` (assigned in ctor) | `forward` | ✅ `Mailer::send` |
[...]

PR #1640, opened 2026-08-28
fix(go): resolve a package-qualified accessor chain on the factory's return type instead of a same-named method

jiang-bx:
## The shape

`pkg.Factory().Method()` reaches the resolver as a bare `Method` ref, so the name
fallbacks match it against **any same-named method on an unrelated type** — and,
when the enclosing method happens to share the name, against **itself**.

That is not hypothetical on a GoFrame app. In hotgo, one of the repos in the
coverage playbook:

```go
func (c *cTable) List(ctx context.Context, req *sysin.TableListReq) (*sysin.TableListRes, error) {
	list, totalCount, err := service.SysTable().List(ctx, &req.TableListInp)
	//                                        ^ resolved to cTable::List — a self-edge
}
```

`service.X()` is the accessor `gf gen service` writes for every service in every
GoFrame project, so this is the normal shape of a call into business logic, not
an edge case. The same failure hits a concrete-returning factory —
including the `Foo.factory().method()` case #760 landed for Go, whose stated
invariant is that *"a wrong inference produces no edge, never a wrong one"*
(#750).

## Root cause — two layers

**Extraction.** A chained receiver is re-encoded as `<innerCallee>().<method>`
so resolution can infer the type from what the inner call returns (#645/#608).
For Go this fired only when the inner callee was a bare `identifier`
(`New().Method()`). A package-qualified factory is a `selector_expression` — the
*same node type* as an instance chain (`obj.Method().Other()`), whose receiver
type isn't recoverable and which must stay bare — so both were dropped and the
ref went out as a bare name.

**Resolution.** Even re-encoded, `matchDottedCallChain` treated the dotted prefix
as a receiver type and looked up `service::SysTable`. Go package-level functions
carry a **bare** `qualifiedName` (`SysTable`, not `service.SysTable`), so that
lookup can never match.

## The fix

1. **`src/extraction/tree-sitter.ts`** — re-encode a Go `selector_expression`
   inner callee when its operand names a package the file imports (alias
   included). The import set is what separates a package qualifier from a
   variable receiver, so instance chains keep the behaviour they had.
2. **`src/resolution/name-matcher.ts`** — a Go branch in `matchDottedCallChain`:
   look the factory up by name, map the call-site qualifier back through the
   file's imports (an alias doesn't name its directory), disambiguate by the
   import path's tail, and validate the method on the inferred type via
   `resolveMethodOnType`. Candidates that disagree on their return type yield
   **no edge** rather than a guess.

An interface return lands on the **interface's** method, which the
dynamic-dispatch pass already bridges to the implementation — closing
`route -> controller -> service -> implementation` end to end.

## Validation on the three public GoFrame repos

Baseline vs. this build, same clone, `codegraph init` each way. **Node count is
identical on all three**, and the `route` edges #747 builds are untouched:

| repo | files | nodes | edges | `controller -> service -> logic` 2-hop paths | route edges |
|---|---|---|---|---|---|
| gf-demo-user | 38 | 248 → 248 | 474 → 474 | **0 → 7** | 7 → 7 |
| gfast | 184 | 1,875 → 1,875 | 4,260 → 4,236 | **0 → 82** | 65 → 65 |
| hotgo/server | 697 | 9,592 → 9,592 | 21,739 → 21,609 | **0 → 327** | 243 → 243 |

The chain from a controller into the logic it calls did not exist before, in any
of the three.

**Every removed edge sampled was wrong**, in four flavours:

- the controller self-edge above;
- a call matched to a same-named **type** — `service.SysTable().View(…)` landed
  on `model.View`, a struct;
- a chain through a dependency's accessor — `gjson.New(req).String()` landed on
  an unrelated `apiItem::String` in `hggen/internal/cmd/genctrl`;
- likewise `g.Redis().Do(ctx, "GET", k)` landing on a *project* function named
  `Do`.

Those now resolve to nothing, which is the correct answer for a receiver whose
type lives in a dependency.

## Tests

`__tests__/go-package-accessor-chain.test.ts`, modelled on
`php-property-receiver-resolution.test.ts` — 8 cases: interface accessor vs.
decoy, reaching the impl through the interface, concrete accessor vs. decoy,
import alias, picking the aliased package the *import path* names over another
exporting the same function, absent-method safety, a factory in a package
outside the index, and a guard that an instance chain stays on the bare-name
path.

Reverting the two source changes turns **6 of the 8 red**. The two that stay
green are the instance-chain guard and absent-method safety — both pin behaviour
that must not change. Reverting only the alias mapping reds exactly the one
alias test.

Full suite: `179 passed | 16 skipped`, 3,060 tests.

## Also measured on a large private repo

A GoFrame monorepo, 3,328 files / 44,564 nodes, where `gf gen ctrl` and `gf gen
service` make the controller method, the service interface method and the
implementation share one method name by construction. Grouping every `calls`
edge out of its test tree by whether a same-named controller method exists:

| target name | lands on the implementation | lands on a controller method |
|---|---|---|
| no same-named controller method | 1,950 | 0 |
| has a same-named controller method | **0** | **2,910** |

After: calls landing on a controller method fall 3,067 → 268 (the remainder are
ORM fluent chains — `dao.X.Ctx(…).Update(…)` — through an external type, a
different shape not addressed here), and two-hop paths from a test through the
service interface to the implementation go 0 → 5,619.

One concrete consequence, `explore` with the implementation file pinned:

```
before   implementation -> "no tests found within 3 caller hops"   (it has 16)
         controller     -> "20 callers; tests: create_test.go, ..."

after    implementation -> "tested via callers: create_test.go, index.go +1"
         controller     -> "2 callers; no tests found"
```

## Related

#750 (tracking; the invariant), #760 / #469 (Go chained factories), #747 (the
GoFrame route resolver that builds the first half of this chain), #1496 (the
same self-edge shape in TypeScript), #1585 / #1545 (Rust / PHP).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixes
+
+- Go calls made through a package-level accessor — `service.Order().Create(...)`, the shape `gf gen service` generates for every GoFrame project — now resolve to the method the accessor's return type actually declares. Previously the chain was dropped while reading the code and only the method name was matched, so any same-named method on an unrelated type could win. In a layered Go app, where the controller, the service interface, and the implementation deliberately share one method name, the call landed on the controller: the path from a route down to the business logic was never joined up, and blast radius credited the implementation's tests to the wrong function. An accessor returning an interface now lands on the interface's method, which the dynamic-dispatch pass already bridges to the implementation. Instance chains (`obj.Method().Other()`), whose receiver type isn't recoverable, keep the name-based behaviour they had.
+
 
 ## [1.6.0] - 2026-08-26
 
--- a/__tests__/go-package-accessor-chain.test.ts
+++ b/__tests__/go-package-accessor-chain.test.ts
@@ -1,0 +1,265 @@
+/**
+ * Go package-qualified accessor chains — `pkg.Factory().Method()`.
+ *
+ * The extractor re-encodes a chained receiver as `<innerCallee>().<method>` so
+ * resolution can infer the method's type from what the inner call RETURNS
+ * (the #645/#608 mechanism). Go used to re-encode only a BARE inner callee
+ * (`New().Method()`, an `identifier`), which left the package-qualified form
+ * `service.Order().Method()` — the `gf gen service` accessor every GoFrame app
+ * has — emitting a bare `Method` ref that the name fallbacks then matched
+ * against ANY same-named method on an unrelated type.
+ *
+ * A package-qualified inner callee is a `selector_expression`, the same node
+ * type as an instance chain (`obj.Method().Other()`), whose receiver type is not
+ * recoverable and which must therefore stay bare — re-encoding it would drop the
+ * edge instead. The file's import set separates the two, so the guards below pin
+ * both directions.
+ *
+ * Resolution then reads the factory's declared return type. Go package-level
+ * functions carry a BARE qualifiedName (`Order`, not `service.Order`), so the
+ * `Class::method` lookup used by the dot-notation languages can never match;
+ * the factory is looked up by name and disambiguated by its package directory.
+ * The method is validated on the inferred type through `resolveMethodOnType`, so
+ * a type that lacks the method yields no edge rather than a decoy. An interface
+ * return lands on the INTERFACE's method, which the dynamic-dispatch pass then
+ * bridges to the implementation — closing caller -> interface -> impl.
+ */
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+import * as os from 'node:os';
+import { CodeGraph } from '../src';
+
+describe('Go package-qualified accessor chains', () => {
+  let dir: string;
+  beforeEach(() => { dir = fs.mkdtempSync(path.join(os.tmpdir(), 'go-pkg-accessor-')); });
+  afterEach(() => { fs.rmSync(dir, { recursive: true, force: true }); });
+
+  const write = (rel: string, body: string) => {
+    const p = path.join(dir, rel);
+    fs.mkdirSync(path.dirname(p), { recursive: true });
+    fs.writeFileSync(p, body);
+  };
+
+  const load = async () => {
+    const cg = await CodeGraph.init(dir, { silent: true });
+    await cg.indexAll();
+    const db = (cg as any).db.db;
+    const calls: { src: string; tgt: string; tgtQn: string }[] = db
+      .prepare(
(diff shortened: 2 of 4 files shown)


jiang-bx:
Force-pushed to `0abaef2`. Three changes since the first push, and I've rewritten
the description to match.

**1. Validation on the three public GoFrame repos from the coverage playbook.**
The original description only had numbers from a private repo, which isn't
reproducible for you. Baseline vs. this build, same clone:

| repo | files | nodes | edges | `controller -> service -> logic` 2-hop paths | route edges |
|---|---|---|---|---|---|
| gf-demo-user | 38 | 248 → 248 | 474 → 474 | **0 → 7** | 7 → 7 |
| gfast | 184 | 1,875 → 1,875 | 4,260 → 4,236 | **0 → 82** | 65 → 65 |
| hotgo/server | 697 | 9,592 → 9,592 | 21,739 → 21,609 | **0 → 327** | 243 → 243 |

Node count identical on all three, #747's route edges untouched. hotgo shows the
sharpest form of the bug — `service.SysTable().List(...)` inside
`func (c *cTable) List(...)` resolved to `cTable::List`, a self-edge, so the
`callers`/`impact`/blast-radius answers for every GoFrame service method were
wrong. Every removed edge I sampled was wrong: that self-edge, calls matched to a
same-named *type* (`service.SysTable().View(...)` → `model.View`), and chains
through a dependency's accessor (`gjson.New(req).String()` → an unrelated
`apiItem::String`; `g.Redis().Do(...)` → a project function named `Do`).

**2. An alias no longer loses package disambiguation.** The first push matched
the call-site qualifier against the candidate's directory, which an alias
detaches (`ctrlcart "…/order/cart"`). It now maps the qualifier back through the
file's imports and matches on the import path's tail, so aliased and
renamed-package imports disambiguate like any other. New test; reverting just
that mapping reds exactly that one test.

**3. I withdraw the open question from the original description.** I wrote that a
recovered receiver type lacking the method would still fall through to the
bare-name path and let a decoy win. It doesn't — the re-encoded ref is
`inner().method`, and the downstream fallbacks look that whole string up and find
nothing. So #750's invariant holds here as written; there's no design call left
for you to make on it. There's now a test pinning it (absent-method safety),
which stays green with the fix reverted, as it should.

Tests are 8 now; reverting the two source changes reds 6 of them. Full suite
`179 passed | 16 skipped`.

danusha2345:
Reproduced a NO-GO case against this branch: when the imported factory lives OUTSIDE the index, `ext.Build().Run()` resolves to a same-named local — a project-local `Build()` returning `IWrong` gets the edge `→ IWrong::Run`. The fallback to "all same-named candidates" is what does it; a receiver whose type cannot be pinned should yield no edge rather than the nearest name. Gating the fallback on the factory's definition being in the index (or on a single candidate) would keep the fix for the in-index case.

PR #1641, opened 2026-08-28
fix(pascal): detect unqualified paren-less calls (`DoWork;`)

equipados:
## What

A Pascal statement that is nothing but an identifier — `DoWork;` — is an
unqualified call to a parameterless routine. `visitPascalBlock` didn't treat it
as one, so only the parenthesised form `DoWork()` produced a `calls` edge.

Pascal lets a no-arg routine drop its parens, and in Delphi that is the
convention rather than the exception, so on a real Delphi codebase most
intra-unit calls were invisible to `callers` / `callees` / `impact`. Those tools
didn't return a wrong answer — they returned an empty one, which is harder to
notice.

The qualified form (`Obj.Free;`) is already handled by
`extractPascalParenlessCall` (thanks — that landed in 1.5). This is the same
idea one step less qualified.

## Measured on a real Delphi codebase

285 `.pas`/`.dpr` files from a production Delphi 10.2 ERP (~1.1M LOC):

| | `calls` edges |
|---|---|
| 1.5.0 as-is | 2316 |
| with this patch | **2711** |
| | **+395 (+17%)** |

## How it's scoped

Gated on the identifier *being the whole statement*:

```ts
child.type === 'identifier' &&
node.type === 'statement' &&
node.namedChildCount === 1
```

Anywhere else — assignment RHS, `if` condition, any expression — a bare
identifier is indistinguishable from a variable read without type information,
so those are deliberately left alone. That is what the second test locks in:

```pascal
procedure TFoo.DoThing;
var
  Total, Other: Integer;   // and TFoo has a method also called Total
begin
  Other := Total;                    // variable read, NOT a call
  if Total > 0 then Other := 0;      // variable read, NOT a call
end;
```

`isCalled('TFoo::Total')` must stay `false`.

## Tests

Two added next to the existing paren-less test in
`__tests__/resolution.test.ts`:

- `extracts UNQUALIFIED paren-less calls (`Reset;`, the dominant Delphi idiom)` — fails without the change, passes with it.
- `does not turn a bare identifier that is NOT a whole statement into a call` — the negative case above.

`resolution.test.ts` + `extraction.test.ts`: 788 passed. The 5 failures on my
machine (3× PHP include, 1× C/C++ include, 1× gitlink) are identical before and
after the change — pre-existing and unrelated, likely Windows-specific.

`tsc --noEmit` clean.

## Notes

Only `src/extraction/tree-sitter.ts` changes; Pascal is not part of the Rust
kernel so there is no parity fixture to update.

### DIFF
--- a/__tests__/resolution.test.ts
+++ b/__tests__/resolution.test.ts
@@ -4818,6 +4818,65 @@ end.
       expect(isCalled('TFoo::Reset')).toBe(true);
     });
 
+    it('extracts UNQUALIFIED paren-less calls (`Reset;`, the dominant Delphi idiom)', async () => {
+      fs.writeFileSync(
+        path.join(tempDir, 'main.pas'),
+        `unit Main;
+interface
+type
+  TFoo = class
+    procedure DoThing;
+    procedure Reset;
+    procedure Refresh;
+  end;
+implementation
+procedure TFoo.Reset; begin end;
+procedure TFoo.Refresh; begin end;
+procedure TFoo.DoThing;
+var
+  Total: Integer;
+begin
+  Reset;
+  Refresh;
+  Total := 1;
+end;
+end.
+`
+      );
+      cg = await CodeGraph.init(tempDir, { index: true });
+      expect(isCalled('TFoo::Reset')).toBe(true);
+      expect(isCalled('TFoo::Refresh')).toBe(true);
+    });
+
+    it('does not turn a bare identifier that is NOT a whole statement into a call', async () => {
+      fs.writeFileSync(
+        path.join(tempDir, 'main.pas'),
+        `unit Main;
+interface
+type
+  TFoo = class
+    procedure DoThing;
+    procedure Total;
+  end;
+implementation
+procedure TFoo.Total; begin end;
+procedure TFoo.DoThing;
+var
+  Total, Other: Integer;
+begin
+  Other := Total;
+  if Total > 0 then Other := 0;
+end;
+end.
+`
+      );
+      cg = await CodeGraph.init(tempDir, { index: true });
+      // `Total` here is a variable read on an assignment RHS and inside a
+      // condition, never a statement of its own: without type info the two are
+      // indistinguishable, so only the statement form counts as a call.
+      expect(isCalled('TFoo::Total')).toBe(false);
+    });
+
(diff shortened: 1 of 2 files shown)


danusha2345:

equipados:
Thanks for putting it through a full integration build — the forced
re-index on
codegraph's own repo is exactly the check I wanted to see, since the risk
here was
changing edges where the parser already behaved.

If it looks good to you, could you submit it as an approving review? The PR
still
shows REVIEW_REQUIRED, so it can't be merged on the comment alone.

Happy to rebase on current main if that helps.
Guillermo Álvarez Leal

*https://equipados.es* <https://equipados.es>


38611 Granadilla de Abona
Tlf:: 655 778 795

Este correo electrónico y, en su caso, cualquier fichero anexo al mismo,
contiene información de carácter confidencial exclusivamente dirigida a su
destinatario. En caso de su divulgación, copia o distribución a terceros
sin la previa autorización escrita de Equipados, usted podría estar
vulnerando la legislación vigente. En el caso de haber recibido este correo
electrónico por error, se ruega que nos notifique inmediatamente esta
circunstancia mediante reenvío a la dirección electrónica del remitente. De
conformidad con lo establecido en la LOPD Equipados garantiza la adopción
de las medidas necesarias para asegurar el tratamiento confidencial de los
datos de carácter personal. Así mismo le informamos de inclusión de
sus datos en un fichero con la finalidad de mantener su
relación contractual con Equipados. y de la posibilidad de ejercer los
derechos de acceso, rectificación, cancelación y oposición en la
siguiente dirección: Apartado de correos, nº 190, C.P 38611 – Granadilla
de Abona - Santa Cruz de Tenerife.

danusha2345 (approved):
Approving as a fellow contributor, with one caveat worth stating plainly: my review does not unblock the merge. I have no write access to this repository, so `REVIEW_REQUIRED` can only be cleared by @colbymchenry — this shows up as an approval in the list, nothing more. Treating it as a validation record rather than a gate.

What it rests on: the branch merged clean into a local integration build of current `main` plus the other open fixes, full suite with the native kernel 245 files / 4323 passed / 0 failed, and a forced re-index of codegraph's own repo with no edge regressions.

No rebase needed — GitHub still reports the PR `MERGEABLE` against today's `main` (unchanged since `b9ca4b7`, 31 Aug), and the branch is one commit ahead of the merge-base. Rebasing would only change the SHA.

Issue #1512, opened 2026-08-06
CLI `callers`/`callees`/`impact` merge every same-named definition into one unlabeled result and expose no `--file`, while the MCP tools group and filter on the same index — the "CLI parity" surface returns a union true of no single definition — verified end-to-end


inth3shadows:
## Summary

`codegraph callers|callees|impact <symbol>` merge the edges of **every** same-named definition into one flat result printed under the bare symbol name. The merge is deliberate (`// Merge impact subgraphs across all exact-matching symbols`, `src/bin/codegraph.ts:2030`), but there is no `--file` option to scope the query and nothing in the output says a merge happened. When a name exists in more than one file — `render`, `init`, `get`, `handle`, `String` — the answer is a **union that is true of no individual definition**.

The MCP tools already solve both halves. `handleCallers` (`src/mcp/tools.ts:2130`), `handleCallees` (`:2203`) and `handleImpact` (`:2273`) each accept a `file` argument and route through `groupDefinitions` (`:2093`), so every definition keeps its own edge set, the header says how many definitions were found, and a `file` that matches nothing is reported rather than ignored.

The CLI commands read the same index and answer the same question with neither capability — while their own header comment states the purpose is *"CLI parity with the MCP graph tools (codegraph_callers/callees/impact) so the traversal queries work in scripts, CI, and git hooks without a running MCP server"* (`src/bin/codegraph.ts:1843`). The surface documented as the no-MCP-server equivalent is the one that silently merges.

## Root cause

`callers` (`src/bin/codegraph.ts:1848`), `callees` (`:1927`), `impact` (`:2005`). All three declare only `-p/--path`, `-j/--json`, and `-l/--limit` (`-d/--depth` for `impact`). `callees` shows the shape; `callers` is structurally identical:

```ts
const matches = cg.searchNodes(symbol, { limit: 50 });
...
for (const match of matches) {
  const exactMatch = match.node.name === symbol || match.node.name.endsWith(`.${symbol}`) || match.node.name.endsWith(`::${symbol}`);
  if (!exactMatch && matches.length > 1) continue;
  for (const c of cg.getCallees(match.node.id)) {
    if (!seen.has(c.node.id)) {          // <- one dedupe set ACROSS definitions
      seen.add(c.node.id);
      allCallees.push({ name: c.node.name, kind: c.node.kind, filePath: c.node.filePath, startLine: c.node.startLine });
    }
  }
}
...
[...]

PR #1656, opened 2026-08-31
fix(cli): route callers/callees/impact through one symbol resolver

ferrine:
**Branch:** `ferrine:fix/symbol-lookup-consistency` → `main`
**Size:** 5 files, +453/−110 · **Tests:** `__tests__/symbol-lookup.test.ts` (28 new)

## Why

`codegraph callers`, `callees` and `impact` each resolved the symbol name their
own way, and none of them understood a qualified name. Asking for something like
`Accounts.Format.group` matched nothing, so the command quietly fell back to
whichever symbol full-text search happened to rank first — or reported "not
found" for a symbol that plainly exists.

Two root causes, and the second is the one that made it look intermittent:

1. The comparison was against the **bare** name, so a qualified query never
   matched.
2. The separator model assumed `::`, so a language whose module names contain
   dots (Elixir, protobuf FQNs, C# namespaces) could not be addressed at all —
   `A.B.C` is genuinely ambiguous between "container `A.B`, member `C`" and a
   module literally named `A.B.C`.

Because the fallback was FTS ranking, the wrong answer arrived with no signal
that it was a guess.

## What changed

- New `src/graph/symbol-lookup.ts` — one "what did the user mean by this name?"
  path shared by all three commands.
- Lookup consults the **exact-name index first** (complete and uncapped) and
  uses FTS only as a fallback, so a name that exists is never missed because
  ranking put it past a cap.
- Matching canonicalises the container so a dotted module name resolves whether
  written `A.B.C` or `A.B::C`.
- When a plain name matches several distinct definitions, the commands now say
  so and print the qualified spelling that narrows it, instead of silently
  merging unrelated symbols into one list.

## Notes for review

- No behaviour change for a name that was already resolving correctly; the tests
  cover the previously-broken shapes.
- This is the base for two other PRs (protobuf indexing needs dotted-container
  matching, and the multi-definition diversity fix builds on the same path).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixes
+
+- `codegraph callers`, `codegraph callees` and `codegraph impact` now understand qualified symbol names. Asking for something like `Accounts.Format.group` used to match nothing, so the command quietly fell back to whichever symbol full-text search happened to rank first — or reported "not found" for a symbol that plainly exists. Qualified names now select exactly the definition you named, including in languages whose module names themselves contain dots.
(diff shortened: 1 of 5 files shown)


colbymchenry:
Superseded by Forge PR https://github.com/colbymchenry/codegraph/pull/1801 which ports this branch onto current main and extends it with --file / per-definition JSON attribution for #1512.

PR #1662, opened 2026-08-31
fix(resolution): don't believe a bare-name call across language families

ferrine:
**Branch:** `ferrine:fix/cross-language-call-confidence` → `main`
**Size:** 3 files, +181/−12 · **Opt-out:** `CODEGRAPH_CROSS_FAMILY_CALL_FLOOR=0`

> ⚠ This changes resolution behaviour for every user and **removes** edges.
> Happy to discuss the principle before the code — see "Why this is your rule,
> not a new one" below.

## Why

Two problems at the same comparison, one masking the other.

**1. The single-candidate confidence penalty compares raw language tags.** A
`.tsx` file calling a `.ts` helper counts as cross-language and is scored 0.5
instead of 0.9 — that is most calls in any React codebase. `sameLanguageFamily`
already exists two functions above for exactly this.

**2. With that fixed, 0.5 identifies the real problem.** Cross-family `calls`
are deliberately allowed — FFI bindings and native bridges are real. But a
language's own builtins and macros have no definition in the graph to bind to,
so the lone same-named foreign symbol wins by default. Observed on a
mixed-language repository: an ExUnit `test` macro binding to a test runner's
fixture in TypeScript; Python's `dataclasses.field` binding to an unrelated
Elixir helper; `apply`, `max`, `min`, `path`.

The matcher already recognised these and wrote the verdict into the edge as a
low confidence — and every consumer then believed the edge anyway, so `callers`
and `impact` reported it as fact. On that repo it was thousands of impossible
edges, concentrated on a small set of generically-named symbols whose reported
callers were then mostly or entirely fictional. One helper's `impact` returned
dozens of affected symbols, nearly all of them in a language that cannot call it.

It is also **unstable under indexing scope**: indexing more files of one
language moved thousands of these edges elsewhere, because the fallback fires
only when no same-language candidate exists.

## Why this is your rule, not a new one

`src/resolution/index.ts` already hardcodes exactly this, for one language:

> *"The reverse direction is just as impossible: no other language can
> symbolically call into a .nix binding (interop is eval/CLI, never a linkable
> symbol) — without this, a Python script's `split()` lands on some module's
> `split = ...` binding as a low-confidence match."*

This PR generalizes that from a per-language special case to the rule it is an
instance of.

## What changed

- The single-candidate penalty and the fuzzy penalty compare **family**, not raw
  language tag.
- A name match that crosses a family and lands at or below the floor is
  **declined** rather than recorded.
- Tunable via `CODEGRAPH_CROSS_FAMILY_CALL_FLOOR`, `=0` restores the old
  behaviour — same convention as `CODEGRAPH_AMBIGUOUS_NAME_CEILING`.

## Blast radius, checked

- **RN/Expo JS→native is untouched.** Those resolve through the framework
  strategy (Strategy 1/2); this gate is in name matching (Strategy 3). The
  `Cross-language type/import gate (RN name collisions)` test passes.
- **Genuine FFI keeps its edge**, because it resolves through an import or a
  qualified name (0.85+), never a bare one.
- On the repository above: impossible cross-language call edges dropped from
  ~3,800 to ~100, with **no loss of same-language edges** and all `tsx`↔`ts`
  edges retained.
- I checked the one case that looked like a real cost — a pybind11 binding in
  the repo. Every removed `python`↔`cpp` edge pointed at C++ *test* files and
  headers via Python builtins (`bytes`, `iter`); **none** pointed at the actual
  bindings file. The real bridge was never captured before or after — that is a
  separate coverage gap, not a regression from this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixes
+
+- Calls no longer resolve to a same-named symbol in an unrelated language on the strength of the name alone. Nothing in Python can call a function in Elixir, but a language's own builtins and test macros — `test`, `describe`, `apply`, `max`, `field` — have no definition of their own to bind to, so a lone same-named symbol in another language was winning by default. On a mixed-language repository that produced thousands of impossible call edges, and the symbols they landed on reported callers and impact that were mostly or entirely fictional. Genuine cross-language calls — FFI bindings, native bridges — are unaffected, because those resolve through an import or a qualified name rather than a bare one. Set `CODEGRAPH_CROSS_FAMILY_CALL_FLOOR=0` to restore the old behaviour.
+- A `.tsx` file calling a `.ts` helper is no longer treated as crossing a language boundary. The two are the same runtime, but they were compared by file type rather than by language family, so ordinary calls throughout a React codebase were scored as low-confidence guesses.
+
 
 ## [1.6.0] - 2026-08-26
 
--- a/__tests__/resolution.test.ts
+++ b/__tests__/resolution.test.ts
@@ -4,7 +4,7 @@
  * Tests for Phase 3: Reference Resolution
  */
 
-import { describe, it, expect, beforeEach, afterEach } from 'vitest';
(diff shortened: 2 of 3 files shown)


danusha2345:

PR #1663, opened 2026-08-31
fix(resolution): never resolve a production call into a test file

ferrine:
**Branch:** `ferrine:fix/no-production-to-test-calls` → `main`
**Size:** 3 files, +267/−5 · **Opt-out:** `CODEGRAPH_TEST_TREE_GATE=0`

> ⚠ This changes resolution behaviour for every user and **removes** edges.
> Happy to discuss the principle before the code.

## Why

Tests depend on the code they exercise; the code never depends on its tests. So
a production symbol resolving *into* a test file is always wrong — and it happens
constantly, because the names that reach the name-matching fallback are exactly
the ones with no real definition to bind to.

The clearest case, from a real React codebase: components call `t(...)` from
`const { t } = useTranslation()`. `t` is destructured from a hook, so there is no
local node; it is never exported; and `useTranslation` is re-exported straight
from `react-i18next`, so the owner is outside the repo entirely. The only symbols
named `t` anywhere were four test-file locals — and **2,820 calls across 137
production page components all resolved to one i18n stub defined in a single
test**, making it the second most-called symbol in the codebase. Across the repo
this shape accounted for ~4,100 edges over 67 targets (`say`, `step`,
`monotonic`, `field`, …), and the symbols they landed on reported callers and
impact that were almost entirely fictional.

## What changed

Test-file definitions are withheld from references that come from production
code. A reference *from* a test is left alone — tests calling test helpers is the
normal direction, and shared fixtures legitimately live in the test tree.

Applied at three strategies, because the defect reaches all of them: exact-name,
fuzzy, and — the one that needed it most — the receiver-name overlap fallback,
which scores a receiver *name* against a class name, so a short receiver
(`rep.say()`) matches a test's stand-in class (`_Rep`) on one shared word. Stand-
ins are by definition named after the collaborator they replace.

The gate is **unconditional, including when it empties the candidate set.** A
"keep it if nothing else matches" fallback would leave the entire defect in
place, since the fabricated edges are precisely the ones with no production
alternative — and production cannot depend on a test file in any case, as it
would not build.

## Identification: runner-enforced filenames only, never directories

This deletes edges, so a wrong identification is silent — and *which directories
hold a project's tests is the project's own decision*. `spec/` is an OpenAPI
document in one repo and RSpec in another; `testing/` is often a shipped utility
library. Guessing at directories would remove real edges from projects that
merely name a folder unluckily.

A filename is not a preference. pytest collects `test_*.py` and `*_test.py` and
nothing else, `go test` requires `_test.go`, vitest and jest default to
`*.test.*` / `*.spec.*`. A file named that way *is* a test, in every project, by
the runner's own rule.

I measured the split before choosing: of the impossible production→test edges on
that repo, **runner-enforced filenames accounted for all but one**. Directory
patterns would have added the risk without the reward, so they are not used.

The bare CamelCase suffix (`OrderTest.java`) is scoped to `.java/.kt/.kts/
.scala/.cs/.swift`, whose runners collect on it. Applied to JavaScript it
misreads any identifier ending in "Test" — `useTests.ts` is an ordinary hook and
vitest would never collect it. That one was caught by your existing
`Cross-language type/import gate (RN name collisions)` test, whose fixture uses
`useTests.ts`.

## Result

Production→test call edges dropped from ~4,100 to 20. The three highest-degree
symbols that remain are all legitimate, checked individually: two are called
exclusively *from* test files (the gate correctly leaves those alone), and the
third is a real utility with hundreds of genuine production callers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixes
+
+- Calls from your code no longer resolve into your tests. Tests depend on the code they exercise, never the other way round, but a name with no real definition to bind to — a value destructured from a hook, a helper that lives in a package — would land on any same-named helper inside a test file. On one project every `t(...)` call across 137 page components pointed at a translation stub defined in a single test, making it the second most-called symbol in the codebase and its callers and impact almost entirely fictional. Test files are recognised by the names their test runner enforces (`test_*.py`, `*.test.ts`, `*_test.go`, `*Test.java`), never by which folder they sit in, so a `spec/` directory holding an API document or a `testing/` utility library is left alone. Set `CODEGRAPH_TEST_TREE_GATE=0` to restore the old behaviour.
+
 
 ## [1.6.0] - 2026-08-26
 
--- a/__tests__/resolution.test.ts
+++ b/__tests__/resolution.test.ts
@@ -11,7 +11,7 @@ import * as os from 'os';
 import { CodeGraph } from '../src';
 import { Node, UnresolvedReference } from '../src/types';
 import { ReferenceResolver, createResolver, ResolutionContext } from '../src/resolution';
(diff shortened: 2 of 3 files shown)


danusha2345:

PR #1678, opened 2026-09-03
fix(extraction): index CommonJS export assignments as functions

### LINKED ISSUE #1675: CommonJS exports.X = async () => {} route handlers aren't indexed as symbols
**Version:** v1.6.0

Handlers written as `exports.foo = () => {}` / `exports.foo = async () => {}` (a very common Express/CommonJS controller style) don't seem to be indexed as function nodes. `node`/`callers` can't find them, and calls into them collapse to file level (the calling function is lost).

**Repro:**
1. `controller.js`:
   ```js
   exports.getItems = async (req, res) => { /* ... */ };
   ```
   wired in a router: `router.get('/items', getItems)`
2. `codegraph node getItems` → **"Symbol not found"**, although it's a real, route-wired handler.

**Impact:** in one real Express codebase, 60+ route handlers written in this style are invisible to the graph, so impact/callers analysis silently misses them.

**Ask:** index the `exports.<name> = (arrow | async arrow | function)` assignment form as a function node — the same treatment `const X = () => {}` already gets.

danusha2345:
Fixes #1675.

## Problem

`exports.getItems = async (req, res) => {…}` and `module.exports.deleteItem = function (req, res) {…}` — the common Express/CommonJS controller style — produced no function node. The arrow's parent is an `assignment_expression`, not a `variable_declarator`, so `extractFunction` left it `<anonymous>`, walked its body at file scope (calls attributed to the file), and `codegraph node getItems` / `callers` answered "Symbol not found" for a route-wired handler.

## Change

- `extractFunction` (wasm walker) and `extract_function` (kernel) resolve the name from the export property when the function is the whole right-hand side of `exports.NAME = …` / `module.exports.NAME = …`, and mark the node exported. Any other member assignment (`handlers.onSave = () => {}`) is untouched.
- Both arms move together; `torture.js` gains the shapes so `kernel-tsjs-parity` pins them.

## Verification

- New `__tests__/commonjs-exports.test.ts`: both forms become exported function nodes, `isAsync` preserved, calls belong to the handler not the file, `exports.plain = 42` and `module.exports = {…}` stay non-functions, other member assignments unchanged.
- `npm run build:kernel && npm run build`, full suite with `CODEGRAPH_KERNEL_EXPECT=1`: 237 files, 4233 passed, 9 skipped.

Re-index JavaScript projects after upgrading.


danusha2345:
Merged current `main` (`3adf067`, post-#1770) into this branch: head `be5ba15`. The only conflict was the CHANGELOG entry; no source conflict. `tsc` clean, the branch's tests pass.

colbymchenry:
Thanks @danusha2345 — verified the fix on Linux (wasm + kernel). #1678 was conflicting with `main` after #1770/#1684, so the same change is landed conflict-resolved as #1771. Feel free to close this PR in favor of that one.

danusha2345:
Closing in favor of #1771 — thanks for carrying it over.

Issue #1674, opened 2026-09-02
callers/node silently truncate at default --limit 20 (no "…N more" indicator)


azorlu80:
**Version:** v1.6.0

`codegraph callers <sym>` and `codegraph node <sym>` cap the caller list at the default `--limit 20` with **no indication** the result was truncated — the header shows the displayed count, never "…N more". `impact` (uncapped) and `node`'s callee trail (`+N more`) don't have this; the `callers` list and the JSON / MCP output do.

**Repro** (a high-fan-in symbol, e.g. a widely-used hook):
1. `codegraph callers useAuth` → **20** results
2. `codegraph callers useAuth --limit 500` → **174** results
3. The output/header gives no hint that 154 were dropped.

**Expected:** when truncated, print something like `…154 more (raise --limit)`, or don't cap silently.

**Impact:** consumers under-count and trust the result → wrong "who calls this / is this dead code?" conclusions. It's especially harmful when an AI agent drives the tool over MCP, since the tool output inherits the cap with no truncation signal.

danusha2345:
Same symptom as #1639. #1647 makes `callers` / `callees` / `query` report the cut (`shown of total`, plus `total`/`limit`/`truncated` in the callers/callees JSON), and its latest commit adds the same "showing N of M; pass `limit` to widen" note to the `codegraph_callers` / `codegraph_callees` MCP answers. `node`'s caller trail already prints `+N more` on 1.6.0.

colbymchenry:
Fixed in https://github.com/colbymchenry/codegraph/pull/1772 — lands the verified #1647 changes onto current `main` (that PR was conflicted after #1771).

- CLI `callers`/`callees`: header shows `N of M` + widening hint; JSON adds `total`/`limit`/`truncated`
- CLI `query`: truncation notice (stderr for `--json`)
- MCP `codegraph_callers`/`codegraph_callees`: `Showing N of M` / `… +N more`
- `codegraph node` already reported `+N more` on its caller trail; no change needed

Verified fail→pass on Linux with `__tests__/mcp-callers-truncation.test.ts` and `__tests__/cli-truncation.test.ts`.

PR #1679, opened 2026-09-03
fix(extraction): index const-bound functions inside a body as symbols

### LINKED ISSUE #1669: Function-valued consts inside a function body are never indexed (React handlers invisible to callers/impact)
## Summary

Function-valued `const`s are indexed at module scope but not inside a function
body. In a React codebase that means most event handlers are absent from the
graph, and `callers` / `callees` / `impact` answer `Symbol not found` for them
— which reads identically to "this symbol has no callers".

The boundary is scope, not syntax: a top-level `const fn = () => {}` indexes
fine, and the same declaration inside a component does not. `useCallback` is
not involved — a bare arrow inside the component is missing too.

## Reproduction

Two files, no dependencies, `codegraph init`:

`src/widget.jsx`
```jsx
import { useCallback } from "react";

export function formatLabel(value) {          // 1. top-level declaration
  return String(value).trim();
}

export const parseLabel = (raw) => raw.split(":")[1];   // 2. top-level arrow const

export class Formatter {                       // 3. class + method
  render(value) { return formatLabel(value); }
}

export default function Widget({ items, onPick }) {
  const handleSelect = useCallback((item) => {  // 4. const in a component, via hook
    onPick(formatLabel(item), parseLabel(item));
  }, [onPick]);

  const handleClear = () => {                   // 5. plain const in a component
    onPick(null, null);
  };

  return items.map((i) => (
    <button key={i} onClick={() => handleSelect(i)} onDoubleClick={handleClear}>
      {formatLabel(i)}
    </button>
  ));
}
```

`src/page.jsx`
```jsx
import Widget, { formatLabel, parseLabel } from "./widget";

export default function Page() {
  const label = formatLabel("a:b");
  const parsed = parseLabel("a:b");
  return <Widget items={[label, parsed]} onPick={() => {}} />;
}
```

## Result

`codegraph init` reports 10 nodes. Every node in the index:

```
file         page.jsx         src/page.jsx:1
import       ./widget         src/page.jsx:1
function     Page             src/page.jsx:3
file         widget.jsx       src/widget.jsx:1
import       react            src/widget.jsx:1
function     formatLabel      src/widget.jsx:4
function     parseLabel       src/widget.jsx:9
class        Formatter        src/widget.jsx:12
method       render           src/widget.jsx:13
function     Widget           src/widget.jsx:18
```

`handleSelect` (line 19) and `handleClear` (line 23) are absent.

```
codegraph callers formatLabel   → 4 callers
codegraph callers parseLabel    → 3 callers
codegraph callers Widget        → 1 caller
codegraph callers handleSelect  → Symbol "handleSelect" not found
codegraph callers handleClear   → Symbol "handleClear"  not found
```

Note line 9: `parseLabel` is an arrow function and indexes correctly. Only the
two declared inside `Widget` are missing.

## Expected

`handleSelect` and `handleClear` are declared functions with a name, a body and
call sites, and should appear as nodes the same way `parseLabel` does.

## Why it matters

This is the dominant shape for handlers in React, so on a real front end a
large share of the functions an agent needs are invisible. The failure is also
silent in the worst direction: `Symbol not found` is indistinguishable from
"nothing calls this", so an agent asked to update every call site of a handler
gets an empty answer and concludes there are none. Grep finds them, which is
the fallback CodeGraph is meant to replace.

The same gap presumably affects `const x = function () {}`, object-literal
methods, and IIFE-scoped helpers in plain JS/TS, though I only measured the
React case.

## Environment

- codegraph 1.6.0 (npm `@colbymchenry/codegraph`)
- node v26.7.0
- macOS 26.4.1, arm64

## Possibly related

#1656 (route callers/callees/impact through one symbol resolver) looks adjacent
— though this appears to be extraction rather than resolution, since the nodes
never reach the database at all.

danusha2345:
Fixes #1669.

## Problem

`const handleClear = () => {…}` inside a component — every React handler written without `useCallback` — was never a symbol. The body walker names nested function *declarations* and hook-bound arrows (`const x = useCallback(…)`), but a plain declarator-bound arrow or function expression fell through, so the handler was absent from `callers` / `callees` / impact ("Symbol not found", which reads exactly like "no callers") and its calls attributed to the component. The same declaration at module scope already names a function.

## Change

- Body walker (wasm) and kernel: an anonymous `arrow_function` / `function_expression` that is the whole `value` of a `variable_declarator` with a plain identifier name is extracted as a function, contained by the enclosing one, with its own calls. `extractFunction` already resolves the name from the declarator. A destructuring binding, an inline JSX arrow and a non-function value stay as they were. JS family only.
- `torture.tsx` gains the shape so `kernel-tsjs-parity` pins both arms.

## Graph shape

A navigation such a handler makes is now the handler's own `navigates` edge, and the handler is a hop in the Screens `via` chain — the shape a `useCallback` handler already has (see the Next.js test's `navs(sym('handleSubmit'))` and `push.via`). The react-router and expo-router expectations are updated to that convention; the screen still reaches the handler through its `onSubmit={…}` reference and the links resolve as before, with the trigger now visible in Steps.

## Verification

- New `__tests__/nested-declarator-functions.test.ts`: arrow, function expression and `let`-bound arrow become `Widget::name` functions with their own calls and containment; a value, a destructuring and an inline arrow do not; Python is untouched.
- `npm run build:kernel && npm run build`, full suite with `CODEGRAPH_KERNEL_EXPECT=1`: 237 files, 4233 passed, 9 skipped.

Re-index after upgrading to pick up the new symbols.


bompus:

Merges clean. The six closures we had listed as missing (`findAndAssign`, `assignFirstMatch`, `finish`, `collect`, `clear`, `tick`) all appear as `function` nodes under their parent (+154 nodes, +699 edges). PR test files: 100/100.

Two things the new nodes attract that were not there before:

1. Fuzzy call edges from unrelated files onto generically named closures: total `resolvedBy: fuzzy` edges 89 → 226. The new ones concentrate on `readSeedFailureState::text` (72 edges), `buildRows::push` (41), `captureListeners::event` (10). Every `.text(...)` or `.push(...)` call in the repo now has a fuzzy candidate. Worth scoping closure nodes out of the fuzzy resolver, or restricting them to same-file callers.
2. Self-call edges 31 → 42. The eleven new ones are all closures calling themselves (`walk`, `walkBag`, `tick`, `wrapElement`), which is real recursion, so no action needed there.

danusha2345:
Thanks — the fuzzy finding is real and #1709 is the right fix for it, so I took #1709 into my integration build ahead of this one as you suggested. Measured on two real repos with both PRs (and #1511) in, kernel on, same trees re-indexed before/after #1709: `resolvedBy: fuzzy` call edges 337 → 2 on this repo's own tree (787 files) and 142 → 0 on a Kotlin+Go+TS app (274 files); the new closure nodes stay, nothing else moves, full suite 253 files / 4351 passed. The self-call edges you list are genuine recursion here too. I would keep this PR as is and let #1709 land first.

bompus:
@danusha2345 before you sequence this behind #1709 — I ran your metric on a third repo and it moves the other way there.

Details and full tables on #1709: https://github.com/colbymchenry/codegraph/pull/1709#issuecomment-5557821470

Short version. Same vite checkout indexed at #1709's merge-base and at its tip, counting every edge by `resolvedBy`:

- `fuzzy`: **13 → 60 (+47)**
- every other resolver: **0 delta**

Cross-tabbed against the edge-set diff: 12 edges lost (all fuzzy, all correct removals of unreachable nested functions) and 59 gained (all fuzzy — 52 calls, 7 imports), nearly all pointing `node:path`'s `resolve`, imported on line 1 of a vite playground config, at `pluginContainer.ts:resolve`.

I had assumed your metric simply couldn't see this failure mode. It can — that's why I measured before replying. On your two trees the filter only removes bad guesses (337 → 2, 142 → 0); on vite it leaves exactly one surviving candidate and the matcher commits to it. vite has ~12 symbols named `resolve`, most nested, which is the shape that triggers it.

So your measurement stands for your trees and #1709's removal half is doing real work. My concern is only with landing it first: the underlying bug is one layer up — when filtering leaves a single survivor, we commit without checking the call site's binding is internal — and until that's handled, #1709 is not a strict improvement on every tree.

Nothing here affects this PR's own change. I'd just rather the ordering decision be made with the vite number in view than have it re-raised as a blocker on you later.

bompus:
Closing the loop I opened here. I raised the vite counter-example as a reason not to sequence this behind #1709; the underlying bug is now fixed in #1713, and I no longer think the sequencing concern should hold anything up.

#1713 is standalone — one commit off `main`, not stacked on #1709 — and on vite it removes 4 wrong edges and adds none, with every other resolver at 0 delta. With it in place, #1709 + #1713 is +3 fuzzy on vite rather than the +47 I reported for #1709 alone.

So the objection I raised against ordering is spent. Nothing here needs to wait on me.

Full tables and the two failure modes that remain: #1709 and #1713.

danusha2345:
Merged current `main` (`3adf067`, post-#1770) into this branch: head `ae6547d`. The only conflict was the CHANGELOG entry; no source conflict. `tsc` clean, the branch's tests pass.

danusha2345:
Merged current `main` (`43271f3`, post-#1771) into this branch: head `f023c39`. Conflicts were the CHANGELOG and src/extraction/tree-sitter.ts (the new declaratorBoundFunction and #1771's commonJsExportName sit side by side, both kept). Kernel rebuilt; `tsc` clean; the branch's tests and the kernel parity suites pass.

danusha2345:
Merged current `main` (`85550eb`) into this branch: head `35093c8`. The only conflict was the CHANGELOG entry; no source conflict. `tsc` clean, the branch's tests pass.

colbymchenry:
Thanks — verified and rebased onto current `main` (only CHANGELOG conflicted with #1772 / #1674+#1639). Landed as https://github.com/colbymchenry/codegraph/pull/1774.

Linux fail→pass + wasm/kernel coverage (`nested-declarator-functions`, `kernel-tsjs-parity`, router suites, `fuzzy-lexical-reach`) all green. Closing this PR as superseded by #1774.

colbymchenry:
Superseded by #1774 (rebased onto main after #1772).

PR #1690, opened 2026-09-04
fix(resolution): read a quoted Python annotation as a receiver type

### LINKED ISSUE #1684: Python: a quoted forward-reference annotation (o: "Alpha") drops the call edge that the identical unquoted annotation resolves
## Summary

Python: a parameter annotated with a **string (forward-reference) annotation** — `def f(o: "Alpha")` — loses receiver-type resolution that the identical unquoted annotation gets. The method call produces **no `calls` edge** at all.

Quoted annotations are ordinary Python (forward references, and everything under `from __future__ import annotations`), so this silently drops real edges in normal code.

## Environment
`@colbymchenry/codegraph` **1.5.0** and **1.6.0** (npm latest), Linux, fresh `codegraph init` per version.

## Repro

```python
# pkg/a.py
def render(x):
    return x


class Alpha:
    def render(self):
        return "a"


class Beta:
    def render(self):
        return "b"
```

```python
# pkg/b.py
from .a import Alpha


def quoted(o: "Alpha"):
    return o.render()      # -> NO EDGE

def bare(o: Alpha):
    return o.render()      # -> calls Alpha.render (pkg/a.py:6)
```

## Observed — identical on 1.5.0 and 1.6.0

```
caller   target                    line
bare     render @pkg/a.py:6          9
(quoted produces no row at all)
```

Same file, same call expression, same class — only the quoting of the annotation differs.

## Note on the trigger

The miss needs a competing same-named method to be present: with only `Alpha.render` (plus the top-level `render`) in the tree, the quoted form still resolved. Adding `Beta.render` is what makes the quoted form drop out while the bare form keeps resolving. So the quoted annotation appears not to establish the receiver type, leaving the call to name matching, which then declines under the extra candidate.

`codegraph callers render` correspondingly under-reports: `quoted` never appears.

## Verification
Reproduced on both versions with a fresh index each time. Controls in the same run: the unquoted form resolves, and in a two-candidate variant of the same fixture the quoted form resolves — so this is the quoting interacting with candidate count, not a blanket annotation failure.

danusha2345:
Fixes #1684.

## Problem

`def f(o: "Alpha")` is the same annotation as `def f(o: Alpha)` — a forward reference, and what every file under `from __future__ import annotations` writes — but the receiver-type pattern (`\bo\b\s*:\s*([A-Z][\w.]*)`) stopped at the quote, read no type, and `o.render()` produced no edge at all while the unquoted twin resolved to `Alpha::render`.

## Change

`buildLocalReceiverTypePatterns` (Python) admits the quoted form, `o: "Alpha"` / `o: 'pkg.Alpha'`, tried before the unquoted pattern. `resolveMethodOnType` still validates the method exists on that type, so nothing new can be guessed.

## Verification

New `__tests__/python-quoted-annotation.test.ts` with the issue's repro: quoted, single-quoted and unquoted annotations all resolve to the same method; it fails on `main` (`expected [] to deeply equal ['Alpha::render']`). Full suite green.


colbymchenry:
Thanks for the clean fix and regression — the approach is correct.

This PR was behind current `main`, and checking out `name-matcher.ts` wholesale would have dropped later resolution fixes (`NO_NESTED_FUNCTIONS` / #1729, lexical reachability / #1708–#1709). Landed the same surgical change + test on current `main` as #1770, which supersedes this PR. Closing in favor of that.

colbymchenry:
Superseded by #1770 (same fix rebased onto current main).

Issue #1681, opened 2026-09-04
Python: a top-level function named like a collection method (append/update/get) gets ZERO real callers and fabricated ones — uncovered shapes left by #715 and #1317, live in 1.6.0


inth3shadows:
## Summary

On Python, a top-level project function whose name collides with a common **collection-method** name (`append`, `update`, `get`, `close`, `run`, …) is wrong in **both directions at once**, and the two errors compound into a silently plausible answer:

- **False negative** — a real `module.append(x)` call through a plain (non-aliased) module import produces **no `calls` edge**. Its neighbour `module.path()` on the very next line resolves fine, so the loss is name-dependent, not import-dependent.
- **False positive** — an unrelated `.append(...)` on a local collection, reached through a **call-expression receiver** (`d.setdefault(k, []).append(v)`), degrades to a bare `append` ref and exact-matches the project's top-level `append` as the only same-named symbol.

Net effect: `codegraph callers append` returns a non-empty, confident-looking list in which **every entry is fabricated and every real caller is missing**. An empty result would be safer — it is visibly wrong. This one is not.

This is **not a regression**. It is the shape left uncovered by two shipped fixes:

- **#578 / #899 → #715** fixed plain module-attribute calls, but only for names that do *not* collide with a builtin collection method. `mod.path()` works; `mod.append()` does not.
- **#1230 → #1317** fixed bare-name fabrication for **literal** receivers (`", ".join(...)`) and nested-local scope. A **call-expression** receiver still degrades to a bare ref and still fabricates.
- Distinct from **#1626** (aliased module imports). Nothing here is aliased — `from . import ledger` then `ledger.append(...)`.

## Environment

- `@colbymchenry/codegraph` **1.5.0** and **1.6.0** (current npm latest) — reproduced identically on both
- Linux (WSL2), Node via the package's own `npm-shim.js`
- Fresh `codegraph init` per run; no stale index

## Minimal repro (12 lines)

```python
# pkg/ledger.py
def path():   return "/tmp/x"
def append(row):  return row     # top-level function named like a collection method
def join(vals):   return vals
```

```python
# pkg/uses.py
from . import ledger

def real_module_call(row):
    ledger.append(row)              # A  -> NO EDGE          (expected: calls ledger.append)
    return ledger.path()            # B  -> calls, "import"  (correct)

def literal_receiver(vals):
    return ", ".join(vals)          # C  -> no edge          (correct, #1317)

def plain_identifier(out, k):
[...]

PR #1692, opened 2026-09-04
fix(extraction): never fabricate an edge from a call-result receiver

### LINKED ISSUE #1683: Call-expression receivers drop the receiver and fabricate a calls edge to any same-named top-level symbol — Python AND JavaScript, live in 1.6.0
## Summary

A member call whose receiver is a **call expression** loses the receiver at extraction time, degrades to a bare method-name ref, and then exact-matches any top-level project symbol with that name — fabricating a call edge from an unrelated function.

Reproduced in **Python and JavaScript**, so this is the shared extractor, not a per-language heuristic.

Not the same as #1230/#1317 (literal receivers, fixed — control C below passes) or #645 (C++ chained getters binding to the wrong *class*; here the receiver is dropped entirely and the edge lands on a **top-level function**).

## Environment
`@colbymchenry/codegraph` 1.5.0 and 1.6.0 (npm latest), Linux, fresh `codegraph init` per run.

## Repro — Python

```python
# pkg/ledger.py
def append(row): return row
def path():      return "/x"
```
```python
# pkg/uses.py
from . import ledger

def plain_identifier(out, k):
    out.append(k)                   # C: plain ident      -> no edge   (correct)

def chained_receiver(d, k, v):
    d.setdefault(k, []).append(v)   # D: call-expr recv   -> calls ledger.append  WRONG
```

## Repro — JavaScript

```js
// src/lib.js
export function push(row) { return row; }
export function path() { return "/x"; }
```
```js
// src/use.js
import * as lib from './lib.js';

export function realNamespaceCall(row) {
  lib.push(row);              // A -> calls lib.push  "import"    (correct)
}
export function plainIdent(out, k) {
  out.push(k);                // C -> no edge                     (correct)
}
export function chained(m, k, v) {
  m.getOrInit(k, []).push(v); // D -> calls lib.push  "exact-match"  WRONG
}
```

## Observed — both languages, identical

| shape | receiver | edge |
|---|---|---|
| plain identifier (`out.append(k)`) | ident | none — correct |
| literal (`", ".join(v)`) | literal | none — correct (#1317) |
| namespace/module member | ident | resolves via `import` — correct |
| **call expression** (`d.setdefault(k,[]).append(v)`) | call | **fabricated `exact-match` edge to the top-level `append`/`push`** |

Only the receiver *shape* differs between the correct and incorrect rows.

## Impact

Any project with a top-level export named like a common collection method (`append`, `push`, `get`, `update`, `run`, `close`) accumulates fabricated inbound edges proportional to how often that method is used on chained locals elsewhere. `callers`/`impact` then return a populated, plausible list that is wrong — worse than an empty one, which is visibly wrong.

## Suspected mechanism

In the extractor, the qualified `receiver.method` shape is preserved only when the receiver is a plain identifier; a call-chain receiver degrades to a bare `method` ref. That bare ref passes the codebase-wide name check (a real top-level symbol of that name exists) and is exact-matched as the sole candidate.

## Verification
Both versions, both languages, fresh index each time; cross-checked via the `codegraph` CLI and direct SQL against `.codegraph/codegraph.db`, which agree.

danusha2345:
Fixes #1683. Also covers the false-positive half of #1681.

## Problem

A member call whose receiver is itself a call — `d.setdefault(k, []).append(v)`, `d.items().get(k)`, `make().run()` — lost the receiver at extraction time and degraded to the bare method name. The bare name then exact-matched (or, via the fuzzy split, fuzzy-matched) any top-level project symbol of that name — a `calls` edge from an unrelated function, reproduced in Python and JavaScript alike, so it is the shared extractor, not a per-language rule.

## Change

- **Extraction** (wasm walker, `tsjs/extractors.rs`, `python.rs`): for TS/JS/Python a call receiver is kept and encoded as `<inner>().<method>` — the marker the Java / Kotlin / C++ chains already use, which no plain name can match. An inner callee with no static name (`(await x)()`, `arr[0]()`) emits nothing: a silent miss, never a wrong edge. The inner call is visited on its own either way, so `make()` keeps its edge.
- **Resolution**: the chain shape is routed past `resolveViaImport` (which bound `useStore.getState().reset` to the imported store constant, so the method was never looked up) and stops before the fuzzy strategy for these three languages. The one fallback kept is the store-accessor idiom the object-literal coverage relies on — Zustand's `get()` inside the factory and `useStore.getState()` outside it, followed by a method name with exactly one callable in the project. Everything else resolves to nothing: what an inner call returns is not knowable from its name (an earlier draft that allowed any project-rooted chain bound `db.prepare(sql).all()` to a project function named `all` on this repo's own index — 363 such edges — which is the very defect).
- Parity fixtures: `torture.js` and `torture.py` gain the shapes.

## Verification

- New `__tests__/call-receiver-no-fabrication.test.ts`: the issue's Python and JS repros produce no edge into the same-named top-level functions, the inner call keeps its edge, the encoding is as stated, a receiver with no static callee is dropped. On `main` the fabricated `append` / `get` / `run` callers appear.
- `object-literal-methods` (Zustand `get().reset()` and `useStore.getState().reset()`) still resolves; its extraction assertion is updated to the chain-encoded ref.
- Full suite with the rebuilt kernel: 237 files, 4234 passed, 9 skipped.
- On real repositories (a TS/Dart app, a Kotlin+Go+TS app, this repo, a 4k-file C tree): cross-file `calls` edges landing on top-level functions named like collection methods drop from 22 → 5, 31 → 20, 2 → 1; C unchanged.

Re-index after upgrading.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -201,6 +201,8 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 #### Symbols, tests and the viewer
 
+- A method called on the result of another call — `d.setdefault(k, []).append(v)`, `make().run()` — no longer produces a call edge to an unrelated top-level function that merely shares the name, in Python and JavaScript/TypeScript. The receiver is kept so the inner call still resolves; the outer method stays unresolved rather than guessed. Re-index after upgrading. (#1683, #1681)
+
 - **Files under an `e2e/` directory count as tests.** Their calls no longer appear as production callers in Steps, dead-code and test badges.
 
 - **Production code under a `samples` or `examples` package path is no longer treated as test code.** A Kotlin or Java project whose package path runs through `com/google/samples/…` (Now in Android, for one) had nearly every file counted as a fixture, so the Map opened on `build-logic`, the entry points hid the app, and dead-code and test badges were wrong. Only the project layout above a `src/` folder decides now; the package path below it never does.
--- a/__tests__/call-receiver-no-fabrication.test.ts
+++ b/__tests__/call-receiver-no-fabrication.test.ts
@@ -1,0 +1,81 @@
+/**
+ * A member call whose receiver is itself a call never fabricates an edge
+ * (#1683, #1681). `d.setdefault(k, []).append(v)` used to lose its receiver at
+ * extraction time, degrade to the bare `append`, and exact-match any top-level
+ * project function of that name — a call edge from an unrelated function,
+ * reproduced in Python and JavaScript alike. The receiver is now kept as
+ * `<inner>().<method>`, which nothing name-matches; the inner call resolves
+ * on its own as before.
+ */
+import { describe, it, expect, beforeAll, afterAll } from 'vitest';
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+import { CodeGraph } from '../src';
+import { extractFromSource } from '../src/extraction';
+import { initGrammars, loadAllGrammars } from '../src/extraction/grammars';
+
+let dir: string;
+let cg: CodeGraph;
+
+beforeAll(async () => {
+  await initGrammars();
+  await loadAllGrammars();
+  dir = fs.mkdtempSync(path.join(os.tmpdir(), 'cg-1683-'));
+  fs.mkdirSync(path.join(dir, 'py'));
+  fs.mkdirSync(path.join(dir, 'js'));
+  fs.writeFileSync(path.join(dir, 'py', '__init__.py'), '');
+  fs.writeFileSync(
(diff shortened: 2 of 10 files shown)


colbymchenry:
Thanks for this — the approach is correct and the regression coverage is solid.

#1692 was dirty against current `main` (CHANGELOG conflict with #1731/#1719 after #1746). I verified the fix on Linux (issue fixtures + your new tests), rebased onto latest main, rebuilt the kernel, and opened #1748 to land it. Please see that PR; this one can be closed as superseded once #1748 merges.

danusha2345:
Superseded by #1748, which landed this approach on `main` (thanks for carrying it over). Closing.

PR #1695, opened 2026-09-04
fix(prompt-hook): cap the injection under Claude Code's 10k inline hook-output limit

### LINKED ISSUE #1694: `prompt-hook` injection cap (16k chars) exceeds Claude Code's 10k inline hook-output limit, so the context is persisted to a file and the model sees a 2 KB preview
## Summary

`codegraph prompt-hook` caps its injected `<codegraph_context>` block at `MAX = 16000` characters (`src/bin/codegraph.ts`, "Cap the injection so a large-repo explore can't flood the prompt"). Claude Code persists any hook stdout over **10,000 characters** to a file and shows the model a 2 KB preview instead. So on a repo where explore fills the cap, the hook delivers a file path and the first 2 KB, not the context.

Claude Code's limit is not documented; I measured it (below). The fix is one constant.

## What the model sees

From a Claude Code transcript with the installed hook, on every prompt of a 6-cell run (hook stdout 14.9–15.8 KB each time):

```
<persisted-output>
Output too large (15.8KB). Full output saved to: .../tool-results/hook-<id>-stdout.txt

Preview (first 2KB):
<codegraph_context note="Structural context from CodeGraph for this prompt — treat returned source as already read; call codegraph_explore for more.">
**Exploration: ...**

Found 47 symbols across 1 file.

**Blast radius — what depends on these (update/verify before editing)**
...
</persisted-output>
```

The model then either Reads the persisted file (one extra tool call per prompt) or answers from the preview.

## Measured threshold

Claude Code 2.1.261, headless `claude -p`, a stub `UserPromptSubmit` hook that prints N characters:

| hook stdout (chars) | delivered               |
| ------------------- | ----------------------- |
| 8,031               | inline                  |
| 8,991               | inline                  |
| 9,631               | inline                  |
| 9,871               | inline                  |
| 10,031              | persisted, 2 KB preview |
| 12,031              | persisted               |
| 16,031              | persisted               |

So the inline limit is 10,000 characters, and a payload at the hook's own 16k cap is always persisted.

## Effect

Same 6 discovery tasks, Opus, headless Claude Code, repo rules identical, paired cells:

| setup                            | tool calls / task | `codegraph_explore` calls / task | correct |
| -------------------------------- | ----------------: | -------------------------------: | ------: |
| MCP server only (no hook)        |              3.83 |                             1.33 |     6/6 |
| `codegraph install` (hook + MCP) |              4.33 |                             0.83 |     6/6 |

With the hook the tool was called less, not more: the persisted preview is treated as "context already delivered" and the follow-up explore the note asks for often does not happen.

Related: #1654 notes that a keyword hit "injects the full explore payload unconditionally". This is what that payload becomes on the Claude Code side once it exceeds 10 KB.

## Proposed fix

`MAX = 9000` (leaves headroom under 10,000 for the wrapper and the `projectPath` nudge lines), keeping the existing "…(truncated; call codegraph_explore for the rest)" notice. The blast-radius list already leads the payload, so the surviving 9 KB is the part the model uses. Optionally read the cap from an env var for other hosts. Happy to send the one-line PR if that direction is right.

## Environment

- codegraph v1.6.0 (npm global install, Windows 11, bundled Node 24.16.0)
- Claude Code 2.1.261
- Repo: ~450 TS/JS files

bompus:
Fixes #1694.

`codegraph prompt-hook` capped its `<codegraph_context>` injection at 16,000 characters. Claude Code shows hook stdout inline only up to 10,000 characters; above that it persists the output to `tool-results/hook-<id>-stdout.txt` and the model sees a 2 KB preview plus the path. So on any repo where explore filled the cap, the hook delivered a file reference instead of the context (measured in #1694: 9,871 characters inline, 10,031 persisted, on Claude Code 2.1.261).

## Change

- `MAX` 16000 → 9000 in `src/bin/codegraph.ts`, with a comment naming the host limit. 9,000 leaves ~1,000 characters for the `<codegraph_context>` wrapper and the `projectPath` nudge lines that are appended after the cap is applied.
- CHANGELOG entry under Unreleased › Fixes.

The existing "…(truncated; call codegraph_explore for the rest)" notice is unchanged, so a capped payload still tells the model where the rest is. The blast-radius list leads the explore output, so the part that survives the cap is the part the model acts on.

## Not in this PR

An env var to set the cap per host. Happy to add one if you want other hosts to be able to raise it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -232,6 +232,8 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 - **The Map covers a multi-root project.** A React Native app's `ios/` beside its `src/` — or any second root holding a fifth of the code — is now on the picture, one level deeper, instead of the map silently drawing only the larger root.
 
+- **The Claude Code prompt hook's context now arrives inline.** The hook capped its injection at 16,000 characters, but Claude Code shows hook output inline only up to 10,000 and otherwise persists it to a file with a 2 KB preview, so on any repo where explore filled the cap the model saw a file path and the first 2 KB. The cap is now 9,000 characters, under the limit with room for the wrapper. (#1694)
+
 - Fixed a long-running `codegraph ui` session serving a symbol that a sync had already deleted. The viewer keeps one connection to your index open, and its in-memory lookup didn't notice when another process — your agent's sync, or `codegraph sync` — rewrote the file underneath it, so a symbol screen could keep showing a body with no callers while search correctly reported it had moved. Because a symbol's identity includes the line it starts on, this happened after almost any edit above it.
 
 ## [1.6.0] - 2026-08-26
--- a/src/bin/codegraph.ts
+++ b/src/bin/codegraph.ts
@@ -1412,7 +1412,11 @@ program
             const text = result.content[0]?.text ?? '';
             if (!result.isError && text.trim()) {
               // Cap the injection so a large-repo explore can't flood the prompt.
-              const MAX = 16000;
+              // Claude Code shows hook stdout inline only up to 10,000 characters;
+              // above that it persists the output to a file and the model sees a
+              // 2 KB preview (#1694). 9,000 leaves room for the wrapper and the
+              // projectPath nudge lines below.
+              const MAX = 9000;
               const body = text.length > MAX ? `${text.slice(0, MAX)}\n…(truncated; call codegraph_explore for the rest)` : text;
               // For a front-loaded SUB-project, a follow-up explore needs its path.
               const more = plan.viaSubScan


danusha2345:
Confirmed the symptom and the fix on a live setup. The installed `prompt-hook` (1.6.0-era build, 787-file TS repo) returned 16,360 characters on ordinary prompts — every session got the persisted-file preview, exactly as #1694 describes. With this change merged into a local integration build of current `main` the same prompt returns 9,348 characters and lands inline. Full suite 252 files / 4349 passed with it in.

bompus:
Measured this one end to end on a real repo (vite, indexed at 9,354 nodes / 27,778 edges) rather than a fixture. It holds up.

**The premise is documented, not inferred.** From the Claude Code hooks reference:

> Hook output strings, including `additionalContext`, `systemMessage`, and plain stdout, are capped at 10,000 characters. Output that exceeds this limit is saved to a file and replaced with a preview and file path.

So the old `MAX = 16000` was above the cap by design, and every prompt that filled the budget got the file-and-preview treatment — which is exactly the failure mode where the model sees a 2 KB stub instead of the context.

**Measurement.** I drove `codegraph prompt-hook` directly (JSON `{prompt, cwd}` on stdin) against vite with five structural prompts, and counted stdout characters:

| arm | stdout chars | over the 10,000 cap |
| --- | --- | --- |
| base (`MAX = 16000`) | 16,223 on all five prompts | 5 / 5 |
| this PR (`MAX = 9000`) | 9,223 on all five prompts | 0 / 5 |

Both arms are pinned at the ceiling on every prompt, which is the point — on a repo this size the budget is always the binding constraint, so the old value wasn't an occasional overflow, it was overflowing every time.

The 223-char delta over `MAX` in both arms is the wrapper plus the `projectPath` nudge lines, and it's constant. That leaves **777 characters of headroom** under the cap, which is enough room for the wrapper to grow without another silent truncation.

No concerns. LGTM.

bompus:
Rebased onto `cd4e65b` (current `main`) — no conflicts. New head `f6f1f1d`.

Checks on the rebased branch: `tsc --noEmit` clean; the prompt-hook tests pass (45 tests, 3 files); `installer-targets` passes; `cli-context-command` passes 6/6 once `dist` is built.

colbymchenry:
Superseded by #1769, which lands the same MAX 16000→9000 change on current `main` (this branch was CONFLICTING on CHANGELOG only) and adds an exported `capPromptHookInjection` / constant unit tests. Closing in favor of that merge — thank you for the clear repro and the original PR.

colbymchenry:
Closing as superseded by merged #1769.

Issue #1279, opened 2026-07-13
Claude never calls codegraph


crearo:
Claude never uses codegraph.
I've specified in claude.md.
I've even put a hook for tool use requesting claude to use codegraph instead of plain grep.
It still uses grep.

When I asked it, why:

"So on the turns where that block reached me (the hook), I under-weighted it and reached for grep anyway. That one's on me, and I won't dress it up"

<img width="1699" height="124" alt="Image" src="https://github.com/user-attachments/assets/33d04688-d395-4d4b-99ca-bd971247ba81" />

qoder1337:
same for me. pretty sure it started to occur since sonnet 5 came up.

Jotalz:
same issue.Claude shows that the MCP codegraph is connected, and the claude `settings.json` contains the following permissions configuration:
```json
"permissions": {
"allow": [
"mcp__codegraph__codegraph_explore",
"mcp__codegraph__codegraph_search",
"mcp__codegraph__codegraph_node",
"mcp__codegraph__codegraph_callers",
"mcp__codegraph__codegraph_callees",
"mcp__codegraph__codegraph_impact",
"mcp__codegraph__codegraph_files",
"mcp__codegraph__codegraph_status"
]
}
```
Reinstalling codegraph does not resolve the issue.

devenyang000:
This is the real problem and seems no one cares about this. The tool maybe good, but as I can tell in my real coding work, agents rarely call it

GeroVanMi:
Edit: Superseded by a better solution in the follow-up comment.


A few issues that I've been able to identify regarding this (time of writing: 08.09.2026):

1. The Claude Code harness does not list MCP tools actively. Models have to search for them and they don't do so unless the user prompts them to do so actively. (`CLAUDE.md` instructions seem to be ignored, even when the explicit ToolSearch call is written there, though I may also have made an error)
2. The anthropic models just seem to have a very **very** strong preference for using `grep` (see also original message by @crearo). Even when I install a one-time hook to force the first search call to attempt using codegraph, the model at one point just straight-up ignored it and used `grep` again. (see example below)

> User:
>  Look at the transcript here. Why wasn't codegraph used for the search?

> Assistant: 
[...]

PR #1697, opened 2026-09-04
fix(mcp): keep codegraph_explore loaded in Claude Code and Copilot CLI

### LINKED ISSUE #1696: Claude Code defers `codegraph_explore` behind ToolSearch, so most sessions never call it; the tool and the installer should opt out (`alwaysLoad`)
## Summary

Claude Code defers every MCP tool behind its `ToolSearch` step by default: a fresh session sees only the tool's name, not its schema, until the model searches for it. So the server instructions' "call `codegraph_explore` instead of Read" arrive with nothing loaded to act on. Measured on a default `codegraph install` setup (no project rules naming the tool), 4 of 6 sessions never called `codegraph_explore` at all and did the task with Grep/Read; with the schema in context from the first turn, 6 of 6 did.

| 6 code-discovery tasks, Opus, headless, no project rules | sessions that called `codegraph_explore` | tool calls / task | cost / task | correct |
| -------------------------------------------------------- | ---------------------------------------: | ----------------: | ----------: | ------: |
| deferred (as installed)                                  |                                      2/6 |              4.83 |       $0.46 |     6/6 |
| schema loaded, tool search off                           |                                      6/6 |              3.67 |       $0.38 |     6/6 |

Same entry both rows; the second row runs the session with tool search disabled, which puts the schema in context the same way `alwaysLoad` does. The four deferred sessions that skipped the tool ran 9, 3, 7 and 4 Grep/Read/Bash calls; the tool was in their list the whole time, by name only.

Claude Code has two opt-outs, and CodeGraph sets neither:

- `"alwaysLoad": true` on the server entry in `.mcp.json` / `~/.claude.json` (docs: https://code.claude.com/docs/en/mcp, "Exempt a server from deferral").
- `_meta: { "anthropic/alwaysLoad": true }` on the tool definition itself, which works on existing installs without touching the user's config.

## Reproduction

Claude Code 2.1.261, Opus, headless, `--strict-mcp-config` with only `{"mcpServers":{"codegraph":{"command":"codegraph","args":["serve","--mcp"]}}}` (what `codegraph install` writes today), prompt "Use codegraph_explore to find which function defines clearExtensionStorage":

| server                                       | tool calls in order               |
| -------------------------------------------- | --------------------------------- |
| 1.6.0 as installed                           | `ToolSearch`, `codegraph_explore` |
| same, `"alwaysLoad": true` on the entry      | `codegraph_explore`               |
| same config, tool carries `_meta` alwaysLoad | `codegraph_explore`               |

That prompt names the tool outright, so the model searches for it. The tasks in the summary table do not name it, which is the normal case, and there the deferred session mostly never searches. Of the two deferred sessions that did use the tool, one searched first and one called it by name unsearched and got through, so the search turn is usual, not required.

With a project rule that names the tool (our own repo's `CLAUDE.md`), 4 of 6 sessions used it with or without the key, so a rule recovers most of the adoption. A default install has no such rule.

## Proposed fix

Both, so existing installs and new ones are covered:

1. `codegraph_explore` carries `_meta: { "anthropic/alwaysLoad": true }` in its tool definition. Claude Code reads it; other hosts ignore `_meta`.
2. The Claude Code installer target writes `"alwaysLoad": true` on the server entry (Claude only, since the key is host-specific), and the README's manual `~/.claude.json` snippet shows it.
3. Same shape for GitHub Copilot CLI: its tool search is on by default and defers MCP tools once ~30 tools are connected (Claude and GPT-5.4+ models), and the per-server opt-out is `"deferTools": "never"` in `~/.copilot/mcp-config.json` ([docs](https://docs.github.com/en/copilot/concepts/agents/copilot-cli/tool-search)). Not measured, docs only.

Checked the other installer targets' docs for the same thing: Codex, Cursor, Antigravity, Kiro and Hermes load MCP schemas eagerly by default (Kiro's and Hermes's tool search are opt-in, with no per-server pin), so nothing to set there. OpenCode 2 is a different case and gets its own issue: it puts MCP tools behind Code Mode by default, and the per-server opt-out `"codemode": false` only survives on its native v2 entry shape, which the installer does not write yet.

PR incoming.

bompus:
Fixes #1696.

Claude Code defers every MCP tool behind its `ToolSearch` step by default, so a fresh session sees only `codegraph_explore`'s name until the model searches for it. On a default `codegraph install` setup with no project rule naming the tool, 4 of 6 headless Opus sessions never called it and did the task with Grep/Read; with the schema in context from the first turn, 6 of 6 did (numbers and the reproduction in #1696).

## Change

- `codegraph_explore` carries `_meta: { "anthropic/alwaysLoad": true }` in `src/mcp/tools.ts`, and `ToolDefinition` gains an optional `_meta`. Claude Code reads the key and keeps the tool's schema loaded from the first prompt. It reaches an existing install on upgrade without touching the user's config; other hosts ignore `_meta`.
- The Claude Code installer target writes `"alwaysLoad": true` on the server entry (`src/installer/targets/claude.ts`), on both the project `.mcp.json` and the global `~/.claude.json` path. Re-running `codegraph install` adds the key to an entry that predates it. The README's manual `~/.claude.json` snippet shows the key.
- The Copilot CLI target writes `"deferTools": "never"` on its entry (`src/installer/targets/copilot-cli.ts`). Copilot CLI's tool search defers MCP tools once ~30 tools are connected, and that key is its per-server opt-out ([docs](https://docs.github.com/en/copilot/concepts/agents/copilot-cli/tool-search)). Docs-verified only; the host is not installed here.
- Tests: `mcp-tool-annotations.test.ts` asserts the `_meta` on every `tools/list` surface (the master array, `getStaticTools()`, `ToolHandler.getTools()`, and the live server); `installer-targets.test.ts` covers the key on the Claude project and global paths, the Copilot CLI entry, and the re-run migration for both.
- CHANGELOG entry under Unreleased › Fixes.

## Verified

- Headless Claude Code 2.1.261, Opus, `--strict-mcp-config` with the plain entry: the released 1.6.0 server runs `ToolSearch` then `codegraph_explore`; this branch's build runs `codegraph_explore` alone.
- `vitest`: 233 passed, 3 skipped. `tsc` clean.

## Not in this PR

OpenCode 2's Code Mode, which puts MCP tools behind a code interface by default. Its per-server opt-out, `"codemode": false`, only survives on the native v2 entry shape, which the installer does not write yet. Separate issue to follow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -232,6 +232,8 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 - **The Map covers a multi-root project.** A React Native app's `ios/` beside its `src/` — or any second root holding a fifth of the code — is now on the picture, one level deeper, instead of the map silently drawing only the larger root.
 
+- **`codegraph_explore` is loaded from the first prompt in Claude Code.** Claude Code defers every MCP tool behind a tool-search step, so a fresh session saw only the tool's name until the model searched for it, and the server's "call `codegraph_explore` instead of Read" had nothing loaded to act on. The tool now carries `anthropic/alwaysLoad` in its `_meta`, which exempts it on existing installs, and `codegraph install` writes `alwaysLoad: true` on the Claude Code server entry (re-run it to add the key). Copilot CLI's tool search holds MCP tools back the same way once ~30 tools are connected, so its entry now carries `deferTools: "never"`. (#1696)
+
 - Fixed a long-running `codegraph ui` session serving a symbol that a sync had already deleted. The viewer keeps one connection to your index open, and its in-memory lookup didn't notice when another process — your agent's sync, or `codegraph sync` — rewrote the file underneath it, so a symbol screen could keep showing a body with no callers while search correctly reported it had moved. Because a symbol's identity includes the line it starts on, this happened after almost any edit above it.
 
 ## [1.6.0] - 2026-08-26
--- a/README.md
+++ b/README.md
@@ -508,11 +508,14 @@ npm install -g @colbymchenry/codegraph
     "codegraph": {
       "type": "stdio",
       "command": "codegraph",
-      "args": ["serve", "--mcp"]
+      "args": ["serve", "--mcp"],
+      "alwaysLoad": true
     }
   }
 }
 ```
+
+`alwaysLoad` keeps `codegraph_explore` loaded from the first prompt. Claude Code otherwise defers every MCP tool behind a tool-search step, so a fresh session sees only the tool's name until the model searches for it.
 
 **Add to `~/.claude/settings.json` (optional, for auto-allow):**
 ```json
--- a/__tests__/installer-targets.test.ts
+++ b/__tests__/installer-targets.test.ts
@@ -976,6 +976,22 @@ describe('Installer targets — partial-state idempotency', () => {
     expect(fs.existsSync(path.join(tmpCwd, '.claude.json'))).toBe(false);
     const cfg = JSON.parse(fs.readFileSync(path.join(tmpCwd, '.mcp.json'), 'utf-8'));
     expect(cfg.mcpServers.codegraph).toBeDefined();
+    // Exempt from Claude Code's tool-search deferral (#1696).
+    expect(cfg.mcpServers.codegraph.alwaysLoad).toBe(true);
+  });
+
+  it('claude: re-running install on an entry that predates alwaysLoad adds the key (#1696)', () => {
+    const claude = getTarget('claude')!;
+    fs.writeFileSync(
+      path.join(tmpCwd, '.mcp.json'),
+      JSON.stringify({ mcpServers: { codegraph: { type: 'stdio', command: 'codegraph', args: ['serve', '--mcp'] } } }, null, 2),
+    );
+    const result = claude.install('local', { autoAllow: false });
+    const mcp = result.files.find((f) => f.path.replace(/\\/g, '/').endsWith('/.mcp.json'));
+    expect(mcp?.action).toBe('updated');
+    const cfg = JSON.parse(fs.readFileSync(path.join(tmpCwd, '.mcp.json'), 'utf-8'));
+    expect(cfg.mcpServers.codegraph.alwaysLoad).toBe(true);
+    expect(cfg.mcpServers.codegraph.args).toEqual(['serve', '--mcp']);
   });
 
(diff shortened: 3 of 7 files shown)


danusha2345:
Merged into a local integration build of current `main` (one trivial conflict in `targets/claude.ts` against #1029's `CLAUDE_CONFIG_DIR` helper — both functions kept). Verified over stdio: `tools/list` now returns `codegraph_explore` with `_meta: {"anthropic/alwaysLoad": true}`, so an existing install picks it up on upgrade without touching `~/.claude.json`. `installer-targets` and `mcp-tool-annotations` green; full suite 252 files / 4349 passed.

bompus:
Verified both halves of this against first-party docs and then over the wire. It's correct.

**Key names check out.** `alwaysLoad` is the documented Claude Code MCP key, including the per-tool `_meta` form this PR emits (`"anthropic/alwaysLoad": true`), and `deferTools` is documented for Copilot CLI with exactly the `"auto" | "never"` domain the PR uses. Worth stating explicitly because these are the kind of keys that get copied from the wrong guide and then silently do nothing — a typo'd key is indistinguishable from a working one at the config layer.

**Measured over stdio,** not read off the source: I spoke MCP to the built server (`initialize` → `notifications/initialized` → `tools/list`) against a real indexed repo and inspected the advertised tool descriptors.

| arm | `codegraph_explore` `_meta` |
| --- | --- |
| base | `null` |
| this PR | `{"anthropic/alwaysLoad": true}` |

So the flag actually reaches the client in the tool listing, which is the only place it can matter. That's the whole claim, and it's confirmed at the protocol boundary rather than at the config file.

LGTM.

bompus:
Rebased onto `cd4e65b` (current `main`) — no conflicts. New head `097cd19`.

Checks on the rebased branch: `tsc --noEmit` clean; `installer-targets` and `mcp-tool-annotations` both pass.

GeroVanMi:
Also addresses #1279

colbymchenry (approved):
Verified on Linux against current main (cherry-pick clean):

- Fail→pass: with production alwaysLoad/_meta/deferTools removed, mcp-tool-annotations (4) and the #1696 installer assertions fail; with the PR changes restored they pass.
- Full `__tests__/installer-targets.test.ts`: 234 passed, 3 skipped.
- `__tests__/mcp-tool-annotations.test.ts`: 4 passed (explore `_meta` on master array, getStaticTools, withRequiredProjectPath, and live getTools).

LGTM — landing this for #1696.

PR #1704, opened 2026-09-05
fix(resolution): Python module-qualified calls colliding with builtin methods

inth3shadows:
## Problem

Two Python call-resolution bugs with one root cause: the method-call heuristics assume a common collection-method name always means a builtin, and never account for a project module exporting a function of that name.

**1. Real calls silently dropped.** `ledger.append(row)` — where `ledger` is a project module exporting a top-level `append` — was classified as `list.append` by `isBuiltInOrExternal` and discarded. The only escape hatch was a capitalized receiver matching a known class, so a module receiver never qualified. The ref never reached `resolveViaImport` / `resolvePythonModuleMember`, which already resolve it correctly. `ledger.append`'s actual callers went uncounted.

**2. Wrong edges fabricated.** A method call through a non-identifier receiver — an attribute chain (`self.data.append(x)`), a subscript (`d[k].append(x)`), a call chain (`rows.setdefault(k, []).append(x)`) — degraded at extraction time to a bare `append` ref. That bare ref then exact-matched an unrelated top-level `append` as the sole same-named symbol project-wide, inventing a call edge between functions with no relationship.

## Fix

- `src/resolution/index.ts` — before declaring a qualified call a builtin, check whether the receiver is an imported module in that file (`getImportMappings`). If it is, let it through to import resolution.
- `src/extraction/tree-sitter.ts` — for Python, keep the receiver's source text as a qualifier instead of collapsing an unresolvable receiver shape to a bare method name. An unresolved qualifier is then a silent miss, never a wrong edge.

Same philosophy as #1230 / #1276: prefer a missing edge over a fabricated one.

## Tests

`__tests__/resolution.test.ts` — one test covering both directions: the module-qualified call resolves to the module's function, and the chained-receiver call does *not* attach to it.

```
resolution.test.ts   190 passed
extraction.test.ts   622 passed
```

## Provenance

Found by running [testgraph](https://github.com/inth3shadows/testgraph)'s trace-derived ground-truth comparison against its own codebase — the Python graph showed `ledger.append` with zero callers while the runtime trace showed several, and showed callers it did not have (inth3shadows/testgraph#66).

CHANGELOG entry added under `[Unreleased]` → Fixes.


danusha2345:
Two notes from merging this into a local integration build of current `main`:

1. **The extractor half is dead once the native kernel is present.** Python extraction runs in `codegraph-kernel/src/python.rs` whenever the `.node` is staged (every published bundle ships it), so a change made only in `src/extraction/tree-sitter.ts` never reaches runtime and the two arms diverge — the same trap #1518 / #1521 fell into. Your own `rows.setdefault(k, []).append(x)` test passes in that build only because #1692 already keeps call receivers in **both** arms (`<inner>().<method>`, TS/JS/Python, `tsjs/extractors.rs` + `python.rs`, parity fixtures). The broader receiver shapes you cover (`self.data.append`, `d[k].append`) would need the same mirror in `python.rs` plus a line in `__tests__/fixtures/kernel-parity/torture.py`.

2. **The resolver half is the part #1692 does not have, and it is good.** `isBuiltInOrExternal` letting a receiver through when it is an imported module of the caller's file (`ledger.append` → `resolveViaImport` / `resolvePythonModuleMember`) closes the false-negative side of #1681. I took exactly that hunk plus your tests into the integration build (`resolution` 208 → 262 passed, no regressions); the extractor hunk was dropped as above.

Suggestion so the maintainer does not get two competing PRs for one bug: keep this PR to the resolver half (the `ledger.append` recall fix, which stands on its own and merges clean), and let #1692 carry the receiver encoding for both arms. Happy to cross-reference either way.

inth3shadows:
Thanks — point 1 is correct, and I verified it rather than taking it on trust. It turned out to be a stronger argument for **mirroring** the extractor hunk than for dropping it, so I've pushed the mirror (2c5319c) instead of narrowing the PR.

### The bug reproduces on a published bundle

Installed `@colbymchenry/codegraph` 1.5.0 (which ships `codegraph-linux-x64/lib/kernel/codegraph-kernel.node`), three files, `codegraph index`, then reading the `edges` table directly:

```
build_map (unrelated.py) --calls--> append (ledger.py)   <- fabricated (self.data.append)
build_map (unrelated.py) --calls--> append (ledger.py)   <- fabricated (rows["k"].append)
add_outcome --calls--> ledger.append                     <- MISSING
```

So both halves of #66 are live in the shipped build, which is exactly your point: a fix in `src/extraction/tree-sitter.ts` alone never runs there.

### #1692 covers one of the three shapes

`python.rs`'s `extract_call` filters the receiver to `identifier | simple_identifier | field_identifier` and sends everything else to a bare `method_name`. #1692 adds one branch for `receiver.kind() == "call"`. That fixes `d.setdefault(k, []).append(x)` — but an **attribute chain** (`self.data.append`) and a **subscript** (`d[k].append`) still fall through to the bare name, and the bare name is what exact-matches an unrelated project function.

### The parity suite already fails without the mirror

`__tests__/fixtures/kernel-parity/torture.py` line 26 is `self.registry.lookup("x")` — the attribute-chain shape. With this PR's TS hunk applied and no kernel mirror, `kernel-tsjs-parity` fails on it:

```
wasm:   referenceName "self.registry.lookup"
kernel: referenceName "lookup"
```

That failure is invisible in a from-source checkout: with no `.node` staged, every `kernel-*-parity` suite `describe.skipIf`s itself and the extraction tests silently exercise the wasm arm. I only saw it after installing Rust and running `scripts/build-kernel.sh`.

### What 2c5319c does

- Mirrors the branch into `codegraph-kernel/src/python.rs` for **all** non-identifier receivers, with a `collapse_js_whitespace` helper rather than `char::is_whitespace` — the two sets differ (U+0085 in one, U+FEFF in the other) and the parity sweep compares arms byte for byte.
- Adds the subscript and call-chain shapes to `torture.py`.
- Adds `__tests__/kernel-python-call-fabrication.test.ts`, which forces `CODEGRAPH_KERNEL_LANGS=python` and asserts the end-to-end property on the arm that ships, skipping when no `.node` is staged.

Verified by rebuilding the kernel both ways: `kernel-tsjs-parity` 17/17 with the mirror, 2 failures without it; the new suite passes against a freshly built kernel.

Happy to defer the call-chain shape to #1692 and keep this PR to the attribute-chain/subscript shapes plus the resolver half, if that avoids overlap — the two changes are compatible either way, since #1692's branch is checked before the general one.

inth3shadows:
**Heads-up, and an apology: the resolver hunk you took into your integration build has a regression. Fixed in b34d85f, but please re-check that build.**

A code-review pass over the em-tagged build of this branch caught it, and I reproduced it before believing it.

### The regression

`extractPythonImports` emits an `ImportMapping` for *every* import — stdlib and PyPI included — so `imp.localName === receiver` was true for `os`, `requests`, `np`. That opened the built-in-method filter for them; `resolveViaImport` then found no project file, resolution fell through to the bare-name strategy, and the call bound to whatever project method shares the name.

A project with `class Store: remove(), get()` and a file containing only `import os` / `import requests`:

```
this branch @2c5319c   cleanup -> Store.remove   refName "os.remove"
                       cleanup -> Store.get      refName "requests.get"
installed 1.6.0        (0 edges)
```

Two fabricated edges where the released build produced none — the exact class this PR set out to remove, arriving through its own escape hatch.

### The fix (b34d85f)

The escape now resolves the import specifier and opens only when it names a file **in this project** — the same question `resolveViaImport` asks next, so a receiver that passes is one the qualified path can actually serve. `from . import mod` and `import pkg.mod as m` are both handled; anything else stays a silent miss.

`__tests__/python-import-gate.test.ts` pins both directions and fails on the first without the change (`['remove@store.py','get@store.py']` vs `[]`). resolution + extraction + kernel parity: 832 tests, all pass.

### You were also right that my tests were vacuous for the attribute-chain shape

I checked this properly. `ledger.append` is a top-level `function`, and Strategy 3 only considers `method` kinds — so the negative assertion in my test could never have failed. With a **method** decoy (`class Sink: def append(self, x)`) the picture is:

| site | released 1.6.0 | this branch |
|---|---|---|
| `self.data.append(1)` | wrong edge · `exact-match` 0.9 | wrong edge · `instance-method` 0.7 |
| `rows[k].append(2)` | wrong edge · `exact-match` 0.9 | **gone** |
| `self.inner.get(k)` | wrong edge · `exact-match` 0.9 | wrong edge · `instance-method` 0.7 |

So the extraction hunk is a strict improvement here — three wrong edges become two, and the survivors drop from the top confidence tier to 0.7 with an honest `refName` — but it is **not** the fix its own comment claims. `self.data.append` still reaches the bare-name fallback, because Python has no exclusive chained-receiver branch equivalent to `matchGoFieldChainCall` / `matchRustSelfFieldCall`. Only receivers containing non-word characters (subscript, call chain) get the silent-miss behaviour.

That gap is real and I have not fixed it here — it needs a Python chained-receiver resolution branch, which is a bigger change than this PR should carry.

### Where that leaves the split you proposed

Still happy to go either way. If you take the resolver half, please take **b34d85f with it** — e799cd0 alone is the regression above.

inth3shadows:
**b964128 closes the attribute-chain gap I flagged in my last comment.** It is the discipline Go (#1276), Rust (#1585) and PHP's `this->prop.method` already have in `matchMethodCall`: a dotted python receiver resolves through validated inference or not at all, never through the bare-name strategies.

### It costs recall, and I measured rather than assumed

Indexing the tracked `.py` of four real projects and diffing the `calls` edges. Three unchanged (278 / 643 / 301). A 249-file one: **2605 → 2564**.

| dropped | |
|---|---|
| 21 | a dict `.update` bound to a service's `update` |
| 16 | application code bound to a `get` defined in a **test** file |
| 1 | `self._model.transcribe` on an external Whisper model, bound to the file's own `transcribe` |
| **3** | **genuine** — `self._capture.{start,stop,chunks}()` onto the class the constructor assigns |

So 38 fabrications for 3 real edges. I think that trade is right for a graph agents read, but it is a trade, not a free win.

### The 3 are recoverable, and my first attempt at recovering them was wrong

Python names an attribute's type in the class body — `self.x: T`, `self.x = T()`, a typed `__init__` parameter, a class-level annotation, a base class. I wrote a helper that read those with regexes over the class's source lines. Review killed it, and rightly:

- only `#` was stripped, so a type in a **docstring** beat the real assignment below it — turning a *correct* edge into a wrong one
- a **nested class's** `__init__` was read as the outer class's
- `self.session = requests.Session()` had the package stripped and bound an external object to a project `Session` — the fabrication class this whole thread is about
- the typed-`__init__`-parameter idiom (`def __init__(self, conn: Client): self.conn = conn`) wasn't read at all, so it *lost* edges that worked before

Doing it properly needs the AST rather than a line scan, so it is a separate change. Until it exists this shape is a silent miss.

### On the tests

Worth flagging as a general trap for this area: my first version of these tests passed against the *pre-change* resolver. With a single project-wide symbol of that name, the old bare-name fallback found the right target by luck, so the test proved nothing. Every negative test here now carries a **distractor** — a second project symbol with the same method name — and I verified all three fail against the pre-change resolver while the two boundary guards (a single-segment inferable receiver, a module-qualified call) pass on both arms.

860 tests across resolution, extraction, frameworks and kernel parity.

inth3shadows:
**This is the AST change I said was needed but missing.** My last comment ended with the 3 genuine edges lost to the gate, and: *"doing it properly needs the AST rather than a line scan, so it is a separate change. Until it exists this shape is a silent miss."* Pushed 15 commits; it exists now, and on the 249-file project the count goes back the other way — the same 3 edges (`self._capture.{start,stop,chunks}()`) recovered, with none of the 38 fabrications returning.

It reads the four shapes python actually uses to name an attribute's type — constructor assignment, typed `__init__` parameter, class-level annotation, `Optional` / `| None` unwrapping — out of the tree-sitter AST, and it fixes each defect review killed the regex version for: a docstring type no longer beats the real assignment, a nested class's `__init__` no longer donates to the outer class, and `self.session = requests.Session()` stays external.

Five review rounds hit this. The last one reproduced two defects **through the built binary**, both of which invented edges:

- **A unique path suffix is not an importable module.** The module-file fallback accepted any indexed path ending in `/<module>.py` so long as exactly one did. But the imports reaching it are overwhelmingly *not* project modules — 81 of 172 absolute import sources on my corpus were stdlib — so `from logging import Logger` bound to a project's own `app/utils/logging.py`, and `from redis.client import Redis` bound to a test double under `tests/fixtures/`. Now gated on a stdlib denylist plus a real package-root check (the anchor must not itself carry an `__init__.py`; every intermediate segment must be a package). PEP 420 namespace packages are refused deliberately — under this gate a silent miss is the right answer.
- **A commented-out import decided an edge.** The `__init__.py` re-export hop read import mappings raw, forty lines below the text filter added for the importing file. The top level has an agreement rule that absorbs a bogus binding by refusing both; the hop has none, so the comment was the *sole* binding and picked the file.

That is the same failure twice — a rule applied at one site and not the sibling one level down — which is the trap worth naming for anyone working in this file.

Also here: import text is classified properly now (comments *and* single/triple-quoted strings stripped, then the extractor's own patterns re-run over what's left, so the filter can never be stricter than the thing it filters); the supertype walk asks the pinned receiver **node** for its bases instead of its name, which was costing every inherited edge in any repo with a `Client` or `Config` defined twice; and the per-import file scan is indexed per resolution context, following the existing `luaFileBasenameIndexes` pattern.

Verification:

- `npm test` — 242 files, 4281 passed / 11 skipped.
- Corpus re-index **byte-identical** to the pre-fix build: same 2683 python call edges, none added, none lost. That corpus contains none of the layouts these defects need, which is why each fix ships with a purpose-built fixture instead of leaning on it.
- Every new test verified failing against its own parent commit — and per the distractor trap I flagged earlier, each carries a second project symbol of the same method name. One test is labelled in-place as a guard rather than a fix, because it passes on both arms.
- Indexing cost on a synthetic 4,641-file python project: 32.0 / 34.8 / 32.4 s before, 6.0 / 5.7 / 6.3 s after.

`__tests__/python-attr-type-cli.test.ts` drives the built binary on purpose: grammars load only inside parse workers, so an earlier version of this feature was inert in a real build while the whole suite stayed green over dead code.

The split offer stands unchanged — if you'd rather take only the resolver half, take **b34d85f** with e799cd0.

colbymchenry:
Heads-up: the **false-negative** half of #1681 (`isBuiltInOrExternal` escape for project-module receivers) is landing as a focused extract in #1749 — FP half already on `main` via #1748. This PR still carries the broader attr-type / branch-guard / kernel work; please rebase or drop the overlapping `isPythonProjectModule` / resolution.test.ts bits once #1749 merges to avoid conflict.

inth3shadows:
Done — merged `main` in at `ee83636` and dropped the overlapping bits. The PR is `MERGEABLE` again.

## Dropped, because they landed as your extracts

- **`isPythonProjectModule`** — gone from this branch entirely. #1749's copy is byte-identical to what was here apart from one comment, so this takes yours rather than merging the two. (git had auto-merged both into a duplicate definition; that is removed.)
- **`__tests__/resolution.test.ts`** — took `main` wholesale. Its +347 subsumes the +54 that was here, so **this branch's delta against `main` for that file is now zero**, which is what you asked for.
- **`isKnownClass`** — took the stricter #1776 form, with the module-scope collection binding short-circuiting ahead of the class escape.

## Reconciled rather than dropped

- **#1748 / the extractor half.** Both arms kept, yours first. `<inner>().<method>` claims the call-chain receiver; this branch's arm stays the catch-all beneath it for the two shapes that arm does not match — the attribute chain (`self.data.append`) and the subscript (`d[k].append`). Same ordering mirrored in `codegraph-kernel/src/python.rs`. Both comments here claimed to cover the call-chain shape and no longer do, so they are corrected in place. `torture.py` keeps both fixtures.
- **#1770 / quoted annotations.** Nothing to do and nothing duplicated — that fix is in `buildLocalReceiverTypePatterns` (local variables), while the attribute-type path here goes through `pythonAnnotationType`, which already strips a forward reference's quotes. Different receivers, both covered.
- **#1746 / `strip-comments.ts`.** `livePythonImportSources` had hand-rolled its own comment and string scanner; it now composes `blankStringContents(stripCommentsForRegex(source, 'python'))`, the same way the TS receiver-type read one function above already does. Checked against the eight cases that path exists for before swapping — a commented-out import, a docstring usage example, a single-line string holding import text, an inline comment after a real import, a triple quote opened and closed on one line, and two shapes that could trip the helper's JS regex-literal heuristic (`(a)/b`, an f-string with a slash). All eight agree with the scanner they replace, including the two the hand-written one got wrong.

## Verification

Linux, Node 22.23.2, **with the native kernel built and staged** — so the `kernel-*-parity` suites actually run instead of skipping themselves:

```
tsc --noEmit                       clean
kernel-tsjs-parity                 17/17     (this is what puts torture.py through the .node)
kernel-python-call-fabrication      1/1
npm test                           254 files, 4400 passed, 8 failed, 10 skipped
```

All 8 failures are pre-existing on `main`, checked rather than assumed:

- `object-literal-methods` and two in `ui-steps-api` — reproduce on a clean detached `upstream/main` worktree.
- `ui-server-api`'s "answers in under 100 ms" — a wall-clock assertion; passed and then failed at 113.06ms in the same command on a loaded box.
- **all four `kernel-dart-parity` cases** — reproduce on a clean `upstream/main` worktree *with its own freshly built kernel*. The arms disagree on `km` / `report` line numbers in `TortureCtors.dart`, so the kernel and wasm paths extract that file differently on `main` today. Unrelated to this PR (the kernel diff here is `python.rs` only), but worth someone's attention since published bundles run the kernel arm — happy to open it as an issue if it is not already tracked.

---

Measured what this branch's python work is worth, on repos neither of us chose — and ablated it so the number is attributable to one hunk rather than to "the branch".

**Method.** Three python codebases (318 `.py` files) indexed twice, both arms built from source on Linux/Node 22.23.2 **with the native kernel staged** (python is default-routed, so `python.rs` is live — a wasm-only run would measure the wrong thing). Upstream baseline `2f8cce5`. Diffed the `calls` edges where source and target are both python.

## The trade

```
corpus       upstream-only   ours-only
kb                      79           6
lodestone                8           6
testgraph                3           2
```

**82 of those 90 are fabricated.** I read every call site rather than inferring from the name:

```
config.py::load          -> reads.py::get         site: raw.get("machine", {})      a dict
propose.py::_merge_…     -> ledger.py::append     site: merged[key].append(r)       a list
                            …and propose.py never imports ledger
worker.py::_llm_verdict  -> test_llm_shapes::create  site: client.messages.create(  the SDK
test_tunnel.py::start    -> test_tunnel.py::start    site: self._thread.start()     a Thread
test_tunnel.py::stop     -> db.py::close             site: a connection
```

Production code bound to a symbol in a **test file**, and a function that appears to call **itself**. `merged[key].append(r)` is the subscript receiver; `(ts if cond else other).append(rel)` in the same repo is a parenthesised one.

**The 8 that are real, stated plainly.** All one shape — `idx.lexical.query(...)` where `idx = Index(...)` is an untyped local, so the receiver is two hops from nothing and the gate refuses. Against that, the branch *gains* 5 real edges upstream never had: `self.vector.query` resolves because `self.vector = VectorIndex(...)` and the attribute-type reader reads that assignment. Same file, same method name — `self.lexical.query` is kept, `idx.lexical.query` is dropped. That is the intended line, not an accident of tuning.

## Which hunk earns it

Rebuilt the branch twice, removing one piece at a time.

```
corpus       upstream   V1: no extractor arm   V2: arm, no gate   this branch
kb               3595                   3595               3530          3522
lodestone         839                    839                849           837
testgraph         466                    466                465           465
```

- **V1 — remove the extractor arm** (the python catch-all in `extractCall`, plus its `python.rs` mirror). The edge set becomes **byte-identical to upstream** on all three corpora — zero diff lines, not merely the same count. So the arm is the whole mechanism: every one of the 82 fabrications, and all 14 recovered edges, trace to it. Keeping the receiver's source text as a qualifier is what stops `append` from exact-matching a project function named `append`.
- **V2 — keep the arm, remove the gate.** Worse than either side: 8 fabrications return **and 20 new ones appear**, putting lodestone at 849 — above upstream's own 839. The inventions show why: `idx.lexical.query(...)` binds to **`vector.py::query`**, a different module that merely also has a `query`. Renaming the ref without also refusing to resolve it by bare name is the half-bridged state, and it measures worse than not doing it at all.

So the two halves are one mechanism, not two independent changes — and the attribute-type inference is what makes the strict half affordable, since it is the only thing that turns a refused `self.x.method()` back into a real edge.

## Where it is not better

Cross-copy binding. My kb and lodestone corpora each contain two sibling copies of one project, and **both arms** mis-bind across them — upstream 664 and 113, this branch 668 and 115. It adds 4 and 2. Neither side handles the duplicate-copy ambiguity; I am not claiming this branch fixes it, and the corpus is unusually hostile there.

Reproducible end to end if you want the fixtures — the whole thing is `codegraph init` twice and a `sqlite3` diff over the `calls` edges.

PR #1706, opened 2026-09-05
fix(resolution): an import naming the emitted .js extension resolves to its .ts source

### LINKED ISSUE #1705: TypeScript imports naming the emitted `.js` extension never resolve through the import resolver (wrapper methods become self-edges)
## Summary

A TypeScript relative import that names the emitted extension (`import { x } from './util.js'` for `util.ts`, the form `moduleResolution: node16 | nodenext | bundler` requires) never resolves through the import resolver. `resolveRelativeImport` appends each candidate extension to the specifier (`util.js.ts`, `util.js.tsx`, …) and then tries the specifier as-is (`util.js`), and none of those files exist, so it returns null. Every name imported that way then falls through to bare-name matching.

Two visible effects on a real repo (Chrome MV3 extension, 582 files, 92 of its `.ts` files import with `.js` specifiers):

1. **A method that wraps the same-named function it imports resolves to itself.** `renderDockStyles() { return renderDockStyles(); }` in a class, with `renderDockStyles` imported from `./template.js`, gets a `calls` edge from the method to the method (`resolvedBy: exact-match`, confidence 0.4). Eight such false self-edges here: three dock-template wrappers, four static wrappers on a consensus engine, one `...args` pass-through. This is the shape #1496 does not cover (that one is `this.field.method()`).
2. **Cross-module edges are name guesses.** On this tree 4,002 `calls`/`imports` edges were `resolvedBy: import` and 15,419 `exact-match`; with the remap below it is 7,312 and 12,784. `fuzzy` drops 44 → 23.

## Repro

```ts
// template.ts
export function renderDockStyles(): string { return ".dock {}"; }

// sidebar.ts
import { renderDockStyles } from "./template.js";
export class Sidebar {
  renderDockStyles(): string { return renderDockStyles(); }
}
```

`codegraph init` on those two files: the `calls` edge from `Sidebar::renderDockStyles` targets `Sidebar::renderDockStyles`, not `template.ts::renderDockStyles`. Same for `.jsx` → `.tsx`, `.mjs` → `.mts`, `.cjs` → `.cts`, and for an aliased specifier (`@/lib/util.js`).

## Fix

When the specifier ends in an emitted extension and no file with that exact name exists, retry with the source extensions TypeScript would have compiled from (`.js` → `.ts`, `.tsx`, `.d.ts`; `.jsx` → `.tsx`; `.mjs` → `.mts`; `.cjs` → `.cts`), for the TS/JS-family languages only. A real `.js` file still wins when both exist. PR incoming.

bompus:
Fixes #1705.

## What

`resolveRelativeImport` and the aliased path's `tryWithExt` now fall back to `findSourceForEmittedSpecifier`: when a specifier ends in `.js` / `.jsx` / `.mjs` / `.cjs` and no file of that exact name exists, the source extensions TypeScript compiles from are tried (`.ts`, `.tsx`, `.d.ts` / `.tsx` / `.mts`, `.d.mts` / `.cts`, `.d.cts`). Only for the TS/JS-family languages (`typescript`, `tsx`, `javascript`, `jsx`, `vue`, `svelte`, `astro`, `arkts`); a real `.js` beside the `.ts` still wins because the existing as-is check runs first.

## Why

Under `moduleResolution: node16 | nodenext | bundler` the emitted extension is what TypeScript requires in the specifier, so every import in such a project was invisible to the import resolver and each imported name fell through to bare-name matching. The clearest symptom is a method that wraps the same-named function it imports resolving to itself (a `calls` self-edge at confidence 0.4); the wider one is that cross-module edges in these projects were name guesses instead of import-backed.

## Measured

Real repo, 582 files, 92 `.ts` files with `.js` specifiers, indexed with and without this change on the same build:

| | before | after |
| --- | --- | --- |
| `calls`/`imports` edges `resolvedBy: import` | 4,002 | 7,312 |
| … `resolvedBy: exact-match` | 15,419 | 12,784 |
| … `resolvedBy: fuzzy` | 44 | 23 |
| self-`calls` edges | 31 | 22 (the 8 wrapper-method ones gone; the rest are recursion) |

## Tests

`__tests__/import-emitted-specifier.test.ts`: unit cases through `resolveImportPath` for `.js`, `.jsx`, `.mjs`, `.cjs`, the real-`.js`-wins case, an aliased `@/lib/util.js`, a missing target, and a non-TS language; plus an end-to-end `CodeGraph.init` on the wrapper-method repro asserting the edge lands on the imported function and not on the method. Full suite: no failure outside the host's known EPERM/EBUSY teardown set.


bompus:
## Real-repo check: one edge gained, zero lost

Ran this against **vitejs/vite** (1,635 files, 9,354 nodes, 27,499 edges) — index the same checkout with each build, then diff the full edge set joined back to symbol names and call-site lines. Indexing is deterministic, so the diff is this change and nothing else.

**27,499 → 27,500 edges. One gained, none lost.** The one:

```
imports  playground/resolve/ts-extension/index-js.js@1  ->  hello.ts:msg
```

which is the fixture line `import { msg as msgJs } from './hello.js'`, resolving to `hello.ts` — the exact case in the title.

What I found interesting is the shape of the gap. That fixture imports five sibling modules through their emitted extension, and `main` already resolved four of them:

| line | import | on `main` | with this PR |
|---|---|---|---|
| 1 | `'./hello.js'` → `hello.ts` | **unresolved** | resolved |
| 2 | `'./hellocjs.cjs'` → `hellocjs.cts` | resolved | resolved |
| 3 | `'./hellojsx.jsx'` → `hellojsx.tsx` | resolved | resolved |
| 4 | `'./hellomjs.mjs'` → `hellomjs.mts` | resolved | resolved |
| 5 | `'./hellotsx.js'` → `hellotsx.tsx` | resolved | resolved |

So the only broken variant was the plainest and by far the most common one — `.js` → `.ts` — while `.cjs`/`.mjs`/`.jsx` all worked. That's a good sign the change is aimed at the right hole, and the zero-loss column says it doesn't widen anything else: across 27,499 edges nothing was dropped or re-pointed.

One caveat on scope. This checkout has no kernel binary staged, so extraction ran the wasm path in both arms. Resolution is shared TS and sits downstream of extraction, so the result holds either way, but I have not re-run it against a kernel-extracted graph.

bompus:
Rebased onto `cd4e65b` (current `main`) — new head `9a148ad`. The only conflict was `CHANGELOG.md`; resolved as a union with the entries merged since. No source conflict.

Checks on the rebased branch: `tsc --noEmit` clean; `__tests__/import-emitted-specifier.test.ts` passes. `__tests__/resolution.test.ts` has 4 failures, all `EPERM` temp-directory teardown on Windows — the identical set fails on a clean `cd4e65b` checkout, so they are pre-existing and unrelated (#1717 covers them).

Worth noting for review: #1631 (`a6f52d7`) now keeps the existence probe inside the project root, in `src/resolution/index.ts` and `src/utils.ts`. This PR's remap lives in `src/resolution/import-resolver.ts` and probes for a sibling source file, so the two do not overlap, but they are the same concern.

colbymchenry:
Closed in favor of #1767, which lands this change rebased onto current `main`.

Only conflict was `CHANGELOG.md` (union with entries merged since). Source + tests applied cleanly. Verified fail→pass on Linux for `./util.js` → `util.ts` and the full `__tests__/import-emitted-specifier.test.ts` suite (7/7).

Thanks @bompus — credited in the CHANGELOG entry.

PR #1723, opened 2026-09-06
test: release the handles a temp project holds before removing it (Windows EPERM)

### LINKED ISSUE #1722: Tests leak file handles into their temp projects: 24 fail with EPERM on Windows
On Windows, `bun run test` fails 24 tests across 7 files. Every one of them fails **after its assertions have passed**, in `fs.rmSync` on the temp project:

```
× Resolution Module > PHP Include Resolution > resolves require_once to a file→file imports edge (#660)
  → EPERM, Permission denied: \?\C:\Users\…\Temp\codegraph-php-e2e-zmZ4mt
    ❯ __tests__/resolution.test.ts:3792  fs.rmSync(tempProject, { recursive: true, force: true });
```

Windows refuses to remove a directory that holds an open file; POSIX unlinks one happily, which is why CI does not see this. The tests are not wrong about the graph — they are wrong about the cleanup.

## Two causes

**1. The graph is never closed.** `__tests__/arkts-resolution.test.ts` calls `CodeGraph.initSync` 7 times and `close()` zero times; `frameworks-integration.test.ts` closes in two JVM tests and not in the other three; `resolution.test.ts`'s C++/PHP end-to-end tests additionally open a second `DatabaseConnection` for a raw edge query and never close that either. The SQLite handle is still open when `rmSync` runs. 14 tests.

**2. A killed child still holds its handles.** The four `mcp-*` suites do close the graph and do `child.kill('SIGKILL')` — then call `rmSync` synchronously in the same tick. A signalled process has not exited yet. 10 tests.

Measured, both arms, same file and command back to back: adding `cg.close()` to `arkts-resolution.test.ts`'s `ohpm main entry` test makes it pass; removing that one line makes it fail with EPERM again.

## Affected

| Tests | File | Cause |
| ---: | --- | --- |
| 7 | `arkts-resolution.test.ts` | unclosed graph |
| 4 | `resolution.test.ts` (C++/PHP e2e) | unclosed graph + unclosed `DatabaseConnection` |
| 3 | `frameworks-integration.test.ts` (JVM FQN imports) | unclosed graph |
| 3 | `mcp-initialize.test.ts` | killed child not awaited |
| 3 | `mcp-roots.test.ts` | killed child not awaited |
| 3 | `mcp-subproject-adoption.test.ts` | killed child not awaited |
| 1 | `mcp-daemon.test.ts` | killed child not awaited (fixed 50ms grace period) |

Awaiting the tracked child is necessary but not sufficient for the MCP suites: with only that, a full-suite run still failed 2 of 4,240 — a different pair each time — because the server can leave a detached daemon the suite never tracks, and `mcp-daemon`'s concurrent-launcher test deliberately leaves one alive.

## Environment

Windows 11, Node 22, `vitest run` from a built tree (`tsc` + assets + `ui` workspace). Before: 3,975 passing / 24 failing / 234 skipped. Not reproduced on Linux — the guess is that nobody outside a Windows host ever sees it, and the failure is pure teardown, so no behaviour of the indexer is implicated.

PR follows.

bompus:
Fixes #1722.

24 tests across 7 files fail on Windows, every one of them **after its assertions have passed**, in the `fs.rmSync` that removes the temp project. Windows will not remove a directory holding an open file; POSIX unlinks one happily, which is why CI is green. Nothing about the indexer is implicated — this is teardown only.

## The two causes, fixed separately

**`fb89ac5` — the graph is never closed.** `arkts-resolution.test.ts` calls `CodeGraph.initSync` 7 times and `close()` zero times; three of the four JVM tests in `frameworks-integration.test.ts` omit the `cg.close()` their two neighbours already have; the C++/PHP end-to-end tests in `resolution.test.ts` also open a second `DatabaseConnection` for a raw edge query and close neither.

Closed in the **teardown**, not at the end of the test body, so a failing assertion cannot skip the close and turn one red into two. In `arkts-resolution.test.ts` that means a `let cg` per describe and a `cg?.close()` in the existing `afterEach`; in `resolution.test.ts` the test's own connection is hoisted out of the `try` and closed in the `finally` beside the graph. `frameworks-integration.test.ts` keeps the end-of-body `cg.close()` its passing neighbours use, since that is the file's own convention.

**`476b229` — a killed child still holds its handles.** The four `mcp-*` suites do close the graph and do `child.kill('SIGKILL')`, then `rmSync` synchronously in the same tick; the process has been signalled, not reaped. They now await the actual `exit`. `mcp-daemon.test.ts` had a fixed `setTimeout(50)` standing in for this, now replaced by the same await plus the file's own `waitProcessExit` for the detached daemon it reaps by pid.

Awaiting the tracked child is necessary but **not sufficient**, and I have the failed run that shows it: with only the awaits, a full suite still failed 2 of 4,240 — a *different* pair each time — because `serve --mcp` can leave a detached daemon the suite never tracks, and `mcp-daemon`'s concurrent-launcher test deliberately leaves one alive and asserts it is alive. So the removals also retry, which is what `fs.rmSync`'s `maxRetries` exists for: 1s in the three proxy suites, 6s in `mcp-daemon` where the live daemon is the point of the test.

## Measured

The diagnosis is an A/B, not a reading: adding `cg.close()` to `arkts-resolution.test.ts`'s `ohpm main entry` test turns it green, removing that one line turns it red with EPERM again, same file and command back to back.

| | before | after |
| --- | ---: | ---: |
| Full suite, Windows | 3,975 passing / **24 failing** | **4,017 passing / 0 failing** |

Run twice in a row after the change, both green, because one green run does not close a race. The 7 affected files alone: 214 passing / 24 failing → **238 passing**. `tsc --noEmit` clean.

Built tree in both arms (`tsc` + assets + the `ui` workspace), Windows 11, Node 22. I did not reproduce a Linux run; on POSIX these teardowns cannot fail this way, and the change is inert there apart from the awaits.

### DIFF
--- a/__tests__/arkts-resolution.test.ts
+++ b/__tests__/arkts-resolution.test.ts
@@ -25,7 +25,12 @@ beforeAll(async () => {
 
 describe('ArkTS attribute-chain resolution precision', () => {
   let tmpDir: string | undefined;
+  let cg: CodeGraph | undefined;
   afterEach(() => {
+    // Windows refuses to remove a directory holding an open file, so the
+    // graph's SQLite handle has to go before the temp project does.
+    cg?.close();
+    cg = undefined;
     if (tmpDir) fs.rmSync(tmpDir, { recursive: true, force: true });
     tmpDir = undefined;
   });
@@ -65,7 +70,7 @@ describe('ArkTS attribute-chain resolution precision', () => {
         '}\n'
     );
 
-    const cg = CodeGraph.initSync(tmpDir);
+    cg = CodeGraph.initSync(tmpDir);
     await cg.indexAll();
 
     const fns = cg.getNodesByKind('function');
@@ -99,7 +104,12 @@ describe('ArkTS attribute-chain resolution precision', () => {
 
 describe('ArkTS ohpm workspace import resolution', () => {
   let tmpDir: string | undefined;
+  let cg: CodeGraph | undefined;
   afterEach(() => {
+    // Windows refuses to remove a directory holding an open file, so the
+    // graph's SQLite handle has to go before the temp project does.
+    cg?.close();
+    cg = undefined;
     if (tmpDir) fs.rmSync(tmpDir, { recursive: true, force: true });
     tmpDir = undefined;
   });
@@ -146,7 +156,7 @@ describe('ArkTS ohpm workspace import resolution', () => {
         '}\n'
     );
 
-    const cg = CodeGraph.initSync(tmpDir);
+    cg = CodeGraph.initSync(tmpDir);
     await cg.indexAll();
 
     const classes = cg.getNodesByKind('class');
@@ -168,7 +178,12 @@ describe('ArkTS ohpm workspace import resolution', () => {
 
 describe('ArkUI state → build() re-render bridge (assignment-gated)', () => {
   let tmpDir: string | undefined;
+  let cg: CodeGraph | undefined;
   afterEach(() => {
+    // Windows refuses to remove a directory holding an open file, so the
+    // graph's SQLite handle has to go before the temp project does.
+    cg?.close();
+    cg = undefined;
     if (tmpDir) fs.rmSync(tmpDir, { recursive: true, force: true });
     tmpDir = undefined;
   });
@@ -201,7 +216,7 @@ describe('ArkUI state → build() re-render bridge (assignment-gated)', () => {
         '}\n'
     );
 
-    const cg = CodeGraph.initSync(tmpDir);
+    cg = CodeGraph.initSync(tmpDir);
     await cg.indexAll();
 
     const methods = cg.getNodesByKind('method');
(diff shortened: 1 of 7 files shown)


bompus:
Duplicate of #1717, which I opened this morning and did not check before filing this. #1717 is the better patch and supersedes this one: same close-in-teardown for arkts-resolution / frameworks-integration / resolution, same await-the-exit for the mcp-* suites, plus a shared `__tests__/rm-temp.ts` retry helper instead of retry options inlined at each call site, plus two this PR does not touch — `cli-ui-command`'s waitForMarker existence race and watcher.test.ts. It is also verified over 10 consecutive full runs against my 2 here.

The one thing worth carrying across: I reached the same 'awaiting the tracked child is necessary but not sufficient' conclusion independently here, from a run that still failed 2 of 4,240 with a different pair each time, which is the same detached-daemon case #1717 documents.

Closing this and #1722. Review effort belongs on #1717.

PR #1732, opened 2026-09-07
fix(resolution): a definition its language makes file-local is not a cross-file target

### LINKED ISSUE #1730: C: a call resolves by exact name onto a static function in another translation unit (4,448 edges on betaflight)
A `static` function in C is visible only in its own translation unit, but the exact-name strategy accepts it as the target of a call from any other file. When the caller's file has no same-named definition and another file has a `static` one, the call resolves onto it.

## Reproduction

`core.c`:
```c
void coreRun(void)
{
    usbGetDescriptor();
}
```
`usb_audio.c`:
```c
static void usbGetDescriptor(void)
{
}
```

`codegraph init`, then the edges of `coreRun`:

```
coreRun --calls--> usb_audio.c::usbGetDescriptor   resolvedBy: exact-match
```

The real target is whatever `core.c` links against — a non-static `usbGetDescriptor` elsewhere, a function pointer, a macro — but never that one. When a same-file `static` of the same name exists the proximity score in `findBestMatch` picks it, so the wrong edge appears exactly when the caller's file has no definition, which is the case that matters for cross-file navigation.

## Scale

On a 2,109-file betaflight tree, of 42,460 cross-file `calls`/`imports` edges resolved by exact-match or fuzzy, **4,448 target a function declared `static` in a file other than the caller's** (10%). A typical run of them:

```
lib/main/AT32F43x/middlewares/usb_drivers/src/usbd_sdr.c:usbd_get_descriptor
  → lib/main/AT32F43x/middlewares/usbd_class/audio/audio_desc.c:get_device_descriptor      (static)
  → …/audio/audio_desc.c:get_device_configuration                                           (static)
  → …/audio/audio_desc.c:get_device_lang_id                                                 (static)
```

`usbd_sdr.c` calls those through a descriptor table of function pointers; `audio_desc.c`, `cdc_desc.c`, `hid_desc.c`, … each define their own `static get_device_descriptor`, and the matcher hands every call to the first file it ranks.

## Relationship to #1719 / #1720

This is the C form of the same defect: a binding that the language makes file-local is admitted as a cross-file candidate. #1720's `isSealedModule` is deliberately JS/TS-only. In C the signal is per symbol, not per file — the definition's storage class — and the extractor already sees the `static` specifier on the `function_definition`, so it could either mark the node (a `visibility`/`isStatic` flag) or the resolver could read the definition line as #1720 reads source. Same-file candidates must stay eligible, and a `static` in a header included by the caller is a genuine same-unit call, which is the one exception to keep in mind.

### LINKED ISSUE #1731: Kotlin, Go, Rust: a private / unexported / non-pub definition is accepted as a cross-file name-match target
Sibling of #1730 (C `static`), measured on the same day across the other languages in one workspace. A definition the language makes file-, class- or package-local is accepted as the target of a cross-file name match whenever the names agree. The information to decline is already in the graph — `visibility: 'private'` for Kotlin and Rust, the identifier's case for Go — and is not consulted.

## Kotlin — `private fun`, 17 of 164 cross-file name-matched calls on a 114-file Android app

```kotlin
// AdaptiveVideoBudgetController.kt
private fun apply(bitrateBps: Long): Long { … }

// MainActivity.kt:43
uiPrefs.edit().putBoolean("autoConnect", enabled).apply()    // SharedPreferences.Editor.apply()
```
resolves as `MainActivity::onCreate --calls--> AdaptiveVideoBudgetController::apply` (`exact-match`). Same shape for `DirectEndpoint.parse(...)` onto another class's `private fun parse`, `getSharedPreferences(...).getBoolean(...)` onto a `private fun getBoolean`, a `CountDownLatch.await()` in tests onto a test file's `private fun await` (`instance-method`, ×8), and — from a JavaScript file in the same repo — `.token` onto `WebRtcTransport.kt`'s private `token`.

## Go — unexported identifier from another package, 5 + 55 cross-language

```go
// windows-wails/cmd/turnprobe/main.go
func fail(format string, args ...any) { … }

// windows-wails/turnserver.go:200,216   (package main of a different directory)
fail := func(message string) { … }
…
fail(err.Error())
```
resolves onto `cmd/turnprobe/main.go::fail`. The local closure is not a node (#1669's shape), and the name then reaches an unexported func of another package, which Go cannot call. The same tree has 55 `calls` from JavaScript files (`leaflet.js`, `validate.mjs`, `rcusb.js`) onto unexported Go funcs `add` / `join` — cross-language, cross-package, impossible twice over.

Note for the fix: the Go extractor's `isExported` is `false` for **every** method (148 of 148 on this tree, including `func (w *rotatingLogWriter) Close()`), so a resolver-side rule has to judge by the identifier's case, not the flag; the flag itself is a separate extractor defect.

## Rust — non-`pub` item outside its module subtree, 36 + 23 on two crates

```rust
// crates/skylab-hub-log-server/src/lib.rs
fn count(…) { … }                      // private, different crate

// crates/skylab-hub-core/src/routes.rs:1040
… .iter().count()                       // Iterator::count
```
resolves onto the private `fn count` of the other crate. Likewise `f.finish()` (a `DebugStruct`) onto a `HashingWriter::finish` in another crate, `ui.text(…)` (egui) onto a `ProgressState::text` in a sibling module, and `pub use profile::Profile` onto a private `fn profile` (the module path segment taken for a function).

Two shapes must stay resolvable: a child module reaching its ancestors' private items (`src/net/tcp.rs` → `src/net.rs`'s `fn shared`, via `super::`), and a method in an `impl Trait for Type` block, which carries no `pub` because the trait decides — the extractor records it as `private`.

## Where it lives

All of these come through the name-based strategies (`exact-match`, `fuzzy`, `instance-method`, `qualified-name`), so the natural place is one visibility test on the target the pipeline settled on, applied in `ReferenceResolver` after `matchReference` so that a rejection ends the reference unresolved rather than letting the next strategy commit to a namesake (declining inside `matchByExactName` alone did exactly that: eight new fuzzy edges onto a JS `fail` in a different file, the real `fail` being a local arrow). Measured that way with #1730's C rule included, five corpora, all against `b9ca4b7`: betaflight −4,451, the Android/Go/JS app −142, two Rust workspaces −195 and −92, vite ±0; nothing gained anywhere. PR to follow.

danusha2345:
Fixes #1730. Fixes #1731. Standalone off `main` (`b9ca4b7`), two commits — the second narrows the C rule to translation units after the first cut removed 4,306 real calls onto header `static inline`s; details in the comment below.

## What

Name matching accepted any same-named definition as the target of a reference from another file, however the language scopes that definition. One test, `isVisibleAcrossFiles`, now declines a candidate in another file when:

| language | rule | source of truth |
| --- | --- | --- |
| C / C++ | the function is defined in a **source file** (`.c .cc .cpp .cxx .m .mm`) and its definition line carries `static` — a header's `static inline` is textually included into every unit that names it and stays visible | read from source (the extractor records no storage class; the kernel arm would need the same field) |
| Kotlin, Java, C#, Swift, Scala, Dart, PHP | `visibility === 'private'` | extractor |
| Go | lowercase identifier and a different directory | the name's case — the extractor's `isExported` is unset for every Go method, exported or not |
| Rust | non-`pub` item and the reference is outside the item's module subtree | extractor + path: `src/net.rs` / `src/net/mod.rs` own `src/net/`; a child reaches its ancestors' private items via `super::`. A method in an `impl Trait for Type` block has the trait's visibility and is exempt (read from the enclosing `impl` header) |

Same-file candidates are always visible.

## Where the test runs, and why there

In `ReferenceResolver`, on the target the whole name-matching pipeline settled on (`exact-match`, `fuzzy`, `instance-method`, `qualified-name` all come through it), so a rejection ends the reference unresolved. The first cut declined inside `matchByExactName`, #1720-style, and that let the ref fall through to `matchFuzzy`, which committed to a same-language namesake the ranking had passed over: eight new fuzzy edges on one tree, every one onto a JavaScript `fail` in a different file, the real `fail` being a local `const fail = (msg) => …` the graph does not hold (#1669's shape). `matchFuzzy` checks its own survivor as well, since nothing runs after it.

## Measurement

Five corpora, each indexed at `b9ca4b7` and at this branch (wasm arm), edge rows joined back to symbols and keyed with `resolvedBy`:

| corpus | files | LOST | GAINED |
| --- | --- | ---: | ---: |
| betaflight fork (C) | 2,109 `.c` | **145** | 0 |
| Android / Go / Electron app | 114 `.kt`, 42 `.go`, 96 `.js` | **142** | 0 |
| emmc-reader-gui (Rust workspace) | 71 `.rs` | **195** | 0 |
| skylab_hub (Rust + Kotlin) | 35 `.rs`, 4 `.kt` | **92** | 0 |
| vitejs/vite (JS/TS only) | 1,635 | 0 | 0 |

Read back from source, the removals are the shapes in the two issues: STM32 USB class sources → `static` handlers in GD32's `usbd_enum.c`, `usbd_get_descriptor` → a `static get_device_descriptor` in a USB class file it never links; `editor.apply()` → an unrelated class's `private fun apply`; `latch.await()` in tests → a test file's `private fun await` (×8); `leaflet.js` / `validate.mjs` → unexported Go `add` / `join` (55 cross-language rows); `Vec::new()` → a private `fn new` in another crate (×7); `ui.add(…)` (egui) → a private `add`; `f.finish()` → `HashingWriter::finish` in another crate. The Rust breakdown after the trait exemption: 171 `instance-method`, 8 `exact-match` calls, 16 `references` — none onto a trait-impl method.

Not touched: the Go extractor's `isExported` on methods (a separate extractor defect noted in #1731), the C storage class as a node field, and the local-closure gap behind the JS `fail` (#1669).

## Tests

`__tests__/cross-file-visibility.test.ts`, whole pipeline over source fixtures, each invisible shape paired with the visible one of identical form: C `static` in another source file vs non-static vs a header's `static inline` (kept) vs same-file `static` with the keyword on its own line; Kotlin `private fun apply` reached through an SDK-style `editor.apply()` vs a public `fun apply` on a typed receiver; Go unexported func across packages vs within the package and an exported func elsewhere; Rust a sibling module's private fn vs a parent's private fn from a child, a trait-impl method through a typed receiver, and a `pub fn` anywhere. 11 cases. `resolution.test.ts`, `frameworks-integration.test.ts`, `pr19-improvements.test.ts` unchanged: 200 across `cross-file-visibility` + `resolution.test.ts`, `tsc` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -201,6 +201,8 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 #### Symbols, tests and the viewer
 
+- **A definition its language makes file-local no longer captures calls from other files.** A C `static` function, a Kotlin/Java/C#/Swift `private` member, a Go unexported name in another package and a Rust non-`pub` item outside its module subtree cannot be what a name in another file means, but name matching accepted them whenever the names agreed: on a betaflight tree 4,448 calls landed on a `static` in some other file, an Android `editor.apply()` on an unrelated class's `private fun apply`, a JavaScript `fail(...)` on a Go program's `func fail`. Such a target is now declined outright — the reference stays unresolved rather than being handed to another file's namesake. Same-file definitions and Rust trait-impl methods are unaffected. Re-index after upgrading. (#1730)
+
 - **Files under an `e2e/` directory count as tests.** Their calls no longer appear as production callers in Steps, dead-code and test badges.
 
 - **Production code under a `samples` or `examples` package path is no longer treated as test code.** A Kotlin or Java project whose package path runs through `com/google/samples/…` (Now in Android, for one) had nearly every file counted as a fixture, so the Map opened on `build-logic`, the entry points hid the app, and dead-code and test badges were wrong. Only the project layout above a `src/` folder decides now; the package path below it never does.
--- a/__tests__/cross-file-visibility.test.ts
+++ b/__tests__/cross-file-visibility.test.ts
@@ -61,6 +61,14 @@ describe('C: a static function is local to its translation unit', () => {
     expect(await calleesOf('coreRun')).toContain('usb_audio.c:usbGetDescriptor');
   });
 
+  it('keeps a static inline defined in a header: it lives in every unit that includes it', async () => {
+    project({
+      'protocol.h': 'static inline void mav_put_char(char *buf, char c)\n{\n    buf[0] = c;\n}\n',
+      'core.c': '#include "protocol.h"\n\nvoid coreRun(char *b)\n{\n    mav_put_char(b, 0);\n}\n',
+    });
+    expect(await calleesOf('coreRun')).toContain('protocol.h:mav_put_char');
+  });
+
   it('keeps a same-file static, whichever line the keyword is on', async () => {
     project({
       'core.c': 'static void\nhelper(void)\n{\n}\n\nvoid coreRun(void)\n{\n    helper();\n}\n',
--- a/__tests__/cross-file-visibility.test.ts
+++ b/__tests__/cross-file-visibility.test.ts
@@ -1,0 +1,142 @@
+/**
+ * A definition the language makes file-local is not a candidate for a
+ * cross-file name match: a C `static`, a Kotlin `private fun`, a Go unexported
+ * identifier in another package, a Rust non-`pub` item outside its module
+ * subtree. Each case pairs the invisible shape with the visible one of
+ * identical form, so the assertion discriminates on visibility alone.
+ */
+
+import { describe, it, expect, afterEach } from 'vitest';
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+import CodeGraph from '../src/index';
+
+let tempDir: string;
+let cg: CodeGraph | null = null;
+
+function project(files: Record<string, string>): void {
+  tempDir = fs.mkdtempSync(path.join(os.tmpdir(), 'codegraph-visibility-'));
+  for (const [rel, content] of Object.entries(files)) {
+    const abs = path.join(tempDir, rel);
+    fs.mkdirSync(path.dirname(abs), { recursive: true });
+    fs.writeFileSync(abs, content);
+  }
+}
+
+/** `calls` targets of the function named `caller`, as `file:name` strings. */
+async function calleesOf(caller: string): Promise<string[]> {
+  cg = await CodeGraph.init(tempDir, { index: true });
+  cg.resolveReferences();
+  const from = cg.getNodesByKind('function').concat(cg.getNodesByKind('method')).find((n) => n.name === caller)!;
+  expect(from).toBeDefined();
+  return cg
+    .getOutgoingEdges(from.id)
+    .filter((e) => e.kind === 'calls')
+    .map((e) => cg!.getNode(e.target))
+    .filter((n): n is NonNullable<typeof n> => !!n)
+    .map((n) => `${n.filePath}:${n.name}`);
+}
+
+afterEach(() => {
+  cg?.close();
+  cg = null;
+  fs.rmSync(tempDir, { recursive: true, force: true });
+});
+
+describe('C: a static function is local to its translation unit', () => {
+  it('does not resolve a call onto a static in another file', async () => {
+    project({
+      'core.c': 'void coreRun(void)\n{\n    usbGetDescriptor();\n}\n',
+      'usb_audio.c': 'static void usbGetDescriptor(void)\n{\n}\n',
+    });
+    expect(await calleesOf('coreRun')).not.toContain('usb_audio.c:usbGetDescriptor');
+  });
+
+  it('still resolves onto a non-static function in another file', async () => {
+    project({
+      'core.c': 'void coreRun(void)\n{\n    usbGetDescriptor();\n}\n',
+      'usb_audio.c': 'void usbGetDescriptor(void)\n{\n}\n',
+    });
+    expect(await calleesOf('coreRun')).toContain('usb_audio.c:usbGetDescriptor');
+  });
+
+  it('keeps a same-file static, whichever line the keyword is on', async () => {
+    project({
+      'core.c': 'static void\nhelper(void)\n{\n}\n\nvoid coreRun(void)\n{\n    helper();\n}\n',
+      'other.c': 'static void helper(void)\n{\n}\n',
+    });
+    expect(await calleesOf('coreRun')).toEqual(['core.c:helper']);
+  });
+});
+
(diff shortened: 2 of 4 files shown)


danusha2345:
Correction, pushed as `42a9267`: the first cut's C rule was wrong for headers, and the betaflight number in the description was mostly that error.

Of the 4,451 rows it removed, **4,306 targeted a `static inline` in a header** — MAVLink's generated `mavlink_msg_*.h` calling `protocol.h`'s `_mav_put_char_array` / `mav_array_assign_char` and each other's `_pack` / `_decode` helpers. Those calls are real: a header is textually included, so the function exists in every unit that names it. Only a `static` defined in a translation unit is local to it.

The rule now applies to candidates in a source file (`.c .cc .cpp .cxx .c++ .m .mm`) only. Same tree, same base: **LOST 145, GAINED 0**, header targets in the removed set: 0. The 145 are the shape #1730 describes — STM32 USB class sources onto GD32's `usbd_enum.c` statics, and the descriptor-table `get_device_*` case. A test pins the header case (`static inline` in `protocol.h`, called from `core.c`, edge kept). The other four corpora are unaffected by the change (no C).

I'll correct #1730's count there too; the 10% figure in it was counting the header calls.

colbymchenry (commented):
### Linux verification (Forge) — approve the approach

Reviewed and verified on Linux (Node 22.19, `CODEGRAPH_KERNEL=0`). The placement is correct: `isVisibleAcrossFiles` runs in `ReferenceResolver` **after** `matchReference` settles (and again inside `matchFuzzy`), so a rejection leaves the ref unresolved instead of fuzzy-falling-through — the pitfall called out in #1731.

**Repro on `main` (df435d5)** — all three #1731 shapes wrongly resolve:
- Kotlin: `onCreate` → `Budget.kt:apply` (private)
- Go: `Run` → `cmd/probe/main.go:fail` (unexported, other package)
- Rust: `run` → `src/util.rs:count` (sibling private)

**On this PR's logic (rebased onto latest main)** — same fixtures decline; `__tests__/cross-file-visibility.test.ts` 11/11 + `resolution.test.ts` 189/189 pass. Rust child→ancestor + trait-impl exemptions look right; Go judged by identifier case as required while #1734 is still open.

This branch was two commits behind `main` (#1742 AGENTS.md, #1743 generators). To land without conflict I rebased your commits and opened https://github.com/colbymchenry/codegraph/pull/1745 (changelog cite for #1731 + C source-vs-header clarification only). Happy to close this PR in favor of #1745, or take a force-push rebase here instead — your call.

danusha2345:
Closing in favor of #1745 — thanks for rebasing and verifying it on Linux rather than rewriting. The Go by-name-case rule stays until #1734 lands; I'll rebase #1734, #1715, #1718, #1735 and #1736 onto current `main` today.

PR #1733, opened 2026-09-07
fix(extraction): blank a C macro call's designated-initializer arguments before parsing

### LINKED ISSUE #1729: C: a macro call with designated-initializer arguments and a trailing comma runs the enclosing function to end of file, nesting or dropping every function after it
A C function whose body contains a macro call with designated-initializer arguments and a trailing comma is recorded with an extent that runs to the end of the file. Every function defined after it is either dropped from the graph or nested under it as `outer::inner`, and from there the resolver treats it as a closure it cannot reach.

## Reproduction

`pid.c`:

```c
void f(void)
{
    M(a, b,
        .x = 1,
        .y = { 1, 2 },
    );
}

void g(void)
{
}

int h(void)
{
    return 1;
}
```

`codegraph init` and query `nodes`:

| name | qualified_name | lines |
| --- | --- | --- |
| `f` | `f` | **1–16** (source: 1–7) |
| `g` | — | **missing** |
| `h` | **`f::h`** | 13–16 |

Remove the comma after `{ 1, 2 }` and all three come out right (`f` 1–7, `g` 9–11, `h` 13–16). tree-sitter-c does not accept the shape — a `call_expression` whose arguments are `.x = …` items ending in a comma — and its error recovery closes the `function_definition` at the end of the file. This is what the shipped `tree-sitter-c.wasm` returns; the extractor records the extent it is given.

The shape is not exotic. betaflight resets every config struct through it:

```c
void resetPidProfile(pidProfile_t *pidProfile)
{
    RESET_CONFIG(pidProfile_t, pidProfile,
        .pid = { [PID_ROLL] = PID_ROLL_DEFAULT, … },
        …
        .adrc_sigma_decay_sched = 0,
    );
}
```

On a 2,109-file betaflight tree the graph shows `resetPidProfile` at lines 168–1667 (source: 168–309) with the 45 functions after it as its children, and **310 C functions in 73 files** nested inside another function this way — STM32 HAL sources (`stm32h7xx_ll_utils.c`, `…_ll_tim.c`, `…_ll_usart.c`) have the same shape. (That betaflight variant is one where the trailing comma alone does not explain it; the tiny case above is the reproducible core.)

## Effect on resolution

`isLexicallyReachable` (#1230) reads the nesting as a scope: exact-match declines those 310 functions for every caller in another file, and on `main` the fuzzy fallback then picks them up at confidence 0.5 — 117 real `calls` into `pid.c` resolve that way instead of at 0.9. With #1718 the fuzzy fallback checks reachability too, so the 117 vanished; #1718 now gates C/C++ out of the nesting check (`5eb5750`), which is a resolver-side patch over this extraction defect, not a fix for it. Callers/callees of the dropped `g`-shaped functions have no node at all.

## Where a fix could go

Either handle the recovered tree in the C extractor — when a `function_definition` contains an `ERROR` node, close it at the brace that balances its opening one and re-scan the remainder for top-level definitions — or pre-process the trailing-comma shape before parsing. I have not attempted either; the kernel arm (`codegraph-kernel/src/c.rs`) would need the same treatment, since the grammar is the same.

danusha2345:
Fixes #1729. Standalone off `main` (`b9ca4b7`), one commit, pre-parse only — the blank runs before the kernel route point, so both extraction arms receive the same bytes and the kernel needs no change.

## What

tree-sitter-c has no rule for `.field = value` as a call argument. A statement-level

```c
RESET_CONFIG(pidProfile_t, pidProfile,
    .pid = { [PID_ROLL] = PID_ROLL_DEFAULT, … },
    .pidSumLimit = PIDSUM_LIMIT,
    …                                   // ~140 of these
);
```

parses with an ERROR per argument, and past roughly a hundred of them the grammar's error recovery gives up on the enclosing function: the `function_definition` runs to the end of the file, the function after it produces no node at all, and every later one is nested under the first. That is not a matter of shape — the issue's three-function reproduction, and every smaller variant I tried (trailing comma or not, nested designators, array designators), keeps correct extents with the ERRORs local. It is the count: a prefix search over betaflight's real `resetPidProfile` breaks at the 104th–106th argument whatever those lines contain.

`blankCDesignatedMacroArgs` runs at the head of `preParseCSource`, in the style of the other blanks there: a statement-level call (`);` follows) of a macro-cased name whose argument list holds a designator at argument depth — `.name =` or `[index] =`, not `==` — has that list emptied to spaces with newlines kept, so `RESET_CONFIG(\n\n…\n);` parses cleanly and every byte offset survives. `log(.5)` and `OTHER_MACRO(a == b, c)` are left alone. The references inside the initializer are the price; the broken parse was not yielding them either.

## Measured

betaflight fork, 2,109 `.c` files, indexed at `b9ca4b7` and at this branch, wasm arm:

| | base | this |
| --- | ---: | ---: |
| `resetPidProfile` extent | 168–**1667** | 168–**309** (source: 168–309) |
| C functions nested inside another C function | 310 | 265 |
| C function nodes | 39,991 | 39,993 (+`isTpaActive` and two more the old parse dropped) |
| edges | 183,164 | 183,168 |

Edge-set delta, rows keyed with `resolvedBy`: 117 `fuzzy` calls into `pid.c` gone and 118 `exact-match` calls in their place — the same 117 callers now resolve through the strategy that should have had them (they were declined as unreachable closures, #1230, and picked up by the fuzzy fallback at 0.5), plus one call onto the recovered `isTpaActive`; 45 `contains` edges moved from `resetPidProfile` to the file node, 48 file-level `contains` in their place. Nothing else moved.

The 265 that remain nested are a different shape — `#if`-damaged STM32 HAL sources (`stm32h7xx_ll_utils.c`, `…_ll_tim.c`), where both preprocessor branches are kept and the braces do not balance — and are the deferral case the kernel route already documents; not this PR's.

## Tests

`extraction.test.ts`: the blank itself (offsets and line count preserved, `);` keeps its column, designators gone, `log(.5)` and `a == b` untouched) and an end-to-end index of a 120-field macro call followed by two functions — before the change `resetProfile` ran to the last line and `g` was missing; after it the three functions have their own extents and top-level qualified names. Suite: 624/624.

Interaction with #1718: that PR gates C/C++ out of the nesting check on the resolver side, which is the right guard regardless (C has no nested named functions) and still covers the 265; this PR fixes the extraction that produced the nesting for the designated-initializer case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -201,6 +201,8 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 #### Symbols, tests and the viewer
 
+- **A C macro call written with designated initializers no longer swallows every function after it.** Betaflight resets each config struct with `RESET_CONFIG(type, dst, .field = value, …)`, a shape the C grammar cannot parse; past a hundred or so fields its error recovery ran the enclosing function to the end of the file, the next function vanished from the index and every later one was filed under the first, where name matching then treated it as an unreachable closure. The argument list of such a call is now blanked before parsing, offsets kept, so the file's functions come out with their real extents. On that tree 45 functions in `pid.c` alone moved back to top level and their 117 callers resolve at exact-match confidence. Re-index after upgrading. (#1729)
+
 - **Files under an `e2e/` directory count as tests.** Their calls no longer appear as production callers in Steps, dead-code and test badges.
 
 - **Production code under a `samples` or `examples` package path is no longer treated as test code.** A Kotlin or Java project whose package path runs through `com/google/samples/…` (Now in Android, for one) had nearly every file counted as a fixture, so the Map opened on `build-logic`, the entry points hid the app, and dead-code and test badges were wrong. Only the project layout above a `src/` folder decides now; the package path below it never does.
--- a/__tests__/extraction.test.ts
+++ b/__tests__/extraction.test.ts
@@ -11671,6 +11671,59 @@ describe('C/C++ kernel-port preParse blanks (R7a)', () => {
     expect(blankLoneMacroLines(bare)).toBe(bare);
   });
 
+  it('blankCDesignatedMacroArgs empties a designated-initializer macro call, offsets kept (#1729)', async () => {
+    const { blankCDesignatedMacroArgs } = await import('../src/extraction/languages/c-cpp');
+    const src = [
+      'void resetProfile(profile_t *p)',
+      '{',
+      '    RESET_CONFIG(profile_t, p,',
+      '        .pid = { [PID_ROLL] = PID_ROLL_DEFAULT, [PID_YAW] = { 50, 75 } },',
+      '        .limit = 500, // trailing comma follows',
+      '    );',
+      '    log(.5);',
+      '    OTHER_MACRO(a == b, c);',
+      '}',
+    ].join('\n');
+    const out = blankCDesignatedMacroArgs(src);
+    expect(out.length).toBe(src.length);
+    expect(out.split('\n').length).toBe(src.split('\n').length);
+    expect(out).toContain('RESET_CONFIG(');
+    expect(out).not.toContain('.pid');
+    expect(out).not.toContain('PID_ROLL');
+    // The closing `);` keeps its column; the argument lines are spaces.
+    expect(out.split('\n')[5]).toBe('    );');
+    expect(out.split('\n')[3]).toBe(' '.repeat(src.split('\n')[3].length));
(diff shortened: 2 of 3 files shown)


colbymchenry:
Superseded by https://github.com/colbymchenry/codegraph/pull/1755 — same blanking approach, rebased onto current `main`, with the issue’s tiny trailing-comma fixture as the fail→pass e2e (verified on Linux for both wasm and kernel arms). Closing in favor of that PR.

colbymchenry:
Superseded by #1755.

PR #1734, opened 2026-09-07
fix(extraction): a Go method carries the exportedness of its name

danusha2345:
Standalone off `main` (`b9ca4b7`), one commit, both extraction arms. The extractor defect noted in #1731.

## What

`extractMethod` never set `isExported`, so every Go method was indexed as unexported whatever the case of its name — 148 of 148 on a 42-file tree, `func (w *rotatingLogWriter) Close()` included — while plain functions carried the flag. `codegraph-kernel/src/go.rs` mirrored the omission literally (`// extractMethod passes no isExported`).

A method that is a top-level declaration (`methodsAreTopLevel` — Go is the only such extractor) now receives the extractor's `isExported` exactly as a function does, and `go.rs` sets `is_exported` from the same name-case rule in `extract_method`. Class members in other languages keep the flag unset on purpose: their reachability is the class's, and the JS/TS `isExported` walks the parent chain, so applying it to methods would re-mark every member of an exported class and diverge the tsjs kernel walker.

## Measured

Same Go/Kotlin/JS tree, kernel rebuilt from this checkout and confirmed loaded with `CODEGRAPH_KERNEL_DEBUG=1`:

| Go methods | base | this |
| --- | ---: | ---: |
| exported | 0 | **41** |
| unexported | 148 | 107 |

`Close` / `String` → exported, `close` → not. Functions unchanged (135 / 175). The wasm arm (`CODEGRAPH_KERNEL=0`) gives the same flags.

## Tests

`extraction.test.ts`: a four-symbol Go file (`Close`, `flush`, `Open`, `helper`) asserting the flag on both methods and both functions; passes on the wasm arm and with the kernel loaded. `kernel-scaffold`, `kernel-deep-nesting`, `kernel-grammar-parity`: 31 passed with the rebuilt kernel.

Relationship to #1732: that PR judges Go visibility by the name's case precisely because this flag was unusable; with this in, the flag and the case agree, and #1732's rule can later read the flag.


danusha2345:
Rebased onto current `main` (`2f1a99d`): head `38d2b25`. Conflicts were only the CHANGELOG and the neighbouring #1729 tests in `extraction.test.ts`. Kernel rebuilt; the Go test passes on both arms. Once this lands, the by-name-case rule #1745 carries for Go can read the flag instead.

danusha2345:
Rebased onto current `main` (`b715eb6`): head `18b4b30`, one commit. CHANGELOG was the only conflict. Kernel rebuilt; the Go extraction tests pass on both arms (kernel and `CODEGRAPH_KERNEL=0`).

PR #1735, opened 2026-09-07
fix(resolution): a receiver-less JS/TS call never binds to a method

### LINKED ISSUE #1714: A bare call resolves onto the enclosing method when a same-file function shares its name
In TypeScript and JavaScript a bare call — `serialize(...)` with no receiver — can never bind to a class method. Methods require `this.`, a receiver, or a destructured/bound reference. But when a method and a module-scope function in the same file share a name, the bare call inside the method resolves onto the method itself, as a self-edge.

## Reproduction

One file:

```ts
function serialize(value: string): string {
  return value.trim();
}

export class Record {
  constructor(private readonly raw: string) {}
  serialize(): string {
    return serialize(this.raw);   // the free function above
  }
}
```

Index and resolve:

```
Record::serialize --calls--> Record::serialize
  { "confidence": 0.4, "resolvedBy": "exact-match", "refName": "serialize" }
```

Expected: `Record::serialize --calls--> serialize` (the module-scope function).

## Why it lands there

`matchByExactName` gets both same-named nodes as candidates. Both are in the ref's file and both are lexically reachable, so `findBestMatch` scores them, and its same-file line-proximity term (`score += max(0, 20 - distance / 10)`) favours whichever definition is nearer the call site. The enclosing method always is — the call is inside it. So the method wins its own call.

The `0.4` confidence is `computePathProximity` returning `< 30` for a same-file match, which reads oddly on its own and may be a second, smaller thing worth a look.

## Suggested rule

For a **receiver-less** ref in the JS/TS family, drop `method` candidates before scoring. The language guarantees the ref is not a method call: it can only be a local, a module-scope binding, or an import.

This has to be language-gated. Java, C# and Python-with-`self` aside, several languages *do* let an instance method be called bare from inside the class, so the same filter would be wrong there. In JS/TS it is unambiguous.

A narrower variant, if the broad rule is too blunt: only refuse the candidate when it is the ref's own enclosing method — that kills the self-edge without changing anything else.

## Scope

Distinct from #1691, which resolves `this.<field>.<method>()` on the field's declared type — that path has a receiver and this one has none, which I think is what @danusha2345 meant by "a different receiver-less path".

**Correcting my own earlier description of this:** I reported it as a method calling a same-named *imported* free function. That was wrong, and it matters for anyone reproducing it. I re-tested three variants against #1691's head (`15c7ea6`); only the same-file one misresolves:

| shape | result |
| --- | --- |
| barrel re-export (`export * from`) | correct, `import` @0.9 |
| namespace import (`fmt.serialize`) | correct, `import` @0.9 |

So the import resolver already handles the imported cases; the gap is only where no import statement exists to consult.

I counted ten of these on a private ~600-file repo indexed on #1691's head. That count is from that earlier run and I have not re-measured it; the reproduction above is the part I have verified directly.

Happy to open a PR for whichever rule you prefer.

danusha2345:
Fixes #1714. Standalone off `main` (`b9ca4b7`), one commit.

## What

Two rules for a **receiver-less** JS/TS call, both read from the call site's own source since the extractor emits `this.m()` / `super.m()` under the bare method name:

1. **A bare call never binds to a method** (the rule the issue proposes). The text at the ref's column is the call expression; when it starts with the name itself and nothing but whitespace, an operator or an opener precedes it, `method` nodes leave the candidate set before ranking, and `matchFuzzy` declines a lone `method` survivor. `serialize(this.raw)` inside `Record.serialize` resolves to the module-scope `function serialize`; `this.serialize()` (recursion) and `other.serialize()` are unchanged.

2. **A name the file binds itself has no cross-file candidate.** Once methods stop competing, the function that was out-ranked steps in — on vite, `resolve()` inside `new Promise((resolve, reject) => …)` landed on `config.ts`'s `resolve`. A parameter, a `const`/`let`/`var` (destructuring included), a `function` or `class` declared in the file shadows every other file's symbol of that name; none of these is a node the graph holds (a parameter, a const bound to a call result, `const updateDiag = () => …`), so this is the only way the matcher can know. `const { x } = require('./m')` / `= await import('./m')` binds an import, not a shadow, and is excluded; a parameter list has to consist of parameters, so a string argument containing the word does not match.

Both memoised per context and cleared with the other name-matcher memos.

## Measured

vitejs/vite `8492422` and a 114-Kotlin / 42-Go / 96-JS Electron app, edge sets keyed with `resolvedBy`.

**Standalone against `main`:**

| corpus | LOST | GAINED |
| --- | ---: | ---: |
| vite | 1,018 | 124 |
| Electron app | 185 | 3 |

vite's removals by target: `import(…)` → `module-runner/runner.ts::import` 264, `test(…)` → a fixture's `test` 309, `resolve(…)` → `PluginContainer.resolve` 158, `transform(…)` → `PluginContainer.transform` 59, `next(…)` → `bin/vite.js::next` 53, `log(…)` → a spec's `log` 46, and the rest of that shape. The Electron app's: `const updateDiag = () => shell.updateDiag(…)`, `const logLine = (s) => …`, `const w = (...frames) => …` — every sampled row a local arrow the graph has no node for. The 124 gained on vite are 80 `resolve` bound by `import { resolve } from 'node:path'` (the bare-import case #1715 declines) and 41 `log(…)` onto `playground/hmr-ssr/event.d.ts`'s `declare global { let log }`, which is the right target; 3 are minified-library noise.

**Stacked** on #1715 + #1718 + #1720 + #1732 (all four merged, this on top): vite **LOST 522, GAINED 42** (the 41 `log` globals + 1), the app **LOST 103, GAINED 3** (minified). The `node:path` promotions are gone, as expected.

## Tests

`__tests__/bare-call-no-method.test.ts`, whole pipeline over source fixtures: the issue's `Record.serialize` → free `serialize` (self-edge gone, function edge present); `this.serialize()` keeps the self-edge; `other.serialize()` keeps the method; a Promise-executor `resolve`, a module-scope `const transform = makeTransform()` and a factory's `const now = options.now || …` produce no cross-file edge; a destructured `require` and a string argument containing the name still resolve across files. `resolution.test.ts`, `frameworks-integration`, `pr19-improvements`, `ts-this-field-call`: 255 unchanged. `tsc` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -201,6 +201,8 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 #### Symbols, tests and the viewer
 
+- **A bare call inside a JavaScript or TypeScript method no longer resolves to the method itself.** When a method and a module-scope function share a name, `serialize(this.raw)` written inside `Record.serialize` means the function, but the nearest same-named definition won the tie and the graph recorded the method calling itself. A call written without a receiver can never reach a method in JS/TS, so methods are no longer candidates for it; `this.serialize()` and `other.serialize()` resolve as before. (#1714)
+
 - **Files under an `e2e/` directory count as tests.** Their calls no longer appear as production callers in Steps, dead-code and test badges.
 
 - **Production code under a `samples` or `examples` package path is no longer treated as test code.** A Kotlin or Java project whose package path runs through `com/google/samples/…` (Now in Android, for one) had nearly every file counted as a fixture, so the Map opened on `build-logic`, the entry points hid the app, and dead-code and test badges were wrong. Only the project layout above a `src/` folder decides now; the package path below it never does.
--- a/__tests__/bare-call-no-method.test.ts
+++ b/__tests__/bare-call-no-method.test.ts
@@ -1,0 +1,148 @@
+/**
+ * In JS/TS a receiver-less call can never bind to a class method: `serialize(x)`
+ * inside `Record.serialize` means the module-scope function, and the method
+ * itself — which the same-file proximity term used to pick, producing a
+ * self-edge — is not a candidate (#1714). `this.serialize(x)` still is.
+ */
+
+import { describe, it, expect, afterEach } from 'vitest';
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+import CodeGraph from '../src/index';
+
+let tempDir: string;
+let cg: CodeGraph | null = null;
+
+async function callsFromMethod(source: string, methodName: string): Promise<string[]> {
+  tempDir = fs.mkdtempSync(path.join(os.tmpdir(), 'codegraph-1714-'));
+  fs.writeFileSync(path.join(tempDir, 'record.ts'), source);
+  cg = await CodeGraph.init(tempDir, { index: true });
+  cg.resolveReferences();
+  const from = cg.getNodesByKind('method').find((n) => n.name === methodName)!;
+  expect(from).toBeDefined();
+  return cg
+    .getOutgoingEdges(from.id)
+    .filter((e) => e.kind === 'calls')
+    .map((e) => cg!.getNode(e.target))
+    .filter((n): n is NonNullable<typeof n> => !!n)
+    .map((n) => `${n.kind}:${n.qualifiedName ?? n.name}`);
+}
+
+afterEach(() => {
+  cg?.close();
+  cg = null;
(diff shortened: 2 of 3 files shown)


colbymchenry:
Superseded by https://github.com/colbymchenry/codegraph/pull/1759 — rebased your fix onto current main (resolved conflicts with sealed-module / cross-file visibility guards). Same approach; credit retained. Thanks @danusha2345!

colbymchenry:
Closed in favor of https://github.com/colbymchenry/codegraph/pull/1759 (rebased landing of this fix onto current main).

PR #1736, opened 2026-09-07
fix(extraction): collapse a damaged C `#if` group to its first branch before parsing

danusha2345:
Stacked on #1733 (its commit comes first; rebases to one commit once that lands). Together they close the C extraction extent problem behind #1729: after #1733 a 2,109-file betaflight tree still had 265 C functions the graph filed inside another function; after this, 5.

## What the 265 were

Not one cause, and mostly not real functions. 192 of them are literally *named `if`*, the rest `(NVIC_PRIO_TIMER)`, `(reg)`, `ATMEL_DEVICE_MATCH` — phantoms. Verified shape by shape with the shipped `tree-sitter-c.wasm`:

1. **A `#if` branch that begins with `else`.** tree-sitter-c parses each preprocessor branch as a run of block items, and `else` cannot start one, so the `if (cond) { … }` that follows is read as a K&R implicit-int **function definition named `if`**, nested in the enclosing function:
   ```c
   void f(int id) {
       if (id == 1) { a(id); }
   #ifdef USE_V
       else if (id == 2) { b(id); }
   #endif
       else { c(id); }
   }
   ```
   (`current.c`, `stm32*_ll_usart.c`, `_ll_tim.c`, `_ll_gpio.c`, `flash_w25n.c`, …)
2. **A branch that ends with a bare `if (…)` / `else if (…)` header** whose `{` body sits after `#endif` — the STM32H7 `ll_utils.c` latency tables, 24 phantoms in one file.
3. **Branches that leave braces unbalanced** (CMSIS `arm_mat_*`, `smartport.c`, `bus_spi_ll.c`): here the enclosing function vanished entirely and its locals leaked to file scope — `arm_mat_mult_q15.c` had **zero** function nodes.
4. **All-caps block macros in statement position** — `ATOMIC_BLOCK(NVIC_PRIO_MAX) {`, `PG_FOREACH(reg) {` — which `blankCStatementMacroCalls` matched in lowercase only.

## The fix

`blankCUnbalancedConditionalBranches`: for a `#if … #else/#elif … #endif` group where any branch shows one of shapes 1–3, keep the first branch not written `#if 0` verbatim and blank every other branch **and** the group's directive lines (continuations included) to spaces, newlines kept. Balanced groups are untouched; innermost groups first. It runs at the tail of `preParseCSource`, after `restoreDirectiveLines`, for the same reason `blankCNamedVariadicDefineDots` does — it edits directive lines deliberately. `blankCStatementMacroCalls` also accepts `[A-Z_][A-Z0-9_]*` now (PascalCase still excluded, for the C++-constructor reason in its comment). The C++ grammar parses both snippets cleanly, so the C++ path is left alone; the pre-parse precedes the kernel route point, so the kernel needs no change.

## Measured

betaflight fork, 2,109 `.c` files, wasm arm:

| | #1733 | this |
| --- | ---: | ---: |
| C functions nested inside another C function | 265 | **5** |
| C function nodes | 39,993 | 39,772 (260 phantoms gone, ~39 real functions back) |
| `fuzzy` calls | 154 | 37 |

Edge-set delta against the #1733 index, rows keyed with `resolvedBy`: **1,263 lost / 729 gained**. Lost: 709 `exact-match` calls whose *source* was a phantom (`if@…`) or a file node holding leaked locals, 490 file-level `contains` for those locals (`pInA`, `row`, `flag`), 53 references, 11 function-refs. Gained: 671 `exact-match` calls from the recovered functions (`USBD_LL_Init → HAL_PCD_Init`, `LL_TIM_DeInit → …`, `processSmartPortTelemetry → erpmToRpm`) and 43 `contains` for them. Functions that now exist with their real extents: `spiInternalInitStream/StartDMA/StopDMA` (`bus_spi_ll.c`), `processSmartPortTelemetry` (522–903), `esc4wayProcess` (429–929), `arm_mat_mult_q15`. Ten lost and ten gained rows read against source: every lost row is a re-attribution from a phantom or a file node, not a real call gone.

The 5 that remain: four `if SILABS_DEVICE_MATCH {` / `else if ATMEL_DEVICE_MATCH {` in `serial_4way.c` (a paren-less macro as an `if` condition) and one `STATIC_DMA_DATA_AUTO union { … } testBuffer;` in `flashfs.c` — each its own narrow blank, not folded in here.

## Tests

`extraction.test.ts`: a unit test of the blank (length and line count preserved, a continuation line blanked, the first branch kept, the second branch and the directives blanked, a balanced `#ifdef` untouched, `#if 0` keeps the live branch, identity on a file without groups) and an end-to-end index where a function after a damaged group used to be a phantom `(NVIC_PRIO_MAX)` and now has its own extent and top-level name; the end-to-end test fails without the change. `extraction.test.ts` + `resolution.test.ts`: 815, `tsc` clean.


danusha2345:
Rebased onto current `main` (`2f1a99d`): head `9ce9e30`, one commit. The #1733 commit this was stacked on is dropped since #1755 landed the same blanking; the `#if` collapse applies cleanly on top of it. `tsc` clean; the C extraction tests pass.

danusha2345:
Second commit, `b0ed0d1`, from running the branch on two more C trees.

**jq lost `main`.** `src/main.c` has `#ifdef WIN32 int umain(…) { #else int main(…) { #endif` — two function heads in one group. Keeping the *first* branch kept the WIN32 head, and `main` (289–730) was gone from the index. The survivor is now the **default-build** branch: a group that opens on a feature test (`#ifdef X`, `#if defined(X)`, `#if X`) and has an `#else` keeps the `#else` branch; `#ifndef` / `#if !…` keep the first; `#if 0` the live one. jq: `main` back, phantom `if` functions 1 → 0, fuzzy 23 → 0.

**Two more damage shapes** (betaflight, the 7 phantom `if`s the first cut left): a branch that is a fragment of an expression — `|| (hopamp4 == NULL)` inside an `if (…)` in the STM32 HAL, `isEmpty(cmdline) ||` before the rest of a condition in `cli.c` — and a branch that opens with `} else if (…) {`, closing a block it did not open. Betaflight (2,109 C files, main at `cd4e65b` which has #1755): phantom `if` 203 → **0**; edges −1,339 / +1,429 against main, the removals re-attributions off file-level nodes and the additions the members of enums that now parse (`IRQn_Type`). Fixtures for all three shapes are in the test; the first test's expectations flipped to the `#else` branch.

danusha2345:
Merged current `main` (`1f4379d`, post-#1804) into this branch: head `1c261bc`. One conflict, the tail of the C pre-parse pipeline: #1804 wraps it in the raw-string restore, this PR adds the `#if`-group collapse. Composed as restore ∘ collapse ∘ variadic-define ∘ directive-restore — the collapse is length-preserving, and it counts braces with raw-string bodies still masked, so a brace inside `R"(...)"` can't trip it. `tsc` clean; `extraction`, `cpp-raw-string-delimiter-haserror` and `cpp-raw-string-preparse-1505` pass (697 tests).

danusha2345:
Merged current `main` (`7a23648`) into this branch: head `8dff394`. The only conflict was the CHANGELOG entry; no source conflict. `tsc` clean, the branch's tests pass.

Issue #1533, opened 2026-08-09
extend language coverage with 15 additional languages and markup formats


mochizzan:
#### Context

CodeGraph currently supports **20 languages** through its native Rust parsing kernel:

- TypeScript, JavaScript/TSX/JSX, Java, Python, Go, C, C++, Rust, C#, Ruby, PHP, Swift, Kotlin, Scala, Dart, R, Lua, Luau

Additional languages/markup formats (Svelte, Liquid, Pascal/Delphi) are available via the Tree-sitter/WASM path.

While this coverage is solid for mainstream web, mobile, and backend stacks, significant gaps remain in:

- Functional & JVM-ecosystem languages  
- Infrastructure-as-Code  
- Smart contracts  
- Widely-used config/markup formats in modern monorepos  

These languages are currently either completely ignored by the graph or fall back to a generic path with no meaningful symbols or edges.

---

#### Proposal: 15 New Languages/Formats

| # | Language/Format | Category | Priority Rationale |
|---|-----------------|----------|--------------------|
| 1 | Elixir | Functional / BEAM | Popular for real-time backends (Phoenix); currently not covered |
| 2 | Erlang | Functional / BEAM | Natural pairing with Elixir; often coexists in umbrella projects |
| 3 | Haskell | Functional | Widely used in internal compiler/tooling codebases |
| 4 | OCaml | Functional | Used in infra-critical tools (parts of Rust/JS tooling ecosystem) |
| 5 | Clojure | Functional / JVM | JVM ecosystem gap despite existing Java/Kotlin/Scala support |
| 6 | Groovy | JVM | Extremely common in Gradle build scripts; direct impact on dependency-graph resolution |
| 7 | Objective-C | Mobile (iOS legacy) | Many legacy iOS codebases still mix Swift and Objective-C |
| 8 | Zig | Systems | Growing rapidly as a C alternative |
| 9 | Perl | Scripting / Legacy | Still widely used in legacy systems & DevOps scripting |
| 10 | Julia | Scientific computing | Gap in data-science/scientific ecosystem (complements existing R support) |
| 11 | HCL (Terraform) | Infrastructure-as-Code | Critical for DevOps/infra repos; currently not parsed structurally |
| 12 | Solidity | Smart Contract | Growing Web3/blockchain ecosystem; no graph representation for contract calls/inheritance |
| 13 | GraphQL SDL | Schema / API | Enables query-to-resolver resolution; highly relevant to the `getImpactRadius` feature |
[...]

PR #1739, opened 2026-09-07
feat: add Perl language support

### LINKED ISSUE #1738: Add Perl language support (.pl, .pm, .t) — grammar available, package scoping needs care
### Context

CodeGraph has no Perl support today — `.pl` / `.pm` files aren't detected, so a Perl codebase indexes to zero symbols. Perl is item **#9** on the umbrella language-coverage request in #1533; this is a focused issue for just that item, with a PR attached.

Perl is still the backbone of a lot of long-lived infrastructure: test harnesses, build and release tooling, DevOps automation, and large in-house systems that have accumulated for decades. Those are exactly the codebases where nobody remembers how anything connects, and where a graph is worth the most.

### Why Perl is worth doing

Adding it turned out to be cheap. The `tree-sitter-perl` project publishes an **official prebuilt `tree-sitter-perl.wasm`** on its v2.0.0 release, and it's **ABI 15** — so it drops straight into the existing vendored-grammar path with no emscripten or Docker build, and none of the ABI-13 shared-heap corruption that has bitten other grammars.

### What makes Perl interesting to model

Two things don't fall out of generic tree-sitter handling, and getting them wrong quietly produces a useless graph:

**1. Package scoping.** The dominant idiom scopes by *following sibling*, not by containment:

```perl
package Sample::Widget;

sub new  { ... }
sub name { ... }
```

`new` and `name` are **siblings** of the package node, not children. Generic parent/child containment attaches them to the file and loses the `Sample::Widget::new` qualified name entirely.

And a package name is **always absolute** — this declares `Inner`, never `Outer::Inner`:

```perl
package Outer;
package Inner { sub b { ... } }
```

**2. Method calls.** `$obj->method()` parses as `method_call_expression` with `invocant` / `method` fields — none of the `function` / `object` / `name` fields the generic call path looks for. Left alone, it takes the **invocant** as the callee, so `$obj->configure()` records a call to `$obj` and `configure`'s callers are invisible.

### Where Perl can't be answered statically

Perl resolves a lot at run time, and this is where a naive implementation actively hurts. Each of these was measured emitting **wrong** edges on a real 913-file Perl codebase:

- **Chained invocant** — `$self->service_db()->attribute_add(%opts)` dispatches on whatever `service_db()` returned. Emitting the bare name `attribute_add` let it exact-match the *enclosing same-named sub*, producing hundreds of wrong self-referential call edges.
- **Dynamic method** — `$obj->$method()` emitted `$method`, which name-matching bound to the **variable** of that name — an edge to something that isn't callable.
- **`SUPER::`** — dispatches to a parent's method, but with no inheritance/MRO model the bare name binds to the child's *own override*, i.e. precisely the method Perl will not run.
- **Builtins** — `push`, `die`, `print`, `join` parse as ordinary calls and can never resolve to a project symbol.

Consistent with the project's "silent beats wrong" principle, none of these should emit an edge.

### Ambiguous extensions

`.t` is Perl's standard test suffix but is also Raku/Terra/Turing; `.cgi` is as often Python or shell. A shebang naming another interpreter should win over the extension — the same content-based disambiguation `.h` → C/C++/Objective-C already does.

### Proposal

Wire Perl through the existing vendored-wasm path: `.pl`, `.pm`, `.t`, `.pod`, `.psgi`, `.cgi`, with packages as namespaces, subroutines with qualified names, `use`/`require` imports (pragmas filtered), `use parent`/`use base` inheritance including `qw()` lists, `use constant`, package-level variables, and call/instantiation edges — with the run-time-only cases above deliberately left silent.

Projects using non-standard Perl extensions can opt in through `codegraph.json`'s `extensions` map, so nothing unusual needs to be hardcoded.

I have this working and will open a PR against this issue.

Related: #1533 (item 9).

netbrah:
Closes #1738. Item **9** of #1533.

Adds Perl to the extraction pipeline. No new dependency and no build step: `tree-sitter-perl` publishes an **official prebuilt `tree-sitter-perl.wasm`** on its v2.0.0 release, so it drops into the existing vendored-grammar path. It's **ABI 15**, so it's clear of the shared-heap corruption older ABI-13 grammars cause.

`scripts/add-lang/check-grammar.mjs` → **PASS** (ABI 15, 30/30 clean parses in the multi-grammar runtime).

## Files

| File | Why |
|---|---|
| `src/extraction/wasm/tree-sitter-perl.wasm` | Vendored grammar (official v2.0.0 release asset) |
| `src/extraction/languages/perl.ts` | New extractor |
| `src/extraction/grammars.ts` | Grammar file, extension map, vendored list, display name, shebang disambiguation |
| `src/extraction/languages/index.ts` | Register the extractor |
| `src/types.ts` | `'perl'` in `LANGUAGES` |
| `src/extraction/tree-sitter.ts` | Perl call branch (see below) |
| `__tests__/extraction.test.ts` | 27 Perl tests |
| `README.md` / `CHANGELOG.md` | Docs |

## The two things generic handling gets wrong

**1. Package scoping.** The dominant idiom scopes by *following sibling*, not containment:

```perl
package Sample::Widget;

sub new  { ... }   # a SIBLING of the package node, not a child
sub name { ... }
```

Generic parent/child containment attaches these to the file and loses `Sample::Widget::new` entirely. So `visitNode` intercepts `source_file` and walks children itself, opening and closing namespace scopes.

A package name is also **always absolute** — this declares `Inner`, not `Outer::Inner`:

```perl
package Outer;
package Inner { sub b { ... } }
```

which means **at most one package scope is ever open**: a new package *replaces* the current one rather than nesting under it. The walk therefore threads one shared cell and restores the enclosing package when a block ends. It's recursive, so block-form (`package Outer { package Inner { ... } }`) and block-local (`{ package Inner; ... }`) both scope correctly.

Packages map to `namespace`, not `class` — a Perl package *is* a namespace; whether it's also a class depends on `bless`. This pays off directly: Perl's `::` separator matches `buildQualifiedName`'s join exactly, so a `Sample::Base::helper` call site matches the `namespace::sub` qualified name for free.

**2. Method calls need a core branch.** `$obj->method()` parses as `method_call_expression` with `invocant` / `method` fields — none of the `function` / `object` / `name` fields the generic path expects. It would take the **invocant** as the callee, so `$obj->configure()` recorded a call to `$obj` and `configure`'s callers were invisible. This mirrors the existing Ruby branch immediately above it.

## What is deliberately NOT emitted

Each of these was measured producing **wrong** edges against a real 913-file Perl codebase, so following "silent beats wrong" they emit nothing:

- **Chained invocant** — `$self->service_db()->attribute_add(%opts)` dispatches on what `service_db()` returned. The bare name `attribute_add` exact-matched the *enclosing same-named sub*: **hundreds** of wrong self-referential call edges, the single largest source of them.
- **Dynamic method** — `$obj->$method()` emitted `$method`, which name-matching bound to the **variable** of that name — an edge to something that isn't callable.
- **`SUPER::`** — dispatches to a parent's method, but with no inheritance/MRO model the bare name binds to the child's *own override*, exactly the method Perl will not run.
- **Builtins** — `push` / `die` / `print` / `join` parse as ordinary calls and can never resolve to a project symbol.
- **Lexicals inside subs** (named *and* anonymous) — `my $self = shift;` is in nearly every sub of every Perl file; indexing them multiplies node count for symbols no cross-file query can use. Anonymous subs create no graph node, so this is checked by AST ancestry rather than the scope stack.

## Ambiguous extensions

`.t` is Perl's standard test suffix but also Raku/Terra/Turing; `.cgi` is as often Python or shell. A shebang naming another interpreter now wins over the extension, following the existing `.h` → c/cpp/objc content-check precedent:

```perl
#!/usr/bin/env python3   # a .cgi file — no longer claimed as Perl
```

Projects with non-standard Perl extensions can opt in via `codegraph.json`'s `extensions` map, so nothing unusual is hardcoded.

## Validation

Real-world corpus — a **913-file** production Perl codebase (~150k LOC of OO Perl):

| | |
|---|---|
| Index time | 18.8s |
| Nodes | 13,560 |
| Edges | 44,427 |
| Edge kinds | 25,805 `calls` · 12,647 `contains` · 1,977 `imports` · 790 `instantiates` · 415 `extends` |
| Parse failures | 0 |

Effect of the precision rules on that corpus:

| | before | after |
|---|---|---|
| Self-referential call edges | 962 | **498** |
| Permanently-failed references | 26,125 | **18,272** |
| Node count | 13,560 | 13,560 (no explosion) |

The remaining 498 self-loops were spot-checked and are **genuine recursion** (e.g. `$self->_traverse_for_pool_id($node->{config})` inside `sub _traverse_for_pool_id`) — the wrong edges went, the correct ones stayed.

Tests: **27 Perl tests**, and the full suite is green (**4,080 passed, 0 failed**).

## Notes

- Scale beyond ~1k files hasn't been measured yet; a much larger corpus run is in progress and I'll post numbers.
- Frameworks (Mojolicious, Catalyst, Dancer) aren't wired — no `route` nodes yet. Worth a follow-up.
- Perl's `AUTOLOAD` and symbol-table manipulation (`*{"${class}::$name"} = sub {...}`) are dynamic-dispatch frontiers left uncovered on purpose; per the coverage playbook, partial bridging would be worse than none.


netbrah:
## Validation at scale + retrieval-quality audit

Ran the extractor against a large private production Perl codebase (NetApp ONTAP test infrastructure) — as promised in the PR description. Also pushed two follow-up commits from what it turned up.

### Scale

| | |
|---|---|
| Files indexed | **165,153** |
| Nodes | **2,244,942** |
| Edges | **10,065,434** |
| Wall clock | 39 min (full resolution) |
| Peak RSS | 13.4 GB (worker pool); main process flat at ~20 MB |
| Parse failures | 0 |

No crash, no memory growth in the coordinator, WAL checkpointing normally throughout.

A 913-file subset used for the quality work below indexes in **4.3s** → 13,290 nodes / 41,535 edges.

### Call-edge precision (913-file subset, 25,796 call edges)

| Bucket | Count | Share |
|---|---|---|
| Target name has exactly 1 definition | 16,929 | 65.6% |
| Target name ambiguous (2–5 defs) | 4,569 | 17.7% |
| Target name ambiguous (6+ defs) | 3,312 | 12.8% |
| Target is a `constant` | 986 | 3.8% |

- **Textual confirmation: 400/400 (100%)** — for a random sample of call edges, the target's name really is on the recorded source line. Zero hard false positives.
- **84.6% of call edges are cross-file** (21,815 vs 3,981), so the graph is genuinely connecting across modules rather than just within a file.
- The 986 `calls` → `constant` edges are **correct, not FPs**: a Perl `use constant` *is* a subroutine, so calling one is a real call.
- **99.93% of call sites emit exactly one edge.** Only 19 sites in the whole index fan out to multiple same-named definitions, so the graph isn't padded with speculative alternatives.

Precision work driven by this data (all TDD, in the original commit): dropping chained invocants (`$self->a()->b()` inside `sub b`), dynamic dispatch (`$obj->$method()`), `SUPER::`, and Perl builtins cut self-referential call edges by 48% and unresolved refs by 30%. Remaining self-loops were spot-checked and are all genuine recursion.

### Query latency (warm, in-process — the path an MCP agent actually takes)

| | cold | warm |
|---|---|---|
| `getCallers` | 0.17 ms | **0.03 ms** |
| `getCallees` | 0.10 ms | **0.03 ms** |

Graph open 23 ms; name → node search median 0.98 ms.

### Follow-up commit in this PR

1. **Build fix.** `use constant` handling indexed an array without an existence check, which `tsc` rejects under `noUncheckedIndexedAccess`. vitest transpiles without type-checking, so the unit tests were green while `npm run build` failed — caught only when rebuilding for the scale run. Fixed.

2. **Role composition (`with 'Role'`).** Role::Tiny / Moo / Moose compose a role's subs *into* the consuming package, making a role a genuine method-resolution parent alongside `use parent`/`use base`. But `with` parses as an ordinary function call, not a pragma, so nothing modelled it: the relationship was absent from the graph, and `with` leaked out as a call to a nonexistent sub. Package-level `with` now emits an `extends` reference per role, splitting `qw()` lists the same way `use base qw(A B)` is. A `with` inside a subroutine is left alone. On the corpus this recovered the inheritance edges for a class composing 20+ roles.

Tests: **31 Perl cases**, full suite **4083 passed / 0 failed**.

### Known gap (not addressed here — flagging for maintainer judgement)

For an ambiguous method name whose invocant type isn't statically known, resolution falls back to bare-name matching and doesn't consult the `extends` chain. Concretely, `$self->log(...)` in a class that composes a `log` role resolves to a same-named sub in an unrelated package instead of the composed role — even though the role edge now exists in the graph.

That's the cross-language name-matcher rather than anything Perl-specific, so I deliberately kept it out of a language-addition PR to avoid a 30-language blast radius. Happy to open a separate issue, or to take a crack at it here if you'd prefer.

netbrah:
One more gap found while demoing the tooling on the real corpus, now fixed in this PR.

### `codegraph affected` returned nothing for Perl

Perl marks a test by its **extension**, not by a name segment — `prove` runs `t/*.t`. None of the default patterns (`.spec.`, `.test.`, `__tests__/`, `tests?/`, `e2e/`, `spec/`) match `t/resource.t`, so `affected` reported **zero** affected tests for every Perl project, even though the dependency edges were sitting right there in the graph.

Adding `/\.t$/` to the defaults. On the production corpus, asking what a single API module affects:

```
$ codegraph affected lib/CTL/API/Resource.pm
before:  No test files affected by the changed files.
after:   Affected test files (99):
           t/unit/api/resource.t
           t/unit/api/client.t
           t/unit/api/testbed.t
           ...
```

I matched the **extension** rather than the `t/` directory on purpose: a Perl `t/` tree also holds shared helper modules (`t/test_libs/*.pm`) that aren't tests and shouldn't be reported as such. There's a test for each direction.

### Also now covered: the per-project extension override

Perl in the wild carries house-specific extensions — the corpus I validated against uses `.thpl` for ~47k test files and `.cdefs` for ~94k more. Those deliberately stay out of the shipped `EXTENSION_MAP` and go through `codegraph.json`'s `extensions` map instead. Worth noting that map validates its value against the supported-language list, so **mapping anything to `"perl"` only became legal with this PR**. Added end-to-end coverage that a `.thpl` file mapped that way indexes as Perl, resolves its call edges, and is picked up by `affected --filter`.

### Sample output, since it's the core of the capability

`callers` on a production symbol, every result verified against source:

```
$ codegraph callers resource_add          # 11 results
  client_add                lib/CTL/API/Client.pm:269    -> :326  $self->resource_add(%opts, 'type' => 'client')
  testbed_add               lib/CTL/API/Testbed.pm:789   -> :809  $self->resource_add(return_args_definition => 1)
  tglab_basic_resource_add  lib/CTL/API/TGLab_Basic.pm:64 -> :99  return $self->resource_add(%opts)
  ...
```

All cross-file and cross-package: `$self->resource_add()` in `CTL::API::Client` reaching `CTL::API::Resource` through the `use parent qw(...)` chain on `CTL::API` — the hop grep can't make.

`callees` reaches into `CTL::Roles::General::service_db` and `CTL::Roles::Args_definition::_check_if_return_args_definition_set`, which are only reachable because of the role-composition edges added in the previous commit.

Tests: **35 Perl-related cases**, full suite green (the two failures on a loaded run are the pre-existing `ui-server-api` / `ui-highlight` wall-clock budget assertions; both pass in isolation and neither touches this change).

PR #1745, opened 2026-09-08
fix(resolution): a definition its language makes file-local is not a cross-file target (#1731)

### LINKED ISSUE #1731: Kotlin, Go, Rust: a private / unexported / non-pub definition is accepted as a cross-file name-match target

colbymchenry:
## Summary
Rebases and lands [PR #1732](https://github.com/colbymchenry/codegraph/pull/1732) (`danusha2345`) onto latest `main` (post-#1742/#1743), with a small changelog tweak citing #1731 and clarifying the C source-vs-header rule.

Fixes #1731 (Kotlin / Go / Rust). Also covers #1730 (C `static` in another translation unit).

Name matching accepted any same-named definition as a cross-file target. `isVisibleAcrossFiles` now declines a candidate in another file when:

| language | rule |
| --- | --- |
| C / C++ | function defined in a **source** file (`.c`/`.cc`/…) whose definition carries `static` — header `static inline` stays visible |
| Kotlin, Java, C#, Swift, Scala, Dart, PHP | `visibility === 'private'` |
| Go | lowercase identifier and a different directory (by name case — extractor `isExported` is unset for methods; see #1734) |
| Rust | non-`pub` and the ref is outside the item's module subtree; child→ancestor and `impl Trait for Type` methods stay resolvable |

Applied in `ReferenceResolver` **after** `matchReference` settles (and inside `matchFuzzy`), so rejection leaves the ref unresolved instead of falling through to a fuzzy namesake.

Credits: implementation from #1732 by @danusha2345 / Claude.

## Test plan
- [x] Linux repro on `main`: Kotlin `editor.apply()` → private `apply`, Go cross-package `fail`, Rust sibling private `count` all wrongly resolved
- [x] Same fixtures on this branch: all three declined (`KOTLIN_CALLEES []`, `GO_CALLEES []`, `RUST_SIBLING_CALLEES []`)
- [x] `CODEGRAPH_KERNEL=0 npx vitest run __tests__/cross-file-visibility.test.ts` — 11 passed (Node 22.19)
- [x] `CODEGRAPH_KERNEL=0 npx vitest run __tests__/resolution.test.ts` — 189 passed


colbymchenry:
Related open PR #1734 fixes Go method `isExported` in the extractor; this PR intentionally judges Go by name case so it does not depend on #1734. Once #1734 lands the flag and case agree.

PR #1759, opened 2026-09-08
fix(resolution): a receiver-less JS/TS call never binds to a method

### LINKED ISSUE #1714: A bare call resolves onto the enclosing method when a same-file function shares its name

colbymchenry:
Fixes #1714.

Lands/rebases #1735 onto current main (was CONFLICTING).

Receiver-less JS/TS calls never bind to methods; local bindings shadow cross-file names. Keeps main visibility guards.

Verified fail->pass on Linux; bare-call-no-method 5/5; resolution 198/198; tsc clean.

Credit: @danusha2345 (#1735).


danusha2345:
Thanks for landing it — but the head you rebased (`fd26889`) carries a regression I had found and fixed on the branch an hour before this merged, and it is on `main` now: at `cd4e65b`, `object-literal-methods.test.ts` (`resolves callers of store actions across files`) and `ui-steps-api.test.ts` (2 tests) fail. Rule 2 reads `const { fetchUser } = useStore.getState()` and `const setZipUri = useStore((s) => s.setZipUri)` as local definitions and drops the store-action edges. Fix as a follow-up: #1760 (a same-name re-binding is not a local definition; one new fixture, the three tests pass again, vite numbers in the description).

PR #1761, opened 2026-09-08
fix(kotlin): a property carries its visibility, and a private class hides its members

danusha2345:
Follow-up to #1745 (the Kotlin / Go / Rust file-local rule, #1731). Two commits, off `main` at `cd4e65b`.

## What

1. **A Kotlin property records its visibility modifier.** `private val` / `private var` / `internal` at class or object scope were indexed with no `visibility`, so the Kotlin rule in `isVisibleAcrossFiles` — which declines a `private fun` in another file — read every property as public. Both the WebAssembly extractor and the native kernel now read the `modifiers` child the way they already do for functions; the parity fixture gains a `private val` and an `internal var`.

2. **A member of a `private` class or object is file-local too.** `private class Clock(var value: Long) { fun now() = value }` in a test: the method is public, the class is not, so nothing outside the file can name `now`. For the languages where `private` is file-local (Kotlin, Java, C#, Swift, Scala, Dart, PHP) a candidate whose enclosing class/object/interface/enum in its file is `private` is declined — by line range among the file's nodes, memoised per node.

Applied where #1745 applies, after the whole pipeline settles, so a rejection leaves the reference unresolved.

## Measured

A Kotlin + JS + Go Electron/Android app (114 Kotlin files), `main` at `cd4e65b` vs this branch, wasm arm, edge rows keyed with `resolvedBy`: **LOST 57 / GAINED 0**.

- 24 from the property fix: a validator's `token(...)` onto another class's `@Volatile private var token`; JS `now()` / `bytes()` / `gate()` callers (renderer, a Python tool, the Go signaling server) onto Kotlin test fixtures' `private val`s; `changed`, `handler`, `resources`, `onStatus`, `listener`, `moving`.
- 33 from the private-owner rule: 3 `now()` calls in the gimbal control onto the test's `private class Clock`, and 30 `send` / `cancel` / `close` / `request` calls from `TelemetryUplink.kt` / `WebRtcTransport.kt` (instance-method resolution) onto `TelemetryUplinkTest.kt`'s `private class FakeWebSocket` / `ServerSocketListener`.

A second Kotlin app (FreeFCC, 1,726 nodes): 0 / 0. `pulp` Rust, Python and C corpora: untouched (the rule is gated to the `private`-is-file-local languages).

## Tests

`cross-file-visibility` +2 (a `private var` and an `internal val` carry their modifier and the private one is declined; a public fun of a private class is declined), `kernel-kotlin-parity` / `kernel-grammar-parity` on the rebuilt kernel, `resolution` 198/198, `frameworks-integration`; `tsc` clean. Full suite on Linux: the viewer-less `codegraph ui` files, plus the three store-action tests that fail on `main` since #1759 and that #1760 fixes.


danusha2345:
Merged current `main` (`8c9c476`) into this branch: head `6ea4eb3`. The only conflict was the CHANGELOG entry; no source conflict. `tsc` clean, the branch's tests pass.

danusha2345:
Merged current `main` (`9181dd1`, post-#1802) into this branch: head `3d50e3d`. Both arms of the Kotlin property hook conflicted with #1802's initializer walk — resolved by keeping #1802's walk verbatim and adding this PR's `visibility` to the `createNode` / `create_node` call (TS + kernel). Kernel rebuilt; `tsc` clean; `cross-file-visibility`, `kernel-kotlin-parity` and `extraction` pass (671 tests).

danusha2345:
Merged current `main` (`4453310`) into this branch: head `ddbb8c2`. The only conflict was the CHANGELOG entry; no source conflict. `tsc` clean, the branch's tests pass.

danusha2345:
Merged `main` (58c07e8) into this branch; the only conflict was the CHANGELOG. Kernel rebuilt for the parity fixtures, `tsc` clean, full suite passes on Linux (kernel arm; a first pass had 5-second timeouts under load, all green on a rerun).

PR #1763, opened 2026-09-08
fix(resolution): a store-action binding is not a local shadow

### LINKED ISSUE #1762: Regression (#1759): store-action bindings treated as local shadows; main fails its own object-literal-methods test
### Summary

`#1759` ("a receiver-less JS/TS call never binds to a method") makes `isLocallyBoundJsName` treat a **Zustand-style store binding** as a local rebinding, so bare calls to store actions stop resolving. This regresses the repo's own test suite: **`main` at `cd4e65b` fails one of its own tests.**

### Reproduction

Clean checkout of `main` at `cd4e65b`, `npm ci`, full build, then:

```
npx vitest run __tests__/object-literal-methods.test.ts
```

```
FAIL __tests__/object-literal-methods.test.ts > object-literal method resolution (end-to-end)
   > resolves callers of store actions across files (destructured + chained getState())

AssertionError: expected [] to include 'loginFlow'
```

`__tests__/ui-steps-api.test.ts` fails the same way (2 tests):

- `buildSteps > walks a screen through its handler, the bridge, the event, the store and the request`
- `screen regions > a screen names every step's region: the screen body for its own code, inherited down the walk`

### Cause

`isLocallyBoundJsName` carves out `require(...)` / `import(...)` on the right-hand side, because an imported name is a node elsewhere rather than a shadow. A store binding is the same situation and is not carved out:

```js
const { fetchUser } = useStore.getState();   // action lives in the store file
const setZipUri = useStore((s) => s.setZipUri);
```

Both name the store's own action, so the bare call that follows means that action. `matchStoreAccessorChain` already recognises the accessor shapes; `isLocallyBoundJsName` now discards those candidates before it runs.

### Fix

Extend the existing carve-out to the two accessor shapes plus the selector read. PR follows.

The selector pattern is deliberately narrow — one named parameter and a plain member **read** of the same name — so a genuine local binding such as `const now = options.now || (() => Date.now())` still shadows. That case is covered by `bare-call-no-method`, which stays green.

### Verification

Full suite on the same machine, same build, pristine vs patched:

| | pristine `cd4e65b` | with the fix |
| --- | --- | --- |
| Test files | 13 failed | **7 failed** |
| Unique failing tests | 29 | **24** |

`object-literal-methods` and both `ui-steps-api` tests go green, and **no test that passed before fails after**. The remaining failures are pre-existing on this machine (Windows, native kernel not built) and byte-identical across both runs.

bompus:
Fixes #1762.

`#1759` made `isLocallyBoundJsName` treat a Zustand-style store binding as a local rebinding, so bare calls to store actions stopped resolving. `main` at `cd4e65b` currently fails its own `__tests__/object-literal-methods.test.ts` with `expected [] to include 'loginFlow'`, plus two tests in `__tests__/ui-steps-api.test.ts`.

### The change

`isLocallyBoundJsName` already carves out `require(...)` / `import(...)` on the right-hand side, because an imported name is a node elsewhere rather than a shadow. This extends that same carve-out to store bindings, which are the same situation — the action is a node in the store file, so the bare call that follows means it:

```js
const { fetchUser } = useStore.getState();   // accessor shapes matchStoreAccessorChain already knows
const { fetchUser } = get();                 // inside the factory
const setZipUri = useStore((s) => s.setZipUri);  // selector reading the same name
```

A selector returning a *different* name is a genuine rebinding and still shadows.

The selector pattern is deliberately narrow: one named parameter, and a plain member **read** of the same name (`(?!\s*\()`). `const now = options.now || (() => Date.now())` matches neither clause — no parameter, and `.now()` is a call — so it stays the local binding this predicate exists to find. `bare-call-no-method` covers that case and stays green.

### Verification

Full suite, same machine and same build, pristine `cd4e65b` vs this branch:

| | pristine | this branch |
| --- | --- | --- |
| Test files | 13 failed | **7 failed** |
| Unique failing tests | 29 | **24** |

`object-literal-methods` and both `ui-steps-api` tests go green, and **no test that passed before fails after**. The remaining 24 are pre-existing on this machine (Windows, native kernel not built) and identical across both runs.

### DIFF
--- a/src/resolution/name-matcher.ts
+++ b/src/resolution/name-matcher.ts
@@ -704,7 +704,31 @@ function isLocallyBoundJsName(name: string, filePath: string, context: Resolutio
   );
   let bound = false;
   for (const m of source.matchAll(declRe)) {
-    if (!/^\s*(?:await\s+)?(?:require|import)\s*\(/.test(m[1] ?? '')) { bound = true; break; }
+    const rhs = m[1] ?? '';
+    // A store binding names the store's own action for the same reason an
+    // import does: the action is a node in the store file, so the bare call
+    // that follows means it rather than shadowing it. Two shapes, both naming
+    // the action explicitly — the accessors matchStoreAccessorChain already
+    // recognises (`const { fetchUser } = useStore.getState()`, `get()` inside
+    // the factory), and a selector that reads the same name off the state
+    // (`const setZipUri = useStore((s) => s.setZipUri)`). A selector returning
+    // a different name is a genuine rebinding and still shadows.
+    //
+    // The selector shape is exact on purpose: one named parameter, and a plain
+    // member READ of the same name. `const now = options.now || (() =>
+    // Date.now())` satisfies neither — no parameter, and `.now()` is a call —
+    // so it stays the local binding this predicate exists to find.
+    const selectorRe = new RegExp(
+      '\\(\\s*[\\w$]+\\s*\\)\\s*=>\\s*[\\w$]+(?:\\?\\.|\\.)' + n + '\\b(?!\\s*\\()'
+    );
+    if (
+      !/^\s*(?:await\s+)?(?:require|import)\s*\(/.test(rhs) &&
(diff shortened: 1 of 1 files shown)


bompus:
Duplicate of #1760, which fixes the same regression. Closing; verification data moved to #1760.

PR #1791, opened 2026-09-08
fix(dart): index extension type members in both extraction arms (#1784)

### LINKED ISSUE #1784: Dart: `extension type` members stopped being indexed in #1780 — the `method_signature` gate has no language check
## Dart: `extension type` members stopped being indexed in #1780 — the `method_signature` gate has no language check

`ee83636` (#1780, fixing #1638) gates a bodiless `method_signature` behind `isInsideClassLikeNode()`. That is correct for TypeScript, but Dart spells *ordinary implemented* methods `method_signature` too (`src/extraction/languages/dart.ts` → `methodTypes: ['method_signature', 'constructor_signature']`), and a Dart 3 **`extension type`** body is not class-like. Its members now miss the method branch and are dropped entirely.

### Reproduction

```dart
extension type MetersT(double value) {
  double get km => value / 1000;
}
```

```ts
const r = await extractFromSource('probe.dart', src, 'dart');
r.nodes.filter(n => n.kind !== 'file').map(n => `${n.kind}:${n.qualifiedName}`)
```

| commit | result |
| --- | --- |
| `8c04734` (parent) | `["function:km"]` |
| `ee83636` (#1780) | `[]` |
| `e720f6c` (current main) | `[]` |

Same working tree, same `node_modules`, same probe; only the checkout changed. `km` goes from indexed to invisible.

### Scope — what is and isn't affected

Ran the same probe across the neighbouring constructs on both commits:

| case | `8c04734` | `e720f6c` |
| --- | --- | --- |
| Dart `extension type` | `function:km` | **`[]`** |
| Dart `extension` | `class:StringHelpers`, `method:StringHelpers::shout` | unchanged |
| Dart `mixin` | `class:Logger`, `method:Logger::log` | unchanged |
| Dart `class` | `class:Widget`, `method:Widget::build` | unchanged |
| TS `interface` | `interface:Iface` | `interface:Iface`, `method:Iface::go` |
| TS type literal | `method:Handle::stop`, `type_alias:Handle` | unchanged |

So **#1780 does what it set out to do** — the TS interface-member row is the fix working, and the type-literal row confirms it did not reintroduce the phantom free function from #359. `extension`, `mixin` and `class` are unaffected because their bodies satisfy `isInsideClassLikeNode()`. `extension type` is the only construct that regressed.

### Cause

`extension_type_declaration` is a node type in the bundled `src/extraction/wasm/tree-sitter-dart.wasm`, but it appears nowhere in `src/extraction/languages/dart.ts` — not in `classTypes` (only `class_definition`), not in `extraClassNodeTypes` (only `mixin_declaration`, `extension_declaration`), not in `enumTypes`. Nothing mints a class-like parent for its body, so `isInsideClassLikeNode()` is `false` for `double get km`.

Note the pre-#1780 result was already imperfect — `km` was minted as a top-level `function` rather than a member of `MetersT`, since no class-like node exists for the extension type. The regression is that it went from mis-kinded to absent.

### Two possible fixes

1. **Narrow the gate by language**, which is what we carry downstream:

   ```ts
   const SIGNATURE_METHOD_LANGUAGES = new Set(['typescript', 'tsx', 'arkts']);
   // ...
   && (!SIGNATURE_METHOD_NODE_TYPES.has(nodeType)
     || !SIGNATURE_METHOD_LANGUAGES.has(this.language)
     || this.isInsideClassLikeNode())
   ```

   Restores `function:km` and keeps every TS row above green.

2. **Teach Dart about extension types** — add `extension_type_declaration` to `extraClassNodeTypes`. Strictly better, because it also fixes the pre-existing mis-kinding: `km` would become `method:MetersT::km` rather than a top-level function. I have not measured this one, so treat it as a suggestion rather than a tested patch.

They are not exclusive; (2) alone would fix this case, but (1) is the general guard for any other language that reuses `method_signature` for implemented methods.

### Why CI didn't catch it

There appears to be no Dart extraction fixture exercising `extension type` (or, as far as I can see, a Dart parity fixture at all). We caught this only because our downstream fork runs a Rust-kernel-vs-TypeScript parity suite over a Dart torture fixture, where it surfaced as a node-count mismatch (91 vs 90) across `torture.dart` and `TortureCtors.dart` including their CRLF variants. A `extension type` case in the Dart extraction tests would pin it.

Environment: Windows 11, Node v26.8.1, `vitest` 2.1.9, bundled `tree-sitter-dart.wasm`.

inth3shadows:
Fixes #1784.

Dart spells an ordinary implemented method `method_signature` — the same node type TypeScript uses for a bodiless interface member. #1780 gated that node type behind `isInsideClassLikeNode()` so a TS interface member could not mint a phantom free function. That is right for TS, but a Dart 3 `extension type` body is not class-like, so every member inside one failed the gate and was dropped.

## The declaration was missing too, not just its members

The issue reports the members disappearing. Reproducing it turned up more: on `main`,

```dart
extension type Meters(double value) {
  double get km => value / 1000;
  void show() { print(km); }
}
```

extracts **`function:show`** and nothing else — no `Meters`, no `km`. So `extension type` was never a class-like node to begin with; #1780 only made the omission visible by making the members depend on it.

That is why this takes the reporter's **option 2** rather than option 1. Adding `extension_type_declaration` to Dart's `extraClassNodeTypes` fixes both halves at once and gives the members the right owner — `method:Meters::km` instead of a loose top-level `km`. Option 1 would have restored the members while leaving the extension type itself invisible and its members mis-kinded.

## Mirrored into the kernel — the two arms had already diverged

Dart is default-routed to the native kernel, so a change made only in the TypeScript arm never runs in a published bundle. And the arms were already disagreeing here: the kernel still emitted `function:km` while wasm emitted nothing, which is why **`kernel-dart-parity` is red on `main` today** — all four cases, `torture.dart` and `TortureCtors.dart` including their CRLF variants. `codegraph-kernel/src/dart.rs` listed the class-like node types in two places (`enclosing_type_name` and the extract dispatch); both now include it.

Worth noting for the "why didn't CI catch it" question in the issue: the fixture and the parity suite already exist in this repo — `__tests__/fixtures/kernel-parity/TortureCtors.dart:11` is exactly this construct. The suite simply never runs without a staged `.node`, since every `kernel-*-parity` suite `describe.skipIf`s itself. From source, `npm test` reports **188 skipped**; after `scripts/build-kernel.sh`, **10**, and those four turn red.

## Verification

Linux, Node 22.23.2, kernel built and staged:

```
kernel-dart-parity        18/18    (4 of these fail on main)
dart-extension-type        2/2     (new; both fail without the change)
tsc --noEmit              clean
npm test                  252 files, 4373 passed, 3 failed
```

The three remaining failures — `object-literal-methods` and two in `ui-steps-api` — reproduce on a clean `upstream/main` worktree and are unrelated to this change.

The new test pins both directions: the extension type and its members are indexed, and `extension` / `mixin` / `class` bodies are unaffected.

Note for whoever merges: this changes the kind and qualified name of extension-type members (`function:km` → `method:Meters::km`), so those nodes get new ids and a re-index is needed to see them — the CHANGELOG entry says so.

Thanks to @bompus for the diagnosis; this is their option 2, measured, plus the kernel half.


bompus:
Verified on Linux (WSL2, Node v26.8.2, from-source kernel build) at `133f9fa`: `dart-extension-type.test.ts` (2 new tests) green, and the full `kernel-dart-parity.test.ts` file green 18/18 — including the 4 `torture.dart` / `TortureCtors.dart` cases that are red on current `main` per #1784 (kernel emitting `function:km` with no wasm counterpart). Both arms now agree on `method:Meters::km`.

This was reported from our side in #1784, so noting for the record: the fix matches the diagnosed cause (language check on the `method_signature` gate) and the parity suite we pointed at now passes with it.

bompus:
Follow-up to my earlier validation comment: that run used Node 26, outside this project's supported runtime range. I have now repeated the checks on **Node 24.21.0**, Linux x86_64 / Ubuntu WSL2, at the same exact PR head **`133f9fa30082fc7bc2fec147c395461ca93ce593`**, with a native kernel built from that checkout.

- `dart-extension-type.test.ts`: **2/2 passed**.
- `kernel-dart-parity.test.ts`: **18/18 passed**, with the native kernel present and `CODEGRAPH_KERNEL_EXPECT=1`; no skipped cases. This includes the LF/CRLF `torture.dart` and `TortureCtors.dart` cases discussed in #1784.
- Explicit WASM fallback (`CODEGRAPH_KERNEL=0`): extension-type tests **2/2 passed**.

Both test runs exited 0, with no native crash or dead-worker output. The extension-type assertions retain `class:Meters`, `method:Meters::km`, and `method:Meters::show`; the parity suite agrees on nodes, edges, and unresolved references.

This supplies supported-runtime evidence for the earlier conclusion. I did not rerun the full suite, real-repository sweeps, Windows validation, or the merge-base in this follow-up.

inth3shadows:
Thank you for redoing this on a supported runtime — and for catching that the first pass was on Node 26 in the first place. Independent confirmation at the exact head, with the native kernel present and `CODEGRAPH_KERNEL_EXPECT=1` so the parity cases actually ran rather than skipped, is worth more than anything I can assert about my own branch.

Re-checked just now: the head is still `133f9fa`, `MERGEABLE`, and 0 behind `main` @ `3ed73bc`. Nothing has moved under it since your run, so your evidence still describes what is on the branch.

Nothing outstanding from my side — happy to rebase or re-run anything if it sits long enough to drift.

PR #1814, opened 2026-09-09
fix(extraction): name a function bound through a curried wrapper (#1747)

### LINKED ISSUE #1747: TypeScript/JavaScript: a function bound through a wrapper call gets no node unless the wrapper is a React hook
CodeGraph already names an anonymous function after the `variable_declarator` it
is bound through — `react_hook_bound_name` does exactly this. The mechanism reads
as a general one, but it only runs when the callee is `useCallback`,
`useEffectEvent` or `useEvent`, so the same shape with any other wrapper yields
no function node at all.

## Repro

```ts
declare const Effect: { fn: (n: string) => (b: unknown) => unknown };
declare function useCallback<T>(f: T, d: unknown[]): T;

function helper() { return 1; }

const viaHook   = useCallback(() => { return helper(); }, []);
const viaEffect = Effect.fn("Session.run")(function* () { return helper(); });
```

After `codegraph index`:

```
nodes:  helper · viaHook (function) · viaHook (constant) · viaEffect (constant)
edges:  viaHook -> helper
        b.ts    -> helper          <- the file, not a function
```

`viaEffect` gets a constant node but no function node. Its body is not simply
missing, either: the `helper()` call inside it is attributed to the **file**, so
the file picks up an outgoing call edge that belongs to a function, and
`helper`'s caller list names the file rather than the function that actually
calls it.

## What the two lines have in common

Both are an anonymous function passed as an argument to the call a
`variable_declarator` binds, and the declarator name is available in both. The
only difference is the callee's identity, checked against a literal list:

```rust
// codegraph-kernel/src/tsjs/mod.rs:769 — react_hook_bound_name
if !matches!(hook, "useCallback" | "useEffectEvent" | "useEvent") {
```

```ts
// src/extraction/tree-sitter.ts:396
const REACT_HANDLER_HOOKS = /^(?:React\.)?use(?:Callback|EffectEvent|Event)$/;
```

Other wrappers that land in the same position:

| Shape | Ecosystem |
|---|---|
| `const run = Effect.fn("Session.run")(function* () {})` | Effect-TS |
| `const C = connect(mapState)(function () {})` | Redux |
| `const x = wrap("name")(function () {})` | project-local helpers |

## Scope

The shape is not rare. `sst/opencode` carries 1,056 distinct string-named
`Effect.fn*` wrappers across its packages. Indexing it, 48 of them end up with a
function node — the other 1,008 are functions missing from the graph, each one
also sending its body's calls to a file instead of to a function.

maxmilian:
Fixes #1747.

`reactHookBoundName` already names an anonymous function after the `variable_declarator` that binds it. As @Dshuishui points out, the mechanism reads as a general one — it is bounded to three React hooks only by the callee test. This adds the same shape with a different bound, on both the TypeScript and the Rust kernel paths.

## The bound

Not "any callee". Widening that far would give `const total = useMemo(() => 1 + 1, [])` a function node, and `__tests__/react-hook-handlers.test.ts` already asserts `expect(names).not.toContain('total')` — correctly, since a memo callback is a computation rather than a body worth a node.

The three shapes in the report have something narrower in common: **the callee is itself a call**, i.e. a factory that returns the wrapper.

| shape | callee | admitted |
|---|---|---|
| `Effect.fn("Session.run")(function* () {…})` | `Effect.fn("Session.run")` — a call | ✅ |
| `connect(mapState)(function () {…})` | `connect(mapState)` — a call | ✅ |
| `wrap("name")(() => {…})` | `wrap("name")` — a call | ✅ |
| `useMemo(() => 1 + 1, [])` | `useMemo` — an identifier | ❌ |
| `[1,2].map(() => …)` | `[1,2].map` — a member expression | ❌ |

So the new path is `curriedWrapperBoundName` / `curried_wrapper_bound_name`, kept beside the React one rather than replacing it: the hook allowlist keeps its own meaning and its own tests, and this reads as one additional, independently bounded rule.

## Generators are the second half of the fix

This is worth calling out because it is not visible in the issue's diagnosis. `generator_function` is in `functionTypes` (`languages/typescript.ts:42`, `javascript.ts:6`, `mod.rs:85`), but the guards in `reactHookBoundName`, `react_hook_bound_name` and `declaratorBoundFunction` all accept only `arrow_function | function_expression`.

The report's first repro is `Effect.fn("Session.run")(function* () {…})`, and `function*` is the common form in that ecosystem — **broadening the callee test alone would not have fixed it.** Generators are admitted in the new method and deliberately not added to `reactHookBoundName`: a React handler is never a generator, so nothing is gained there and the existing bound stays exactly as it was.

## Tests

`__tests__/curried-wrapper-handlers.test.ts`, three of them:

1. **the fix** — all four callee shapes (`Effect.fn` with a generator and with an arrow, `connect`, `wrap`) produce a function node named by the declarator.
2. **the guard that makes the bound meaningful** — `useMemo` and `arr.map` still produce none. If these ever start producing function nodes, the callee test has been widened past what this change claims.
3. **the other half of the report** — the half a node count alone does not cover: `refsFrom(viaEffectGen)` contains `helper`, and `refsFrom(file)` does **not**. That is the "a file appeared to call what the function called" symptom.

**Red control** — reverting only `src/extraction/tree-sitter.ts` to `main` and re-running:

```
× names the wrapped function after its declarator, for every callee shape
    AssertionError: expected [ 'helper' ] to include 'viaEffectGen'
× attributes the body calls to the function, not to the file
    AssertionError: no function node for viaEffectGen: expected undefined to be defined
  Tests  2 failed | 1 passed
```

Test 2 passes in both directions, which is what a vacuity guard should do.

## Verification

- `npx tsc --noEmit`: clean.
- `cargo build --release` for the kernel: clean; `npm run build:kernel` stages the binding.
- `__tests__/curried-wrapper-handlers.test.ts` + `__tests__/react-hook-handlers.test.ts`: **8 passed** — the React path is unchanged.
- `__tests__/kernel-tsjs-parity.test.ts`: **21 passed**.
- `node scripts/kernel-parity.mjs` on a file holding all six shapes above: **1/1 files byte-parity, 0 diffs, 0 deferred-to-wasm**. Since the TypeScript side is known to produce the four function nodes, byte-parity is what establishes the Rust side does too.
- Full suite after `npm run build:ui`: **4570 passed, 18 failed, 11 skipped**.

**On those 18 — I checked rather than assuming they were pre-existing.** I ran the same nine files on a clean `upstream/main` worktree: 14 of the 18 fail there identically. The remaining 4 are all `kernel-dart-parity` (`torture.dart` / `TortureCtors.dart`, parity and CRLF), which that baseline run *skipped* because the kernel was not built in it. Building the kernel there and re-running gives **the same 4 failures on unmodified `main`**. So: 18 = 14 + 4, all pre-existing, none of them TypeScript/JavaScript extraction, and this change introduces no regression.

## Notes

- CHANGELOG entry added under `## [Unreleased]` → `### Fixes` → `#### Symbols, tests and the viewer`, in user-facing language, with the re-index note.
- No version bump, no tag, no publish.
- I have not indexed `sst/opencode` to confirm the 1,008-function figure from the report; the mechanism is verified here, that scale claim is @Dshuishui's measurement rather than mine.

https://claude.ai/code/session_01WgtWMywGTyu6twnk1Bu9nP

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -295,6 +295,8 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 - Python calls and file dependencies through `from package import module as alias` now appear in the graph, so renamed imports no longer hide live callers or imported modules. Thanks @JoeyNPP. (#1626)
 
+- **A function bound through a wrapper call is now part of the graph.** `const run = Effect.fn("Session.run")(function* () {…})` — and the same shape with Redux's `connect(mapState)(…)` or a project's own `wrap("name")(…)` — produced no function at all: only React's `useCallback`, `useEffectEvent` and `useEvent` were recognized as naming the function they wrap. The calls inside those bodies were attributed to the file instead, so a file appeared to call what the function called, and asking who calls a helper named the file rather than the function that really calls it. One codebase carried over a thousand of these. Any wrapper that is itself the result of a call now names the function it wraps, generators included; a callback passed straight to a single call, such as `useMemo` or `map`, is a computation and stays anonymous as before. Re-index after upgrading. (#1747)
+
 ## [1.6.0] - 2026-08-26
 
 ### Highlights
--- a/__tests__/curried-wrapper-handlers.test.ts
+++ b/__tests__/curried-wrapper-handlers.test.ts
@@ -1,0 +1,84 @@
+/**
+ * A function bound through a curried wrapper is named by its declarator (#1747).
+ *
+ * `react_hook_bound_name` already names an anonymous function after the
+ * `variable_declarator` that binds it, but only when the callee is one of three
+ * React hooks. The same shape with any other wrapper produced no function node
+ * at all — and the loss is not only a missing node: the body's calls attribute
+ * to the enclosing container instead, so a file or an outer function picks up an
+ * outgoing edge that belongs to the function, and the callee's caller list names
+ * the wrong thing.
+ *
+ * The bound used here is that the callee is itself a call — a factory that
+ * returns the wrapper. That admits `Effect.fn("x")(fn)`, `connect(m)(fn)` and a
+ * project's own `wrap("n")(fn)`, and leaves the single-call forms whose argument
+ * is a computation (`useMemo`, `arr.map`) anonymous exactly as before.
+ */
+import { describe, it, expect, beforeAll } from 'vitest';
+import { extractFromSource } from '../src/extraction';
+import { initGrammars, loadAllGrammars } from '../src/extraction/grammars';
+
+beforeAll(async () => {
+  await initGrammars();
+  await loadAllGrammars();
+});
+
+const refsFrom = (result: ReturnType<typeof extractFromSource>, id: string) =>
+  result.unresolvedReferences.filter((r) => r.fromNodeId === id).map((r) => r.referenceName);
+
+const CODE = `
+declare const Effect: { fn: (n: string) => (b: unknown) => unknown };
+declare function connect(m: unknown): (c: unknown) => unknown;
+declare function wrap(n: string): (c: unknown) => unknown;
+declare function useMemo<T>(f: () => T, d: unknown[]): T;
+
+function helper() { return 1; }
+
+const viaEffectGen   = Effect.fn("Session.run")(function* () { return helper(); });
+const viaEffectArrow = Effect.fn("Session.go")(() => { return helper(); });
+const viaConnect     = connect({})(function () { return helper(); });
+const viaWrapArrow   = wrap("n")(() => { return helper(); });
+
+const total  = useMemo(() => 1 + 1, []);
+const mapped = [1, 2].map(() => helper());
+`;
+
+describe('curried wrapper handlers', () => {
+  it('names the wrapped function after its declarator, for every callee shape', () => {
+    const result = extractFromSource('src/b.ts', CODE);
+    const names = result.nodes.filter((n) => n.kind === 'function').map((n) => n.name);
+
+    // A generator is the common form in the ecosystem this comes from, and it
+    // is the shape the report opens with — `function*` is not accepted by
+    // reactHookBoundName, so broadening the callee test alone would not fix it.
+    expect(names).toContain('viaEffectGen');
+    expect(names).toContain('viaEffectArrow');
+    expect(names).toContain('viaConnect');
+    expect(names).toContain('viaWrapArrow');
+  });
+
+  it('leaves single-call arguments anonymous, so the graph does not gain nodes for computations', () => {
+    const result = extractFromSource('src/b.ts', CODE);
+    const names = result.nodes.filter((n) => n.kind === 'function').map((n) => n.name);
+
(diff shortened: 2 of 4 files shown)


Dshuishui:
Two shapes on `sst/opencode` (`df23b7f`) that this branch does not reach, with what I measured.

The predicate requires the wrapper call's parent to be a `variable_declarator`:

```ts
const declarator = call.parent;
if (!declarator || declarator.type !== 'variable_declarator') return null;
```

In Effect code the wrapper's result often does not land in a declarator.

**Object member.** A service is usually an object a factory returns rather than a top-level const:

```ts
function make() {
  return {
    getMode: Effect.fn("ACP.Session.getMode")(function* (id: string) {
      return lookup(id)
    }),
  }
}
```

**Argument to another call.**

```ts
export default Runtime.handler(
  Commands.commands.api,
  Effect.fn("cli.api")(function* (input) { … }),
)
```

**Measured.** Counting by position rather than by name, since the two approaches name the node differently and a name count would compare nothing: the 1,056 distinct source positions where `Effect.fn*("…")(` is immediately followed by a function literal, against whether the index holds a function node starting at that line.

| | has a node |
|---|---:|
| this branch | 818 / 1056 |
| dropping the declarator requirement — asking, wherever an anonymous function is walked, whether a `fn` / `fnUntraced` call with a string-literal first argument encloses it | 1055 / 1056 |

Of the 238 missed, **zero** are `const` / `let` / `var` bindings — the declarator path itself is complete. 199 are object members and 39 are arguments to another call.

maxmilian:
@Dshuishui — thank you, and for counting by position; a name count would have hidden exactly this. Confirmed on the branch: both the TS and Rust predicates stopped at `variable_declarator`.

I'd split your 238 by what names the function. The 199 object members have a name already in the source — the property key — and this repo already names `key: () => {}` pairs by their key (`extractObjectLiteralFunctions` / `objectKeyName`). So bbdef20 extends the curried-wrapper bound to a `pair` parent, with the same callee-is-a-call rule (`key: useMemo(() => …, [])` stays anonymous). It also covers the member when the object is extracted member by member (an exported object with an inline function in it), where the initializer is never walked. Tests are in `curried-wrapper-handlers.test.ts`, and `kernel-parity.mjs` reports byte-parity on the new shapes.

I re-ran a position count on `sst/opencode@df23b7f`. My pattern is looser than yours: 1,069 positions against your 1,056, because it also catches a doc comment, array elements and arrow bodies. So these numbers are not directly comparable with your table:

| | has a function node starting on that line |
|---|---:|
| 4f2b400 (previous head) | 829 / 1069 |
| bbdef20 | 1028 / 1069 |

The 199 misses that sit behind a `key:` all have a node now. The 41 left are 34 call arguments plus 7 other positions (array elements, arrow bodies, the comment). If you re-run your script on bbdef20, your numbers are the ones that count.

The call arguments have no binding at all. Naming them after the `Effect.fn` string literal would be a new naming rule keyed to one library's API, so I'd rather leave that to @colbymchenry as a follow-up than decide it inside this PR.

Issue #1566, opened 2026-08-18
TypeScript: built-in Map.get/set/has calls resolve to unrelated project methods


netbrah:
## Summary

TypeScript calls to the built-in `Map.get`, `Map.set`, and `Map.has` methods
resolve to unrelated project methods when the project contains exactly one
same-language method with each name.

Both common receiver shapes fail:

- a simple local receiver such as `values.get(...)` is stored as
  `values.get`, then resolves to the sole project `get` method with confidence
  0.7 / `instance-method`; and
- a nested receiver such as `holder.values.get(...)` is extracted as bare
  `get`, then resolves to the project method with confidence 0.9 /
  `exact-match`.

The simple local receiver is the distinct gap demonstrated here. The nested
receiver belongs to the extraction family already reported in #1496 and is
included as a regression boundary rather than as a second new defect.

The resulting callers, callees, impact, and raw graph all report calls that do
not exist. An unresolved external/built-in call would be safer than a confident
wrong project edge.

## Minimal reproduction

One dependency-free TypeScript file:

```ts
export class LRUCache {
  private store = new Map<string, string>();

  get(key: string): string | undefined {
    return this.store.get(key);
  }

  set(key: string, value: string): void {
    this.store.set(key, value);
  }

  has(key: string): boolean {
    return this.store.has(key);
  }
}

export function useLocalMap(): boolean {
  const values = new Map<string, string>();
  values.set("answer", "42");
  values.get("answer");
  return values.has("answer");
}

export function useNestedMap(
  holder: { values: Map<string, string> },
): string | undefined {
  return holder.values.get("answer");
}
```

```console
$ codegraph init

$ codegraph callees useLocalMap --json
{
  "symbol": "useLocalMap",
  "callees": [
    { "name": "set", "kind": "method", "filePath": "repro.ts", "startLine": 8 },
    { "name": "get", "kind": "method", "filePath": "repro.ts", "startLine": 4 },
    { "name": "has", "kind": "method", "filePath": "repro.ts", "startLine": 12 }
  ]
}

$ codegraph callees useNestedMap --json
{
  "symbol": "useNestedMap",
  "callees": [
    { "name": "get", "kind": "method", "filePath": "repro.ts", "startLine": 4 }
  ]
}
```

The stored edge metadata shows two independent fallbacks:

```text
source          target          confidence  resolvedBy       refName
useLocalMap     LRUCache::set   0.7         instance-method  values.set
[...]

PR #1821, opened 2026-09-09
fix(resolution): Python module-qualified calls colliding with builtin methods

inth3shadows:
## Summary
- \`isBuiltInOrExternal\`'s Python built-in-method filter treated any \`x.method()\` as a builtin (\`list.append\`, \`dict.update\`, etc.) whenever \`method\` matched a common collection method name, unless the capitalized receiver matched a known **class**. A real project **module** exporting a same-named top-level function (e.g. \`ledger.append\`) was dropped before \`resolveViaImport\`/\`resolvePythonModuleMember\` ever ran — a false negative (missing edge).
- Separately, the method-call extractor in \`tree-sitter.ts\` only preserved the qualified \`receiver.method\` shape for a plain-identifier receiver. A chained-call receiver (\`rows.setdefault(k, []).append(v)\`) degraded to a bare \`append\` ref, which then exact-matched an unrelated same-named project function — a false positive (fabricated edge).
- Rebased onto current \`main\`; the false-negative half is now largely superseded by \`isPythonProjectModule\` (#1681) already on \`main\`, so this PR's \`src/resolution/index.ts\` change is now just a comment cross-reference. The chained-receiver false-positive fix in \`src/extraction/tree-sitter.ts\` is still needed for Python — the existing TS/JS chain guard (#1566) does not cover it. \`__tests__/resolution.test.ts\` already carries the equivalent test upstream (#1681/#1683/#1748); no test changes needed here.

## Test plan
- [x] Full suite on the rebased branch: 4421 passed / 12 failed / 192 skipped — the 12 failures are pre-existing on a clean, unmodified \`main\` checkout (verified side-by-side, identical failing test names, all in unrelated Next.js/Express/React-Native UI-steps tests)
- [x] \`npm run build && npm test\` green for the two files this PR actually changes

### DIFF
--- a/src/extraction/tree-sitter.ts
+++ b/src/extraction/tree-sitter.ts
@@ -4878,6 +4878,19 @@ export class TreeSitterExtractor {
               // (#1707). Calls inside arguments are visited independently.
               // Mirrored in the kernel's extract_call (tsjs/extractors.rs).
               return;
+            } else if (this.language === 'python' && receiver) {
+              // Any other receiver shape — attribute chain (`self.data`), subscript
+              // (`d[k]`), call chain (`rows.setdefault(k, []).append(x)`) — previously
+              // fell through to a bare methodName. A bare name matching a common
+              // list/dict/str method (`append`, `get`, `update`, ...) can exact-match
+              // an unrelated top-level project function sharing that name, fabricating
+              // a call edge (#66, same class as #1230/#1276, and the Python sibling
+              // of the TS/JS chain guard above (#1566)). Keep the receiver's source
+              // text as a qualifier so the ref can only resolve through
+              // import/module-member resolution, never the bare-name fallback; an
+              // unresolvable qualifier is a silent miss, never a wrong edge.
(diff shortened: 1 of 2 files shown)


inth3shadows:
Closing as redundant — both halves of this fix are already on `main`, independently:

- False negative (`ledger.append` gets no caller): fixed by #1681, via `isPythonProjectModule`.
- False positive (chained receiver `d.setdefault(k, []).append(v)` fabricating an edge): fixed by bb1d309, "Fixes #1683".

Verified directly: built a clean `main` checkout with none of this PR's changes and ran the exact repro from testgraph's issue #66 (`ledger.py:append`, `record.py` calling it via import, `unrelated.py` doing `rows_by_file.setdefault("f", []).append(...)`). Result — `append`'s only caller is the real one (`add_outcome`); the chained `.append()` does not fabricate an edge. No patch needed.

Thanks for landing both fixes — sorry for the duplicate effort, this PR was drafted before I checked for prior fixes on current `main`.

PR #1843, opened 2026-09-11
fix(status): detect committed but unindexed changes

### LINKED ISSUE #1829: status.pendingChanges reports zero for a committed-but-unindexed file — the git fast path never hash-checks what git omits
`status.pendingChanges` hash-checks correctly, but only over the candidate set git nominates — so it reports zero for a **committed but unindexed** file. Committing a change is what makes the signal stop seeing it.

Reproduced twice on 2026-09-10: once on published **1.6.0**, once on a build of **`main` @ `3ed73bc`** (Node 22.23.2). Identical output both times.

## Repro

```bash
mkdir -p /tmp/r/pkg && cd /tmp/r
printf 'def a():\n    return 1\n' > pkg/one.py && touch pkg/__init__.py
git init -q && git add -A && git commit -qm init
codegraph init .
codegraph status --json          # fileCount 2, pendingChanges {added:0, modified:0}

printf 'def b():\n    return 2\n' > pkg/two.py
printf 'def a():\n    return 99\n' > pkg/one.py
codegraph status --json          # fileCount 2, pendingChanges {added:1, modified:1}   correct

git add -A && git commit -qm add
codegraph status --json          # fileCount 2, pendingChanges {added:0, modified:0}   WRONG

codegraph sync                   # Added: 1, Modified: 1
```

Observed:

```
after init:   fileCount 2  pending {added:0, modified:0, removed:0}
uncommitted:  fileCount 2  pending {added:1, modified:1, removed:0}     correct
committed:    fileCount 2  pending {added:0, modified:0, removed:0}     WRONG
then sync:    Added: 1, Modified: 1
```

`fileCount` stays honest at 2 throughout — the index really does have two files and never three. `pendingChanges` is the field that misreports, and only **after** the change is committed. `sync` immediately afterwards finds exactly the work `status` said did not exist.

Seen in the wild the same way: after a merge, `status` reported `pendingChanges: {added: 0, modified: 0, removed: 0}` on a repo where new source files existed on disk with zero nodes and no `files` row. An immediate `sync` found `Added: 3, Modified: 11`.

## Mechanism (verified against `main` @ `3ed73bc`)

The hash comparison inside `getChangedFiles()` (`src/extraction/index.ts:3178`) is **sound**. The defect is one level up, in where the candidates come from:

```ts
const gitChanges = getGitChangedFiles(this.rootDir);       // :3179

if (gitChanges) {
  // === Git fast path ===                                 // :3182
  ...
  for (const filePath of [...gitChanges.modified, ...gitChanges.added]) {   // :3199
    const contentHash = hashContent(content);
    const tracked = this.queries.getFileByPath(filePath);
    if (!tracked) added.push(filePath);
    else if (tracked.contentHash !== contentHash) modified.push(filePath);
```

`getGitChangedFiles` (`:1290`) builds that list from `collectGitStatus` (`:1303`), which runs:

```
git status --porcelain --no-renames -uall                  // :1312
```

A committed file has no porcelain entry, so it never enters the loop and is never hash-compared — regardless of what the DB holds. **Committing is what removes a file from the candidate set.** The check is correct; it is simply never reached.

Two details that shape the fix:

- The **non-git fallback** in the same function (`:3222`, "Fallback: full scan") walks `scanDirectory` and compares against `getAllFiles()`. That path is correct — a working implementation sits ~40 lines below the broken one. This is specific to the git fast path.
- A comment at `:3193` records that untracked (`??`) files had to be hash-compared rather than always counted as added, *"otherwise status reports them as pending forever (see issue #206)."* Same bug class — git state standing in for index state — patched for one case without generalizing.

## Suggested fix

In the git fast path, union the git candidate set with the DB's own file list (`getAllFiles()`), so a file the index tracks is hash-checked whether or not git currently reports it, and a file the index does not track is caught by a scan rather than by git's opinion. The fast path needs to stop treating *"git says nothing changed"* as *"nothing changed."*

If a working-tree diff is genuinely wanted somewhere, it belongs in a differently named field: "pending changes" reads as *"changes this index has not absorbed"*, and that is how it gets used.

## Why it matters beyond the number

This is the second index-currency signal that reports "current" when the index is not — the other being the one discussed in #1798 (`sync` never consults `isIndexStale()`, so an engine upgrade leaves edges stale under files that never changed; I posted a reproduced case there). They are independent mechanisms in different files, but they compound: one hides files that changed and were committed, the other hides edges that went stale under files that did not change. Together they let an index be reported current, freshly synced, and still wrong — with `reindexRecommended: false` sitting alongside. A reader of `status` has no single field to trust.

For what it's worth, a downstream guard that compares `files.indexed_at` against disk mtime — never consulting git — correctly warned `7 source file(s) newer than the index` on a run where `status` reported zero pending changes. The comparison is cheap and it works.

## Environment

```
codegraph   1.6.0 (published bundle) and main @ 3ed73bc built from source
node        bundled runtime / v22.23.2 for the source build
platform    Linux x86_64 (WSL2)
```

Happy to send a PR for the union fix if you want it.

bompus:
Committing an addition, edit, or deletion before indexing currently makes `codegraph status` report zero pending changes. This change compares the full in-scope source inventory with indexed hashes, so pending changes remain visible until sync absorbs them.

Fixes #1829.

The existing inventory comparison now handles Git projects as well as non-Git projects. It retains scope filtering and checks disk existence for removals, since `git ls-files` can still list unstaged deletions. No schema change or new dependency is needed.

Regression tests cover additions, edits, and deletions before and after committing, recovery after sync, committed renames, and graph preservation across repeated no-op syncs. The three addition/edit/deletion cases failed at the post-commit assertion before the fix.

### Validation

Based on upstream `3ed73bc127323e63153bf6ec8354afa82ce36aaf`; head `a5709f47` contains only this fix and its tests.

- **Exact PR branch, WSL/Linux, Node 24.21.0:** 59 tests passed across `sync`, `include-ignored-config`, `git-changed-untracked-dir`, and `extraction-old-git`, with `CODEGRAPH_KERNEL=0`, one worker. TypeScript compilation and diff whitespace checks passed.
- **Deployed downstream integration `3f1c3309`, including this fix and other fork changes:** 4,756 native tests and 4,754 WASM tests passed, zero failures; all five downstream source probes passed. These are integration results, not a full-suite run of this exact PR branch.
- Windows and the viewer build were not tested for this PR.

### Performance tradeoff

Reading and hashing the full inventory costs more than the incomplete Git-status candidate list. A disposable VS Code checkout at `7b7e49c83affacfac726040280da69b4999f3e01` contained **14,486 in-scope files, 166.77 MiB**:

| State | Before median | Fixed median | Fixed min–max |
| --- | ---: | ---: | ---: |
| Clean | 50.2 ms | 614.2 ms | 600.0–642.2 ms |
| One uncommitted edit | 59.1 ms | 676.7 ms | 604.6–780.7 ms |
| Same edit committed without indexing | 51.1 ms (incorrectly clean) | 604.4 ms (correctly modified) | 596.0–634.5 ms |

Measured on WSL/ext4 with Node 24.21.0 and `nice -n 10`: two warm-ups, then ten samples per arm per case, alternating arm order. The actual before/fixed detection methods from downstream `d5607a08` and `6cb53e82` used the same supporting modules and SQLite file records seeded from the checkout's current hashes. Those methods differ only by this fix; no symbol graph was extracted. All returned change sets were asserted.

These are warm-cache **change-detection** timings, excluding CLI startup and other status queries—not end-to-end CLI or cold-storage measurements. The shared host had background services; the Android build had completed before the reported run. An earlier run during build activity was excluded. The measured added cost is about 0.55–0.62 seconds on this repository.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -145,6 +145,8 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixes
 
+- `codegraph status` now reports unindexed additions, edits, and deletions even after they have been committed to Git. (#1829)
+
 - Spring mappings now include every declared path combination and resolve constants declared in the same file, while unresolved paths no longer appear as false root routes. (#1461)
 - `codegraph callers`, `codegraph callees` and `codegraph impact` now resolve qualified names, group results and JSON edges by definition, and accept `--file` to narrow ambiguous names; thanks @ferrine. (#1512, #1656)
 - `codegraph callers`, `codegraph callees` and `codegraph impact` (CLI and MCP) now report missing names with did-you-mean suggestions instead of another symbol's results, and exact matches with no callers stay empty; thanks @uvmplus. (#1473, #1481)
--- a/__tests__/sync.test.ts
+++ b/__tests__/sync.test.ts
@@ -219,6 +219,26 @@ describe('Sync Module', () => {
       }
     });
 
+    it.each(['added', 'modified', 'removed'] as const)(
+      'reports %s files after committing without indexing (#1829)',
+      async (kind) => {
+        const relativePath = kind === 'added' ? 'src/new.ts' : 'src/index.ts';
+        const filePath = path.join(testDir, relativePath);
+        if (kind === 'removed') fs.unlinkSync(filePath);
+        else fs.writeFileSync(filePath, 'export function changed() { return 99; }');
+
+        const expected = { added: [], modified: [], removed: [], [kind]: [relativePath] };
+        // In particular, unstaged deletions remain in git ls-files.
+        expect(cg.getChangedFiles()).toEqual(expected);
+        git('add', '--', 'src');
+        git('commit', '-m', 'changed without indexing');
+        expect(cg.getChangedFiles()).toEqual(expected);
+
+        await cg.sync();
+        expect(cg.getChangedFiles()).toEqual({ added: [], modified: [], removed: [] });
+      }
+    );
+
     it('should detect modified files via git', async () => {
       fs.writeFileSync(
         path.join(testDir, 'src', 'index.ts'),
--- a/__tests__/sync.test.ts
+++ b/__tests__/sync.test.ts
@@ -239,6 +239,31 @@ describe('Sync Module', () => {
       }
     );
 
+    it('detects a committed rename and preserves the graph across repeated no-op syncs', async () => {
+      git('mv', 'src/index.ts', 'src/renamed.ts');
+      git('commit', '-m', 'rename without indexing');
+
+      expect(cg.getChangedFiles()).toEqual({
+        added: ['src/renamed.ts'], modified: [], removed: ['src/index.ts'],
+      });
+      // Status must not discard the last usable graph before sync succeeds.
+      expect(cg.searchNodes('hello').some(r => r.node.filePath === 'src/index.ts')).toBe(true);
+
+      const renamed = await cg.sync();
+      expect(renamed.filesAdded).toBe(1);
+      expect(renamed.filesRemoved).toBe(1);
+      const nodes = cg.searchNodes('hello').map(r => r.node);
+      expect(nodes).toHaveLength(1);
+      expect(nodes[0].filePath).toBe('src/renamed.ts');
+
+      for (let i = 0; i < 2; i++) {
+        expect(cg.getChangedFiles()).toEqual({ added: [], modified: [], removed: [] });
+        const unchanged = await cg.sync();
+        expect([unchanged.filesAdded, unchanged.filesModified, unchanged.filesRemoved]).toEqual([0, 0, 0]);
+        expect(cg.searchNodes('hello').map(r => r.node.id)).toEqual(nodes.map(node => node.id));
+      }
+    });
+
     it('should detect modified files via git', async () => {
       fs.writeFileSync(
         path.join(testDir, 'src', 'index.ts'),
--- a/src/extraction/index.ts
+++ b/src/extraction/index.ts
@@ -3173,53 +3173,10 @@ export class ExtractionOrchestrator {
 
   /**
    * Get files that have changed since last index.
-   * Uses git status as a fast path when available, falling back to full scan.
+   * Compare the full source inventory with the index: git status omits
+   * committed changes that have not been indexed yet (#1829).
    */
   getChangedFiles(): { added: string[]; modified: string[]; removed: string[] } {
-    const gitChanges = getGitChangedFiles(this.rootDir);
-
-    if (gitChanges) {
-      // === Git fast path ===
-      const added: string[] = [];
-      const modified: string[] = [];
-      const removed: string[] = [];
-
-      // Deleted files — only report if tracked in DB
-      for (const filePath of gitChanges.deleted) {
-        const tracked = this.queries.getFileByPath(filePath);
-        if (tracked) {
-          removed.push(filePath);
-        }
-      }
-
-      // Modified + added files — read + hash, compare with DB. Untracked (`??`)
-      // files stay untracked in git even after indexing, so they must be
-      // hash-compared like modified files instead of always counting as added —
-      // otherwise status reports them as pending forever. (See issue #206.)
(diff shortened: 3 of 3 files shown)


inth3shadows:
Opened #1848 as an alternative shape for this same bug — not a competing claim on it, your diagnosis matches mine exactly and you were first. Flagging it here so the two can be compared side by side.

It keeps the git fast path and adds the missing half by stamping the commit the index was built at, then asking git for the committed diff since that commit. All three arms measured on one machine against VS Code at `7b7e49c8` — the same commit your table uses — reading a single index of 14,137 files, timing `getChangedFiles()` in-process, median of 10 with the arm order reversed on a second pass:

| State | `main` @ 3ed73bc | #1843 | #1848 |
| --- | ---: | ---: | ---: |
| Clean | 119 ms | 1470 ms | 132 ms |
| One uncommitted edit | 131 ms | 1488 ms | 136 ms |
| Committed, not indexed | 122 ms (reports 0) | 1547 ms | 146 ms |

Absolute numbers run higher than yours on every arm — different, busier machine — so the ratio is the comparable part.

**One case yours catches and mine does not:** a file hidden with `git update-index --assume-unchanged` and then edited. Your inventory walk sees it; neither `git status` nor `git diff` reports it, so #1848 misses it exactly as `main` does. Verified on all three builds. If that case outweighs the per-call cost, yours is the right fix and mine isn't — happy either way.

bompus:
Thanks for measuring all three arms on one machine — that table is the comparison I'd want a maintainer to see.

I'd take #1848. The per-call cost of the inventory walk is paid by status, every sync, and the watcher's catch-up, and a ~11x hit on a large checkout is a bad trade for catching one rare case. The commit stamp closes the real gap (committed add/modify/delete since the index) at ~1.1x, and the fallback to the full scan when the stamp can't resolve means an old index heals itself after one call.

On the assume-unchanged case: it's real but it's a user opting out of git's own change tracking, and the previous behaviour missed it too, so #1848 is no regression there. If it ever matters, the inventory walk from this PR could sit behind an explicit `status --full` or a periodic check rather than the hot path.

Two things I'd carry over from here if #1848 lands: the committed-rename test, and the disk-existence check on removals so an unstaged delete that `git ls-files` still lists is reported as removed. If both are already covered by your nine cases, ignore this.

Happy for the maintainer to close this in favour of #1848. I'll leave it open until they pick a shape.

bompus:
Correction to my last paragraph: both carry-over items are already covered in #1848 — the committed-rename case has its own test there, and an unstaged delete is still reported by the unchanged `git status` path, so the disk-existence check only matters on the inventory walk #1848 doesn't take. Disregard that paragraph.

bompus:
Closing in favour of #1848. My fork has adopted that shape; the full native and WASM suites pass on it with the same tests. The inventory walk here is available in history if the assume-unchanged case ever needs it.

bompus:
Rebased onto current `main` (`ba3c21e`) — new head `6a14646`.

Reconciliation notes (upstream re-landed #1829 in a different form since this was written against `3ed73bc`):

- `src/extraction/index.ts`: kept upstream's stamp-based git fast path (`INDEXED_AT_COMMIT_KEY` / `indexedDirtyPaths` / `getGitChangedFiles(root, sinceCommit)`) — the rest of the tree now depends on it, and the full-scan fallback already carries this PR's disk check (`fs.existsSync` for unstaged deletions). The PR's full-scan-only replacement is dropped as superseded.
- `CHANGELOG.md`: dropped this PR's #1829 line — `main` already documents it ("Status now detects committed but unindexed changes and restored edits without scanning every source file").
- Kept both test commits as guards: `reports %s files after committing without indexing (#1829)` and the committed-rename + repeated no-op sync recovery test.

Verified on the rebased tree: `__tests__/sync.test.ts` 50/50 green (includes upstream's own 5-test `committed-but-unindexed` suite), `tsc --noEmit` clean, native kernel rebuilt.

bompus:
Disregard my rebase note above — I acted on a stale digest listing. This PR is closed in favour of #1848 (and upstream has since landed its own #1829 fix), so no rebase was needed. Fork branch restored to `a5709f4`; no changes pushed.

Issue #1840, opened 2026-09-11
TypeScript: split() on an awaited Promise<string> result resolves to an unrelated class method (v1.6.0)


yaoguohh:
## Summary

With CodeGraph 1.6.0 and a fresh two-file TypeScript index, calling `split()` on the awaited result of a function explicitly returning `Promise<string>` creates a false call edge to an unrelated project class's `split()` method.

The unrelated class is not imported or instantiated by the caller. The incorrect edge is persisted in SQLite, so this is not just a display problem.

## Environment

- CodeGraph CLI: `1.6.0`
- macOS, Apple Silicon
- Fresh `codegraph init --yes .` index; 2 TypeScript files, 6 nodes, 6 edges
- Reproduced independently of the original project

## Minimal reproduction

Create an empty directory containing these two files.

`strings.ts`:

```ts
export async function listPaths(): Promise<string> { return "a\0b"; }
export async function snapshot(): Promise<string[]> {
  const listed = await listPaths();
  return listed.split('\0');
}
```

`pane.ts`:

```ts
export class PaneManager {
  split(): string { return "new pane"; }
}
```

Run:

```sh
codegraph init --yes .
codegraph node snapshot --file strings.ts
```

## Actual result

```text
Calls → listPaths (strings.ts:1), split (pane.ts:2)
```

The persisted `calls` edge has:

```json
{
  "caller": "snapshot",
  "source": "strings.ts",
  "callee": "PaneManager::split",
  "target": "pane.ts",
  "line": 4,
  "metadata": {
    "confidence": 0.7,
    "resolvedBy": "instance-method",
    "refName": "listed.split"
  }
}
```

## Expected result

Keep the real `snapshot → listPaths` edge. Do not create `snapshot → PaneManager::split`: `listed` is a string, and this call invokes the built-in string method.

If the resolver cannot infer the awaited return type, leaving this member call unresolved is preferable to asserting a call into an unrelated class. Actual calls on a `PaneManager` receiver should still resolve normally.

## Impact and related work

This gives callers/callees and downstream dependency or impact analysis a nonexistent relationship. I initially encountered it while inspecting a filesystem snapshot function, which appeared to call an unrelated terminal-pane splitting method.

[...]

Issue #1839, opened 2026-09-11
C++: local object initialization records type nodes instead of constructor calls


netbrah:
## Summary

C++ local object initialization is not represented consistently in the call graph. Braced and value initialization emit a `class` or `struct` node as a callee instead of the constructor method, while an unbraced default declaration emits no callee. A temporary expression for the same class resolves correctly to the constructor method.

## Environment

- CodeGraph 1.6.0
- Backend: node:sqlite
- Fresh synthetic C++ index

## Minimal reproduction

```cpp
struct Aggregate {
    int value;
};

class WithConstructor {
public:
    WithConstructor();
    explicit WithConstructor(int value);
};

WithConstructor::WithConstructor() {}
WithConstructor::WithConstructor(int value) {}

void aggregate_initialization() {
    Aggregate item{};
}

void constructor_default() {
    WithConstructor item;
}

void constructor_braced() {
    WithConstructor item{};
}

void constructor_value() {
    WithConstructor item(1);
}

void constructor_temporary() {
    WithConstructor();
}
```

Commands:

```sh
codegraph init --yes
CODEGRAPH_NO_WATCH=1 codegraph callees aggregate_initialization -j
CODEGRAPH_NO_WATCH=1 codegraph callees constructor_default -j
CODEGRAPH_NO_WATCH=1 codegraph callees constructor_braced -j
CODEGRAPH_NO_WATCH=1 codegraph callees constructor_value -j
CODEGRAPH_NO_WATCH=1 codegraph callees constructor_temporary -j
```

Actual results:

| caller | callee result |
| --- | --- |
| `aggregate_initialization` | `Aggregate` (`struct`) |
| `constructor_default` | none |
| `constructor_braced` | `WithConstructor` (`class`) |
| `constructor_value` | `WithConstructor` (`class`) |
| `constructor_temporary` | `WithConstructor` (`method`) |

Expected results:

- `Aggregate item{}` has no user-defined constructor and should produce no ordinary `calls` edge.
- Each `WithConstructor` initialization form should point to the indexed constructor method `WithConstructor::WithConstructor`, as the temporary expression already does, never to the class declaration node.

## Why this matters

Type declarations are not executable callees. A type-node edge fabricates an apparent call target, while a missing constructor edge hides a real dependency. The mixed behavior makes callers, callees, and impact analysis inconsistent for common C++ local initialization.

Issue #1838, opened 2026-09-11
C++: macro invocation resolves as a call to a same-named free function in another translation unit


netbrah:
## Summary

A C++ macro invocation is emitted as an ordinary call edge and can bind to an unrelated free function with the same spelling in another translation unit.

## Environment

- CodeGraph 1.6.0
- Backend: node:sqlite
- Fresh synthetic C++ index

## Minimal reproduction

`marker.hpp`:

```cpp
#define TRACE_POINT(value) ((void)(value))
```

`exercise.cpp`:

```cpp
#include "marker.hpp"

void exercise() {
    TRACE_POINT(1);
}
```

`macro_decoy.cpp`:

```cpp
void TRACE_POINT(int value) {}
```

Commands:

```sh
codegraph init --yes
CODEGRAPH_NO_WATCH=1 codegraph callees exercise -j
```

Actual result:

```json
{
  "symbol": "exercise",
  "callees": [
    {
      "name": "TRACE_POINT",
      "kind": "function",
      "filePath": "macro_decoy.cpp"
    }
  ]
}
```

Expected result: the macro invocation should not create an ordinary `calls` edge to a same-named function in another translation unit. It should either be omitted from the call graph or be represented distinctly as a macro expansion.

## Why this matters

A same-spelled function elsewhere can fabricate callers and callees that do not exist in source, making call-graph queries and impact analysis unsafe for macro-heavy C/C++ code.

Issue #1835, opened 2026-09-11
MCP: projectPath queries leave non-default repository indexes without auto-sync


nakisen:
## Summary

Follow-up to #1606 and #1607, addressed by #1614 and released in v1.6.0.

The single-child adoption and diagnostics fixes work as described. The remaining problem is the **multiple indexed children** case: a single MCP server can successfully query each repository through `projectPath`, but those accesses do not start catch-up synchronization or file watching for the repositories. Their indexes remain stale until something else updates them.

The requested outcome is **one MCP server registration/tool surface, separate per-repository indexes, and automatic synchronization for explicitly accessed repositories**, without needing a default repository.

## Environment

- CodeGraph v1.6.0 (still the latest published release when checked on 2026-09-11).
- Windows, native filesystem, PowerShell; no WSL.
- MCP host starts in a coordination repository that gitignores its independent child repositories.
- Each child already has its own initialized `.codegraph/`; the coordination root has none.
- MCP command: `codegraph serve --mcp`, without `--path` or `--no-watch`.

## Reproduction

```text
workspace/                 # MCP host working directory
  .git/
  .gitignore               # contains /service-a/ and /service-b/
  service-a/
    .git/
    .codegraph/
    sample.ts
  service-b/
    .git/
    .codegraph/
    sample.ts
```

1. Initialize CodeGraph inside both child repositories.
2. Start the MCP host from `workspace/` using the command above.
3. Query a known symbol in each child with `codegraph_explore`, passing that child's absolute path as `projectPath`. Queries can read the existing indexes.
4. Add or rename an exported function in each child's source file and save.
5. Wait beyond the default watcher debounce window, then query the new symbols using the same `projectPath` values.
6. The newly added/renamed symbols are not indexed automatically. Manual `codegraph sync` inside each child picks up the changes.

The startup diagnostic correctly reports:

```text
No .codegraph/ at or above <workspace>: no default project, live sync disabled.
Indexed sub-projects found: service-a, service-b. Pass `projectPath` per call, or launch with --path.
```

Passing `projectPath` resolves query routing, but leaves live sync disabled for that target.

## Verification against the installed v1.6.0 engine

To isolate this from the MCP host, I also reproduced the behavior with disposable Git repositories through the installed `MCPEngine` and `ToolHandler` classes. This was an **in-process engine test**, not a separate end-to-end stdio transport test. Exact indexed symbol names were checked after tool dispatch; this was not inferred from response text containing the query string.

[...]

PR #1845, opened 2026-09-11
draft: call evidence, index freshness and Verilog/SystemVerilog — superset being split

danusha2345:
## Status (2026-09-15)

Draft superset branch, kept as a working reference while it is being split into focused PRs. It is based on `3ed73bc` and now conflicts with `main`; it will not be rebased as one piece.

What it contains and where each part is going:

- **JS/TS call evidence and Zustand store actions** (#1760, bompus's #1844, an interface-signature fix for `useStore.getState().reset()`): **superseded** by #1862 on `main`, which covers the same cases; #1760 is closed.
- **Committed-but-unindexed changes / `indexed_at_commit` fast path** (built on #1842 and the #1848 approach, plus guards for `git restore` of an indexed edit, a commit landing during a full index, non-ASCII committed paths, and a failing diff command): the guards are posted as review feedback on #1848 rather than as a competing PR.
- **Atomic edge rebind**: on `main` as #1849.
- **Verilog / SystemVerilog** (port of #402 by @FHYQ-Dong, plus instance/port/signal navigation, generate and procedural scopes, positional and `.*` bindings, read/write access roles, build profiles with include/define context and watcher invalidation, and on-demand `codegraph hdl-semantic` queries backed by slang / pyslang): split out as **#1875** on current `main`.
- **Split out since:** the prompt hook skipping `<task-notification>` messages (#1832) → **#1884**; explicit `projectPath` MCP projects getting a catch-up sync and a watcher (#1835) → **#1886**. Already covered on `main` by others: committed-but-unindexed detection (#1829 → #1878), a live legacy plain-PID daemon lock being preserved (#1834 → #1850).
- **Not yet extracted** (will follow separately if still relevant): C/C++ function-like macros never becoming callees and local-constructor namespace/arity selection (#1838, #1839); UTF-16 column IDs and same-line getter/setter identity in C++ and TS/JS; Dart extension-type container bounds; a busy `codegraph sync` reporting the lock instead of a false success; an awaited `Promise<string>` receiver not binding to an unrelated class method (#1840, cf. #1855).

Validation of the branch as it stands is in `docs/validation-*.md` on the branch (native suite 4956 passed / 9 skipped, WASM 4946 / 19, all at extraction stamp 38).


danusha2345:
Closing this superset: every part that was still relevant now lives in a focused PR on current `main`, and the rest was covered upstream.

- Verilog / SystemVerilog (port of #402 by @FHYQ-Dong plus bindings, access roles, profiles, slang semantics) → **#1875**
- prompt hook skipping `<task-notification>` (#1832) → **#1884**
- explicit `projectPath` projects getting catch-up sync and a watcher (#1835) → **#1886**
- `codegraph sync` reporting a held lock instead of a silent success → **#1889** (the stale-extraction half is #1842 by @bompus)
- C/C++ function-like macros never becoming callees, local constructors chosen by lexical namespace and arity (#1838, #1839) → **#1890**
- committed-but-unindexed detection (#1829) → #1878; live legacy PID locks (#1834) → #1850; awaited `Promise<string>` receivers (#1840) → #1885; the store-callers regression from #1759 → #1862
- Deliberately dropped: UTF-16 column node IDs and same-line getter/setter identity — it changes every C++ and TS/JS node id and deserves its own discussion before code; Dart extension-type bounds — #1780/#1791 cover the case.

The branch stays on the fork as a reference.

PR #1846, opened 2026-09-11
test(mcp): cover per-definition callers truncation

bompus:
## Summary

`codegraph_callers` only groups by definition when the same name has more than one definition. The truncation test on `main` used a single `hot` function, so the grouped `distinct definitions` / `… +N more` assertion failed even though the flat `Showing 20 of 26` marker is correct for that fixture.

This change gives `hot` two definitions so that assertion matches the product path. Handler behavior is unchanged.

## Test plan

- [x] `fnm exec --using codegraph npx vitest run __tests__/mcp-callers-truncation.test.ts --minWorkers 1 --maxWorkers 1` (5/5 pass)

### DIFF
--- a/__tests__/mcp-callers-truncation.test.ts
+++ b/__tests__/mcp-callers-truncation.test.ts
@@ -35,14 +35,17 @@ beforeAll(async () => {
       Array.from({ length: CALLERS }, (_, i) => `export function caller${i}(): number { return warm(${i}); }`).join('\n') +
       '\n'
   );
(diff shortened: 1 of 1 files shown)


bompus:
Closing as superseded: #1862 landed the same per-definition truncation coverage with a stronger fixture (exact-named `hot.ts` + `other-hot.ts`, pinning that a filename is not an overload per #1809). The assertions here are identical to what is now on `main`, so there is nothing left to carry over. Thanks for the review bandwidth.

Issue #999, opened 2026-06-26
codegraph index wedges indefinitely at "Resolving refs" 94% on a repo with a large *tracked* JS/TS asset dir and there's no way to exclude it


DANOX2:
**Summary**
On a Django repo that vendors the **Metronic** admin theme (committed under `static/`, ~1,600 tracked JS/TS files)(I don't know why they did this either....), `codegraph index` / `codegraph init` reaches **`Resolving refs … 94%`** then pins one CPU core at ~100% **indefinitely** observed runs of **28 min and ~15 min with no completion**. Two sibling repos with no such theme index in seconds, so the trigger is the theme's import graph. Compounding it: the theme is **git-tracked**, so there's currently **no way to exclude it** `.gitignore` doesn't apply to tracked files, and `codegraph.json` only offers `includeIgnored`/`extensions`.

**Environment**
- codegraph **1.1.1** (`@colbymchenry/codegraph`, linux-x64), Host Node v20.20.2, Linux 6.17 x86_64
- Target: Django app, **4,956 tracked files**; `static/` (committed Metronic) = **4,489 files** / 124 MB (1,356 `.js`, 236 `.ts`, 206 `.scss`, 2,064 `.svg`, …; 15 files >1 MB correctly auto-excluded); real app code outside `static/` ≈ 467 files

**Steps to reproduce**
1. Repo committing a large JS/TS theme under a **tracked** dir → `codegraph index .`
2. Bar reaches `Resolving refs  ████████████████████████░  94%`
3. Stays at 94%, worker ~100% CPU, never completes.

**Expected**
Either it finishes in reasonable time, or there's a supported way to exclude a large tracked vendor/theme dir.

**Actual**
- Never completes (28 min & ~15 min runs, 100% CPU, frozen at 94%, no DB write, no error).
- Sibling repos (728 files → 7.6 s; 263 files → 4.1 s) index instantly → resolver chokes on the theme's widely-imported JS/TS symbols.
- 1.1.0 changelog says the quadratic "Resolving refs" crawl was fixed, but the symptom persists on 1.1.1 for this graph shape.

**Can't work around via exclusion**
- `static/` is committed → `git check-ignore` returns nothing; git won't apply ignore rules to tracked files, and CodeGraph enumerates via `git ls-files` (#511). So `.gitignore` / `.git/info/exclude` have no effect.
- `codegraph.json` supports `includeIgnored` (opposite) and `extensions`, but **no exclude/ignore-tracked option**.
[...]

PR #1848, opened 2026-09-11
fix(status): see committed-but-unindexed changes without losing the fast path (#1829)

inth3shadows:
**An alternative to #1843 for the same bug — same symptom, same repro, but it keeps the git fast path.** @bompus got there first and his diagnosis matches mine exactly; this is offered as a choice of shape, not as a competing claim on the bug. Take whichever you prefer, or take his and treat this as a follow-up.

The difference is what replaces the incomplete candidate set. #1843 drops the fast path and compares the **full source inventory** on every call. This stamps the commit the index was built at and asks git for the **committed diff since that commit**, so the candidate set becomes complete without the tree walk.

## Why it matters

#1843's own table, on a 14,486-file VS Code checkout:

| State | Before | #1843 |
| --- | ---: | ---: |
| Clean | 50.2 ms | 614.2 ms |
| One uncommitted edit | 59.1 ms | 676.7 ms |
| Committed, not indexed | 51.1 ms (wrong) | 604.4 ms (right) |

`getChangedFiles` is on the hot path — `codegraph status`, every sync, and the watcher's catch-up all land here — so that cost is paid per call, forever, to catch a case that is rare per call.

## Measured here, same checkout, all three arms on one machine

VS Code at `7b7e49c8` — the commit #1843 measured — indexed once (14,137 files, 475,316 nodes) and read by all three builds. Timing `getChangedFiles()` in-process, excluding CLI startup, as #1843's table does: 2 warm-ups then 10 samples, two passes per state with the arm order reversed on the second, `nice -n 10`. WSL2/ext4, Node 22.23.2.

| State | `main` @ 3ed73bc | #1843 | this PR |
| --- | ---: | ---: | ---: |
| Clean | **119 ms** | 1470 ms | **132 ms** |
| One uncommitted edit | **131 ms** | 1488 ms | **136 ms** |
| Committed, not indexed | 122 ms *(reports 0 — the bug)* | 1547 ms | **146 ms** |

Median of 10, both passes within ~1% of each other. Ranges across all samples: `main` 117–184 ms, #1843 1451–2517 ms, this PR 129–150 ms. Both fixes report `modified: 1` in the committed state; `main` reports nothing.

The absolute numbers are higher than #1843's on every arm — different machine, and this box was busier — but all three arms are measured under the same conditions, so the ratio is the point: **the full-inventory walk is ~11× the current cost, the committed diff is ~1.1×.**

## What it does

The index stamps the commit it was built at (`indexed_at_commit`), and the git fast path adds `git diff --name-status --no-renames <stamp> HEAD` to the candidates `git status` already gives it. One extra git call; the rest of the function is untouched.

This is also the only cheap way to catch a committed **modification**. A file the index already tracks is invisible to `git status` once committed, and a DB-side inventory can't see it either without re-hashing the file — which is the walk we are trying to avoid. The committed diff names it directly.

Mechanics worth reviewing:

- The stamp is written by a full index **and** by every whole-tree sync — a `--paths` sync absorbs only part of the diff, so it must not stamp. It is captured **before** change detection runs, so a commit landing mid-sync is never claimed as absorbed.
- Committed lines go through the **same classifier** as working-tree lines, so an excluded dir stays excluded (#766, #999) and a committed `A` the index never saw still lands in `added`.
- `--no-renames` is deliberate: the index keys files by path, so a rename is a removal plus an add.
- A path reaching the list from both sources — committed, then edited again — is counted once.
- A stamp git cannot resolve (rebase, gc, shallow clone) falls back to the existing full scan, as does an index built before stamping existed. The next sync stamps it and the fast path returns, so an existing index self-heals after one slow call and is **never slower than #1843**.
- The exported `getGitChangedFiles` keeps its contract — a working-tree diff for callers holding no index. The trust decision lives in `canTrustGitFastPath`, which only the DB-aware caller asks. (`git-changed-untracked-dir.test.ts` catches this if it regresses; it caught it for me.)

## Where #1843 is strictly better

One case, and it is real: a file hidden from git with `git update-index --assume-unchanged` and then edited. #1843's inventory walk sees it; this does not, because neither `git status` nor `git diff` reports it. Verified on all three builds:

```
base     pending {added: 0, modified: 0}
#1843    pending {added: 0, modified: 1}   <- correct
this PR  pending {added: 0, modified: 0}
```

If that case matters more than the per-call cost, #1843 is the right fix and this one isn't.

## Tests

Nine cases in `__tests__/sync.test.ts`: committed add, committed modify, committed delete, committed rename, zero-after-sync, the double-source count, the exclusion filter, an unresolvable stamp, and a pre-stamp index upgrading. Six of them fail against pre-fix source.

## Validation

Based on upstream `3ed73bc`. Linux/WSL2, Node 22.23.2 for the source build (the source tree's Node gate rejects 26; the published bundle carries its own runtime).

- `npx tsc --noEmit` clean.
- `npm test`: 13 failures, all of which reproduce on pristine `upstream/main` @ `3ed73bc` in a throwaway worktree against the same index — `mcp-callers-truncation`, `nextjs`, `object-literal-methods`, `react-native-bridge`, `ui-server-api`, `ui-steps-api`, `ui-steps-api-servers`, `ui-steps-cross-tier`. No new failures.
- Both repros from #1829 re-run end to end on a built binary.
- Windows and the viewer build were not re-run for this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -151,6 +151,8 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 #### MCP / indexing
 
+- `codegraph status` no longer reports zero pending changes for work you have already committed. Change detection asked git for the working-tree diff, and committing a file is exactly what removes it from that answer — so an edit read as pending until you committed it, then read as nothing while the index still lacked it. It now also asks git what changed between the commit the index was built from and where you are now, so the number matches what a sync would actually do. (#1829)
+
 - The prompt hook no longer injects unrelated projects when run from your home directory or a broader directory containing a stray workspace manifest. (#1454)
 
 - Indexing now succeeds when Node.js's SQLite lacks FTS5, with search falling back to name and fuzzy matching; thanks @aniruddhaadak80. (#1532)
--- a/__tests__/sync.test.ts
+++ b/__tests__/sync.test.ts
@@ -984,6 +984,24 @@ describe('committed-but-unindexed changes (#1829)', () => {
     expect(result.filesModified).toBe(1);
   });
 
+  it('sees a committed RENAME as a removal plus an add', async () => {
+    // `--no-renames` on the committed diff is deliberate: the index keys files
+    // by path, so a rename IS a removal and an add, and pairing them up would
+    // only have to be taken apart again.
+    fs.renameSync(path.join(testDir, 'src', 'one.ts'), path.join(testDir, 'src', 'renamed.ts'));
+    git('add', '-A');
+    git('commit', '-m', 'rename one');
+
+    const changes = cg.getChangedFiles();
+    expect(changes.removed).toContain('src/one.ts');
+    expect(changes.added).toContain('src/renamed.ts');
+
+    const result = await cg.sync();
+    expect(result.filesRemoved).toBe(1);
+    expect(result.filesAdded).toBe(1);
+    expect(cg.searchNodes('alpha').every((r) => r.node.filePath !== 'src/one.ts')).toBe(true);
+  });
+
   it('still filters committed changes by the rules the full index uses', async () => {
     // vendor/ is a built-in exclude git knows nothing about. Sourcing candidates
     // from `git diff` must not smuggle in files `git status` would have had
--- a/__tests__/sync.test.ts
+++ b/__tests__/sync.test.ts
@@ -881,3 +881,147 @@ describe('Scoped sync parity (#watcher-scoped)', () => {
     expect(cg.searchNodes('gamma').length).toBe(1);
   });
 });
+
+// A change that is COMMITTED but not yet indexed used to read as zero pending
+// changes: getChangedFiles' git fast path built its candidate list from
+// `git status --porcelain`, and committing is exactly what removes a file from
+// that output. The hash comparison below it was correct and simply never
+// reached. Committed work is now sourced from `git diff <indexed commit> HEAD`.
+// (#1829)
+describe('committed-but-unindexed changes (#1829)', () => {
+  let testDir: string;
+  let cg: CodeGraph;
+
+  const git = (...args: string[]) =>
+    execFileSync('git', args, { cwd: testDir, encoding: 'utf-8', stdio: ['pipe', 'pipe', 'pipe'] });
+
+  beforeEach(async () => {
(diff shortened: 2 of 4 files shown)


danusha2345:
Tested head 496f1f4 on Linux x64 (Node 24, WASM arm) against a full-inventory control, in throwaway git repos with real SQLite. Your 63 tests pass. Four extra scenarios return `{added: [], modified: [], removed: []}` while the index is stale; the control build reports the changed paths in all four:

1. **`git restore` of an already-indexed uncommitted edit.** Index `source.ts` with a working-tree edit, then `git restore source.ts`. The DB still holds the dirty version, the stamp equals HEAD and `git status` is empty, so nothing is a candidate. The dirty-path set at index/sync time needs to be kept and re-hashed even after it drops out of `git status`; the commit stamp alone can't see it (trust point: `src/extraction/index.ts:1369`).
2. **A commit landing during a full index.** Change + commit a file after it was extracted but before maintenance finishes: the DB has the old symbol, metadata gets the newer HEAD, the next status says clean. The commit should be stamped *before* files are read (`src/index.ts:700`).
3. **Non-ASCII committed path.** Add and commit `тест.ts` after indexing: git quotes the name in textual diff output, the parser doesn't unquote it, the file never becomes a candidate. `-z` (NUL-separated) output fixes it (`src/extraction/index.ts:1337`, `1465`).
4. **A failing diff command hides changes.** Simulated a timeout on the diff `execFileSync`: the inner catch returns `[]`, which reads as "no committed changes" and bypasses the full-scan fallback. The error should propagate to `getGitChangedFiles` so it returns null and falls back (`src/extraction/index.ts:1341`).

(2) is deterministic in the repro (the commit is made inside the maintenance hook); (4) is fault injection on that one call only. The other two use plain git.

The fast-path shape itself is the right one; I didn't re-measure the speedup, those numbers are yours. With the four guards above I'd take this over the full-scan approach.

colbymchenry:
Implemented and merged via #1878 (4871114), preserving your indexed-commit approach and all nine regression cases, with Eric Minish credited as co-author. The follow-up also covers all four cases from the review above, failed sync reads, and dirty-path retention across scoped sync. Full build, focused native/WASM suites, and the original reproduction pass. Closing this proposal as superseded; the fix will be included in the next release. Thank you for the implementation and tests.

PR #1865, opened 2026-09-14
fix(dart): index a Dart 3 extension type as the type it is (#1784)

### LINKED ISSUE #1784: Dart: `extension type` members stopped being indexed in #1780 — the `method_signature` gate has no language check

Dshuishui:
Fixes #1784 — implements the report's option 2. Rebased onto `58c07e8`.

## The change

Four places, in two of which the report already pointed:

```ts
// src/extraction/languages/dart.ts
extraClassNodeTypes: ['mixin_declaration', 'extension_declaration'],   // named in the report
// and dartEnclosingTypeName, which walks up the parents for the enclosing type's name
```

```rust
// codegraph-kernel/src/dart.rs — two matches! arms
"class_definition" | "mixin_declaration" | "extension_declaration" | "enum_declaration"
```

`extension_declaration`, the older `extension`, is in all four;
`extension_type_declaration` is in none. `classTypes` and `enumTypes` are left
alone — `extraClassNodeTypes` is where `extension` and `mixin` already sit, and
that is enough.

## What `main` does today

On `58c07e8`, both extraction paths agree, and both are wrong. The report's
repro yields two top-level functions and no type for them to belong to:

```
function:km   function:report        (no MetersT node)
```

That is the report's option 1 already in the tree. #1862 narrowed the
`isInsideClassLikeNode` gate to TS/JS/ArkTS — the same shape the report
proposed — and it does what the report said it would: `km` is back. It landed as
part of a regression audit rather than against #1784, so the issue reads as
though nothing has moved, which is worth knowing before checking the repro: what
you see now is a top-level function, not the missing member the report describes.

What option 1 leaves is the mis-kinding the report also calls out — a top-level
function rather than a member of its type.

`kernel-dart-parity` is green there — 18/18 — because the two arms agree. The
suite proves the paths match each other; it cannot prove either is right, so
this defect sits behind a green gate with nothing pinning it.

## Result

With this PR, on the report's repro verbatim, both paths:

```
class:MetersT · method:MetersT::km
```

which is the shape the report calls strictly better — a member of its type
rather than a top-level function. With a fixture that has a second member, that
member's span recovers too: it runs to its closing brace instead of stopping at
the signature line. An ordinary `class` is unchanged, and so are `extension` and
`mixin` — those two are already covered by the parity suite's Dart torture
fixtures, which stay green.

The two sides have to move together: with only `dart.ts` changed the walker
reads the type correctly while the kernel keeps minting top-level functions, and
the paths disagree. The numbers for that are under Tests.

## Tests

Adds `__tests__/dart-extension-type.test.ts` — the extension-type fixture the
report notes is missing. It pins the node kind, the qualified name, the span of
the member that follows, and an ordinary class as a control.

Each half of the change is pinned by a different gate, which is worth stating
exactly. Reverting the `dart.ts` hunks fails this suite's first case on the wasm
path (1 of 3). Reverting the `.rs` hunks leaves this suite green — with a kernel
staged it is the kernel that answers — and instead takes `kernel-dart-parity`
from 18/18 to 4 failures. Both measured on this branch.

Full suite on this branch: **4 failures**. `58c07e8`, built and run on its own:
**the same 4**, `git-hooks.test.ts`, which depend on this machine's global
`core.hooksPath`. Nothing here is new. `kernel-dart-parity` stays 18/18, now
agreeing on the right answer.

So this is the report's option 2, on top of option 1 rather than instead of it —
the two are complementary, and #1862 got the first half there ahead of this. What
remains is the difference the report calls strictly better: `method:MetersT::km`
rather than `function:km`.


bompus:
Option 2 matches the diagnosis in #1784: `extension_type_declaration` on `extraClassNodeTypes` and the enclosing-type walk, plus the two kernel `matches!` arms so the native path does not keep minting `km` as a top-level function. The new fixture pins `method:MetersT::km`, the following member's span, and an ordinary class as control.

Option 1 (language-narrow the `method_signature` gate) is still a separate guard, as you said. I have not run the suite on this branch.

bompus:
Independently verified on Linux (REDACTED_SECRETS, from-source kernel build). Two findings, one for the fix and one for the branch:

**1. The fix is correct and complete.** On current `main` (`58c07e8`) both paths emit `function:km` with no `MetersT` node at all — the parity suite is green there only because both arms agree on the wrong answer, so the bug is real and unpinned. With this PR applied: `kernel-dart-parity` 18/18, the new `dart-extension-type` 3/3 (it pins the right shape, `class:MetersT · method:MetersT::km`, plus the second-member span and the ordinary-class control).

**2. The branch needs a rebase, and it is a clean one.** As pushed (based at `3ed73bc`) the branch fails 12 unrelated tests on a current tree — `mcp-callers-truncation`, `object-literal-methods`, `react-native-bridge`, `nextjs`, `ui-steps-*` — purely from staleness (all fixed on `main` by #1849/#1862 after this branched). I cherry-picked `778e372` onto `58c07e8`: zero conflicts (none of its 3 files were touched since), rebuilt the kernel, and re-ran everything — all green, including those 12. So: rebase and this is merge-ready from where I sit.

Dshuishui:
Rebased onto `58c07e8` — and confirmed your reading here first, since it changes what the body should say.

On current `main` both paths emit `function:km` / `function:report` with no `MetersT`, and `kernel-dart-parity` is 18/18 there. So the gate proves the arms agree, not that either is right, and this defect was sitting behind a green one. The body now says that instead of the stale "four parity failures" framing, which was true only against the older base.

The half you flagged as separate is measured too: reverting just the `.rs` hunks leaves the new suite green — with a kernel staged the kernel is what answers — and takes `kernel-dart-parity` from 18/18 to 4 failures. Reverting just the `dart.ts` hunks fails the new suite's first case on the wasm path. That is why both are in one commit.

Full suite here: 4 failures, the same `git-hooks` ones `58c07e8` has on its own.

Dshuishui:
Option 1 is already in the tree — #1862 narrowed the `isInsideClassLikeNode` gate to TS/JS/ArkTS, the same shape the report proposed. The body reflects that now.

Worth flagging because it changes what the repro looks like: on `58c07e8` you get a top-level `function:km` rather than the missing member #1784 describes. Since #1862 landed as a regression audit rather than against that issue, the issue reads as though nothing has moved — easy to check the repro and conclude it is fixed.

What is left is the half the report called strictly better — `method:MetersT::km` instead of `function:km` — which is what this PR does. The two are complementary rather than alternatives.

PR #1867, opened 2026-09-14
fix(explore): retain broad receiver and requested source evidence

bompus:
## Summary

Stacked on #1841 (literal seeds). This adds the follow-up that keeps compound `explore` queries working when the question names a broad receiver or asks for a specific region of a file:

- `src/mcp/explore-source-ranges.ts` — new `requestedSourceRanges()` scores source ranges inside an already-selected file against the question's terms, camelCase identifiers, and seed literals. It never adds graph nodes or edges — it is evidence inside a file the query already chose. Test paths and Vue files get dedicated region handling, and the component's own basename is subtracted from the question so a name that identifies the file doesn't seed every line inside it.
- `src/mcp/tools.ts` + `src/extraction/*` — retain broad-receiver and requested-source evidence through the explore path so readers/writers stay attached instead of degrading to bag-of-words FTS.

## Dependency

Contains the full `contribute/literal-seeds` branch (#1841) via merge commit `4d0f0778`; the diff vs `main` will shrink to just this change once #1841 lands.

## Validation

Regression checks and the upstream-contribution validation runs are recorded in the branch's `docs(explore)` commits (`bca1cc01`, `b56cd7ee`). Not rerun against latest `main` for this draft.


bompus:
Rebased onto current `main` as a single clean commit (the old branch had grown the fork's history; this is now just the feature). No source conflicts beyond the changelog — #1849/#1862 touch different regions of the files this PR changes.

Verified on the rebased tree: `tsc` clean, the PR's 7 test files green (248 passed in the two compiled-gated files included), including `literal-compiled-index` against a fresh `dist` build.

Issue #1779, opened 2026-09-08
Windows: vitest pool worker killed by a V8 compiler-zone OOM, invisible in the JSON report


bompus:
## Summary

Running the full suite on Windows kills a vitest pool worker with a V8 **compiler-zone** OOM. It has reproduced across separate runs with a byte-identical native stack. The crash is **not visible in the JSON report** — a dying worker is not a test failure, so it appears in no `testResults` entry and vitest recycles the worker and continues. A run can therefore report an internally consistent pass/fail summary while a worker was in fact killed.

This is a Zone OOM, not a JS heap OOM, so `--max-old-space-size` does not reach it.

## The crash

    #
    # Fatal process out of memory: Zone
    #
    ----- Native stack trace -----

     1: 00007FF6F95A495D node::MultiIsolatePlatform::DisposeIsolate+5341
     2: 00007FF6FAA12006 v8::base::FatalOOM+54
     3: 00007FF6FA29F5E5 v8::Function::NewInstance+245
     4: 00007FF6F9C19576 v8::SnapshotCreator::GetIsolate+56566
     5: 00007FF6FAE93F5A v8::platform::DelayedTaskQueue::Terminate+1698970
     6: 00007FF6FAE9364B v8::platform::DelayedTaskQueue::Terminate+1696651
     7: 00007FF6FAE92B0D v8::platform::DelayedTaskQueue::Terminate+1693773
     8: 00007FF6FAE925E4 v8::platform::DelayedTaskQueue::Terminate+1692452
     9: 00007FF6FAE91B71 v8::platform::DelayedTaskQueue::Terminate+1689777
    10: 00007FF6FAE918CF v8::platform::DelayedTaskQueue::Terminate+1689103
    11: 00007FF6FAC9F686 v8::base::DiyFp::Multiply+414934
    12: 00007FF6FAC9E3A0 v8::base::DiyFp::Multiply+410096
    13: 00007FF6FAA722EB v8::internal::compiler::CompilationDependencies::FieldTypeDependencyOffTheRecord+192907
    14: 00007FF6F9912F56 DH_get0_engine+658470
    15: 00007FF6F990E693 DH_get0_engine+639843
    16: 00007FF6F990DF79 DH_get0_engine+638025
    17: 00007FF6FAA8519F v8::platform::DefaultJobHandle::UpdatePriority+591
    18: 00007FF6F95A081E node::GetNodeReport+90430
[...]

PR #1872, opened 2026-09-15
Add Android platform analysis and contribution CI

mixxer:
Add AIDL, JNI, HAL, service, and IPC evidence tools with Java/Kotlin extraction and resolution fixes and regression coverage.

Document supported analysis, evidence limits, and public reference inputs. Add contributor guidance and GitHub PR and merge-queue checks for Node 22/24 and native kernel parity.


mixxer:
Pushed an update (f830b36) that fixes two native-kernel parity gaps found in a cross-review of this branch against upstream `main`:

- `codegraph-kernel/src/java.rs`: `extract_anonymous_class` pushed the qualifier-stripped type name as the `extends` reference instead of the full dotted name — unlike the wasm extractor (`tree-sitter.ts`, 2026-09-11 fix). This silently broke AOSP's qualified-name AIDL Stub lookups (`IFoo.Stub`) for any Java file routed to the native kernel.
- `codegraph-kernel/src/kotlin.rs`: no handling existed for `object_literal` (`object : IFoo.Stub() { ... }`), the dominant AIDL Stub implementation idiom in Kotlin AOSP sources. Ported `extractKotlinObjectLiteral` from the wasm extractor verbatim (delegation_specifier walk, extends/instantiates refs, `kotlinObjectScopeDepth` identity binding for anon-object members).

Verified with `cargo build --release` (clean) and the full vitest suite (284 files / 4877 tests passing), including `kernel-kotlin-parity.test.ts`, `kotlin-object-literal.test.ts`, and all `aosp-*.test.ts`.

mixxer:
@colbymchenry could you take a look when you have a chance? This adds the AOSP/AAOS extension plus the native-kernel parity fix noted above.

danusha2345:
I reviewed f830b36 against `main` (ba3c21e) and measured it on a real Android corpus. Summary: the resolver/extraction part is a clear precision win for Java, but the PR bundles several unrelated changes that would be much easier to land separately.

**How I checked it**
- `build:kernel`, `build`, and `tsc` are clean. The full suite passes: 283/284 files. The one failure is the known pool-worker crash in `extraction.test.ts` (#1779), which also reproduces on `main`. Run alone with `--liftoff-only`, that file passes 656/656.
- I indexed the same tree with `main` and with this PR: decompiled SystemUI plus AndroidX, 11,489 Java files.

| | main | PR |
|---|---|---|
| nodes | 499,009 | 499,009 |
| edges | 1,056,261 | 1,057,725 |
| edges only in one build | 1,845 | 3,309 |

I sampled both sides. Most of the edges that disappear were wrong:
- `new IKeyguardService.Stub() {…}` was linked to two unrelated classes, `IAssistHandleService.Stub` and `IOverviewProxy.Stub`, because the bare name `Stub` matched them.
- `View.OnClickListener` anonymous classes were linked to `RadioButtonPreference.OnClickListener`.

With the PR those references stay unresolved, which is the correct result: the real interface isn't in the tree.

Most of the new edges are correct:
- nested supertypes such as `RecyclerView.LayoutManager` → `LinearLayoutManager` now resolve through the `.` → `::` normalization;
- that adds interface-method → implementation call edges on top.

A TypeScript control repo (codegraph's own `src/`) is byte-identical: no edge changes.

**Suggestions**
1. **Split the PR.** The inheritance and anonymous-class changes stand on their own and benefit every Java/Kotlin project, not only AOSP. Right now they share a PR with:
   - 9 MCP tools and 9 CLI commands;
   - a new CI workflow (the repo has none today, so that's the maintainer's call);
   - CONTRIBUTING/README rewrites;
   - an LRU cache for project connections;
   - realpath handling in `validateProjectPath`.

   Separate PRs would be much easier to review and merge.
2. **Bump `EXTRACTION_VERSION`.** The Java anonymous-class `extends` text and the Kotlin `object_literal` extraction change what gets stored, and there's no version bump. Existing indexes won't pick the fix up until someone runs a manual `index -f`.
3. **Remove the duplicated helpers.** `main` already has `isInheritanceRef` and `SUPERTYPE_TARGET_KINDS`, and `matchByExactName` already filters with them. The new `isInheritanceReference` / `TYPE_LIKE_NODE_KINDS` / `filterCandidatesForReference` do the same thing under different names. Both copies now run in the same function. The new set is also narrower: it drops `component` and `namespace`, so the two filters silently disagree.
4. **LRU eviction can close a handle that's still in use.** When the cache is full, `getCodeGraph` calls `close()` on the least-recently-used `CodeGraph` without checking whether a tool call is still using it. Two concurrent MCP sessions touching more than 20 projects could hit this.
5. **Code comments.** Several refer to the review process rather than the code ("Codex adversarial review, 2026-09-10, HIGH-1", "this codebase's real AIDL services", dated regression notes). They'll read as noise to future maintainers.
6. **Known limits of the AOSP tools, just for the docs.** I ran them on the same corpus:
   - `aidl-impl` needs `.aidl` sources and returns `declaration_not_found` even when the generated `IFoo.Stub` Java is in the tree. That's common in vendor and decompiled trees.
   - `trace-permission` lists `<uses-permission>` lines as "Definitions". The evidence text says it can't tell XML element kinds apart; the heading could say that too.

mixxer:
Thanks for the detailed review. I split the changes from this PR into independent PRs based on your suggestions:

- #1927 — Java/Kotlin anonymous extraction and inheritance resolution, including the extraction-version bump and removal of duplicate inheritance filtering (draft pending the final corpus comparison).
- #1928 — Android platform CLI/MCP evidence tools and their guide (draft while public AOSP/AAOS validation and comment cleanup continue).
- #1929 — bounded MCP project cache with eviction deferred until active calls finish.
- #1930 — sensitive-path validation through symlink realpaths.
- #1931 — contributor guide and documentation index.

I have left the proposed CI workflow out of this series; that decision belongs to the repository maintainers. Each new branch starts from upstream `main`, so the PRs can be reviewed independently. I am rerunning the cases you measured against pinned public Android 17 framework, HAL, Car, and AndroidX revisions and will add the exact CodeGraph version, commands, outcomes, and limitations to the Android guide. I will keep #1872 open until the replacement PRs and measurements are complete, then mark it superseded.

mixxer:
The follow-up is now split into independent PRs: #1927 (Java/Kotlin extraction and resolution), #1928 (Android platform tools), #1929 (project-cache lifetime), #1930 (sensitive path validation), and #1931 (contributor docs). The CI workflow is excluded as requested.

All five are ready for review. #1928 passed its macOS build and full Vitest suite (267 files, 4,678 tests passed; 16 files and 192 tests skipped by existing gates). #1930 passed its security suite (47 passed, 2 Windows-only skips). I also ran an independent read-only Orca orchestration review of the split PRs and a second Orca review of the #1928/#1930 fixes; the identified issues were addressed and revalidation passed.

The pinned AOSP/AAOS/AndroidX source revisions and observed tool results are in #1928's guide; the repeat of your three Java cases and the TypeScript control are in its source validation note. These are source-index checks on a sparse checkout, not a direct comparison of file counts against the decompiled corpus used in your review.

I am closing this combined PR as superseded by the focused PRs above.

PR #402, opened 2026-05-25
feat(verilog): index Verilog/SystemVerilog with module-instantiation edges

FHYQ-Dong:
## What

Adds **Verilog / SystemVerilog** (`.v` `.vh` `.sv` `.svh`) as a new indexed language, plus a small traversal change so the flagship tools can walk the HDL design hierarchy.

### Extraction
- Vendored **`tree-sitter-systemverilog@0.3.1`** wasm (ABI 15, ships prebuilt — no Docker/emscripten build). One grammar/extractor covers both Verilog and SV.
- New `src/extraction/languages/verilog.ts` (custom `visitNode` hook — SV nests names deep in `*_ansi_header` / `*_body_declaration`, so generic field-based name extraction can't reach them). Extracts:
  - modules / interfaces / packages → `class` / `interface` (containers)
  - functions / tasks → `function`; parameters / localparams → `constant`
  - typedefs → `type_alias`; package imports → `import`
  - **`module_instantiation` → `instantiates` edge** — the highest-value HDL relationship (the design hierarchy is the HDL equivalent of a call path)
- **Ports and internal signals are intentionally not extracted** (they'd explode the node count without aiding structural queries).

### Traversal (cross-cutting)
`codegraph_trace` / `codegraph_callers` / `codegraph_callees` now follow `instantiates` edges (previously `calls`/`references`/`imports` only). Without this the instantiation edges existed but no flagship tool walked them. `instantiates` is a **precise** tree-sitter edge, so it doesn't widen traces the way fuzzy matches would; `getImpactRadius` already traversed all edge kinds. In OO languages this also surfaces `new X()` construction dependencies.

## Wiring
- `src/types.ts` — `LANGUAGES`
- `src/extraction/grammars.ts` — wasm map, `EXTENSION_MAP`, vendored-path branch, display name
- `src/extraction/languages/verilog.ts` (new) + `languages/index.ts` (`EXTRACTORS`)
- `src/extraction/wasm/tree-sitter-systemverilog.wasm` (vendored; shipped by `copy-assets`; plain git, ~0.77 MB gzipped — consistent with existing wasms, no LFS)
- `src/graph/traversal.ts`, `src/mcp/tools.ts` — `instantiates` traversal

## Testing
- **5 new Verilog cases** in `__tests__/extraction.test.ts` (detection + module/function/task/typedef/import + the `instantiates` reference) — green.
- `graph` / `resolution` / `context` suites: **61/61 green** — no regression from the traversal change.
[...]

PR #1875, opened 2026-09-15
feat(verilog): index Verilog/SystemVerilog, resolve port bindings, signal access and build profiles

danusha2345:
## What

Verilog / SystemVerilog support, carved out of the superset draft #1845 onto current `main` (58c07e8). Five commits, each compiling on its own:

1. **Index Verilog/SystemVerilog** — modules, packages, interfaces and modports, named instances, ports and signals, always/assign blocks, functions/tasks and generate scopes; `tree-sitter-systemverilog` 0.4.0 bundled as wasm. `codegraph_explore` follows `instantiates` edges for Verilog, so `explore "top uart_rx"` prints the path `top → uart_bridge → uart_rx`. Based on the extractor in #402 by @FHYQ-Dong (co-authored).
2. **Resolve instances and port bindings** — named, positional and `.*` connections link an instance's local signals to the instantiated module's formal ports; bit/part selects keep their base signal; local signal references stay in their block scope, and a sync that changes a module header re-derives its whole binding fan-out.
3. **Signal access roles** — `codegraph_explore` with `hdlAccess` (`read` / `write` / `readwrite` / `control` / `event` / `all`; CLI `explore --hdl-access`) lists where a signal is read, written, used as a condition or as a clock/reset event, including the direction of arguments passed to known functions and tasks.
4. **HDL build profiles** — an optional `hdl` section in `codegraph.json` selects source files or filelists, include directories and defines; conditional-compilation branches are indexed under the active profile, `status` reports configured vs indexed profile, and the watcher follows filelists and included headers.
5. **Computed parameters and widths on request** — `codegraph hdl-semantic` runs an installed slang (or pyslang via `--python`) on the active profile and returns evaluated parameters, port widths and macro origins as a separate answer; it never touches the indexed graph and is never run implicitly.

`EXTRACTION_VERSION` 26 → 27: projects with Verilog files need a re-index. No kernel changes (Verilog is not kernel-routed); other languages' node ids are unchanged.

## Verification

- `tsc` clean; HDL suites (`verilog-*`, `hdl-*`, plus `extraction`, `watcher`, `mcp-unindexed`, `sync-rebuild-convergence`): 35 files / 287 tests pass. Full suite on Linux: nothing fails here that passes on `main` (the only differences were `dist/viewer`-less UI tests on both sides).
- Real corpora: a Tang Nano 9k card reader (170 files, 33 `.v`) and a 2-UART bridge; instantiation paths and `hdlAccess write` sites verified by hand. A TypeScript-only project shows no HDL output. `hdl-semantic` was exercised earlier on the pulp-platform AXI wrapper with slang / pyslang (parameter overrides 2/4, macro-derived widths 35/36/34); the machine this PR was cut on has neither installed, so the CLI's refusal path is what ran here.

## Out of scope

Everything else on #1845 (call evidence, index freshness, MCP lifecycle, C/C++ macros) stays there or is superseded on `main`; that draft's description lists where each piece went.


danusha2345:
Merged `main` (51116a2) after #1878 landed. The two overlaps were in `getChangedFiles`: the HDL profile filter now sits inside the single candidate loop #1878 introduced (profile sources join the candidate set, are classified whatever their size, and a non-profile Verilog file is skipped), and the sync tail runs `commitHdlProfile` before `finishGitIndexState` stamps the commit. `tsc` clean; HDL, sync, git-index-currency and watcher suites pass. Note: `extraction.test.ts` currently ends with a `Worker exited unexpectedly` from tinypool on this machine — it reproduces on a clean `main` checkout at 51116a2 too (654/655), so it is not from this branch.

danusha2345:
Merged `main` (d0996a2, after #1885): the only overlap was the `js-builtins` import line in `name-matcher.ts`. `tsc` clean; HDL, awaited-receiver and resolution suites pass.

danusha2345:
Added 8f9d7dd at @FHYQ-Dong's suggestion (#402, now closed in favour of this PR): a parameter references the parameters its value is computed from, so impact on a width/depth parameter follows the derived `localparam` chain inside its module. It reuses the existing scoped signal walk — no new resolver path, no new nodes. Two real designs: +16 and +222 edges, node counts unchanged; Verilog, HDL and extraction suites pass (34 files / 898 tests).

PR #1892, opened 2026-09-16
fix(rust): resolve use super::* callees in #[path] modules

seonghobae:
## Summary

- Map `#[path = "..."] mod name;` inclusions to their declaring parent module file so `super` in an included file refers to the lexical parent, not a path derived from the included file's disk location (Rust Reference, Modules / Paths).
- Extract Rust `use` declarations (including `use super::*` globs) into import mappings and resolve bare callees through the imported module before global name-matching.
- Add regression test and `scripts/repro-rust-path-callee.sh` reproducing ContextualWisdomLab/fast-mlsirm#1837 (same-name `predict_rating_multi` in two crates; `#[path]` test module with `use super::*` must call the parent `scaling.rs` symbol, not the decoy in `lib.rs`).

## Test plan

- [x] `npx vitest run __tests__/rust-path-super-glob.test.ts`
- [x] `npx vitest run __tests__/rust-self-owner.test.ts` (no regression)
- [x] `./scripts/repro-rust-path-callee.sh`

## Context

Fixes false callee edge reported in https://github.com/ContextualWisdomLab/fast-mlsirm/issues/1837 where CodeGraph 1.2.0 linked `pr_elom_rowmean` → `crates/fast-mlsirm-py/src/lib.rs:predict_rating_multi` instead of `crates/mlsirm-core/src/scaling.rs:predict_rating_multi`.

Made with [Cursor](https://cursor.com)


seonghobae:
Maintainer note: this PR is still blocked on required review (`REVIEW_REQUIRED`) with no CI check-runs yet on head `ff944a41`. Local verification previously green per PR test plan (`vitest` + `scripts/repro-rust-path-callee.sh`). Happy to rebase or address any review feedback promptly.

